PR #78, opened 2015-06-02
Add Perl version 5.22 to Travis CI config.

cdybedahl:


pawal:
It seems that Travis cannot do 5.22 yet?

$ perlbrew use 5.22
ERROR: The installation "5.22" is unknown.

cdybedahl:
Yup. I was too hot on the gun :-) It's not exactly an urgent fix. I've tested the engine and the CLI with 5.22, and they work fine.

pawal:
Let's wait then... We could try to as the service when it's going to be supported.

dolmen:
See travis-ci/travis-cookbooks#482 to follow progress on Travis-CI side.

In the mean time you can use [travis-perl/helpers](/travis-perl/helpers).

PR #205, opened 2016-05-30
- First step of dotse/zonemaster-engine#174 Fix.

vlevigneron:
It does not fix the whole thing, because Net::IP::XS will never be used even if present. 
The choice between Net::IP and Net::IP::XS will be add later in Zonemaster::Net::IP when Net::IP::XS issues in this framework will be fixed.
Can be merged if we agree that we use will use only Net::IP for the moment.

### DIFF
--- a/Makefile.PL
+++ b/Makefile.PL
@@ -16,7 +16,7 @@ requires 'Module::Find'          => 0.10;
 requires 'JSON'                  => 2.53;
 requires 'File::ShareDir'        => 1.00;
 requires 'File::Slurp'           => 0;
-requires 'Net::IP'               => 0;
+requires 'Net::IP'               => 1.26;
 requires 'List::MoreUtils'       => 0;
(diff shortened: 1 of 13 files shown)


pawal:
The checks fail.

Have you measured the performance differences between Net::IP::XS? What does the group think? @sandoche2k

pawal:
It seems that you have forgotten the dependency on Net::IP.

vlevigneron:
When I did the performances tests months ago, I noticed it was really
faster with Net::IP::XS. I have not keep the results.

le 30 mai, Patrik Wallström a ?crit :
> The checks fail.
> 
> Have you measured the performance differences between Net::IP::XS? What does the group think? @sandoche2k 
> 
> ---
> You are receiving this because you authored the thread.
> https://github.com/dotse/zonemaster-engine/pull/205#issuecomment-222470690

-- 

PR #229, opened 2016-10-05
Fix perl warning about unquoted 'refs'

mattias-p:
I get lots of warnings from this while running the unit tests. I'm using perl 5.18.


matsduf:
Trivial change. Approved.

PR #244, opened 2016-11-21
Issue zonemaster engine 239

vlevigneron:
The real modification is the one done in Zonemaster::Zone, a key module of the engine, which can have impacts of many parts of the code. So, even if we are able to run all the tests, it is possible that we find out later, that it modifies engine behaviour. So, test as many zones as you can...


lib/Zonemaster/Test/Basic.pm
```diff
@@ -206,14 +206,16 @@ sub basic00 {
...
 
     if ( not $parent ) {
         return info( NO_PARENT => { zone => $zone->name->string } );
     }
 
-    my $p = $parent->query_one( $zone->name, q{NS} );
+    my $p = $parent->query_persistent( $zone->name, q{NS} );
+    if ( not $p ) {
+        $p = $parent->query_one( $zone->name, q{NS} );
+    }
```
matsduf: @vlevigneron, could you add documentation for the new code explaining what it does and when?

vlevigneron:
@matsduf 

In fact, all these methods are documented (pod2text lib/Zonemaster/Zone.pm) .

>METHODS
>    query_one($name[, $type[, $flags]])
>        Sends (or retrieves from cache) a query for the given name, type and
>        flags sent to the first nameserver in the zone's ns list. If there
>        is a response, it will be returned in a Zonemaster::Packet object.
>        If the type arguments is not given, it defaults to 'A'. If the flags
>        are not given, they default to "class" IN and "dnssec", "usevc" and
>        "recurse" according to configuration (which is by default off on all
>        three).
>
>    query_persistent($name[, $type[, $flags]])
>        Identical to query_one, except that instead of returning the packet
>        from the first server that returns one, it returns the first packet
>        that actually contains a resource record matching the requested name
>        and type.

sandoche2k:
@matsduf Thanks for merging or commen in case the PR cannot be merged

matsduf:
@vlevigneron and @sandoche2k, I really would like to see more documentation in the code. I realize that it is not realistic that some go through the code and add documentation, but anytime we do updates in the code we could make sure that the new piece of code is documented or commented.

PR #247, opened 2016-11-22
Issue zonemaster engine 171

vlevigneron:
Add the new step in the algorithm.
MIL and ET still do not pass, but VU and MP do.


lib/Zonemaster/Test/Nameserver.pm
```diff
@@ -216,6 +216,9 @@ sub nameserver01 {
             elsif ( $p->is_redirect and not $p->aa ) {
                 $is_not_recursor{ $local_ns->address->short }++;
             }
+            elsif ( not $p->is_redirect and not $p->aa and not $p->answer and $p->rcode eq q{NOERROR} ) {
+                $is_not_recursor{ $local_ns->address->short }++;
+            }
             $nsnames{ $local_ns->name }++;
```
matsduf: @vlevigneron, can you add a comment that states what use case that the new code will capture?

lib/Zonemaster/Test/Nameserver.pm
```diff
@@ -272,6 +275,9 @@ sub nameserver01 {
                     elsif ( $p->is_redirect and not $p->aa ) {
                         $is_not_recursor{ $local_ns->address->short }++;
                     }
+                    elsif ( not $p->is_redirect and not $p->aa and not $p->answer and $p->rcode eq q{NOERROR} ) {
+                        $is_not_recursor{ $local_ns->address->short }++;
+                    }
                     $nsnames{ $local_ns->name }++;
```
matsduf: @vlevigneron, same thing here, can you add a comment that states what use case that the new code will capture?

vlevigneron:
Well, I added that test because you asked for in a preview message of this
tread. You wrote:

"

The ET server has the same problem as MIL.

The MP servers returns answers with AA unset, ANSWER==0 and status==NOERROR
but no referal. That should also be OK.

One VU server is like the MP servers.

Let us update algorithm to make sure that MP and VU passes, but I see no
algorithm that would let MIL and ET pass.
"
That's what I did. I just added the MP case taht makes MP and VU passes.

2016-11-23 12:16 GMT+01:00 matsduf <notifications@github.com>:

matsduf:
I wish I would find more documentation in the code...

Issue #261, opened 2016-12-13
Stop using JSON::XS in the engine


mtoma:
https://github.com/dotse/zonemaster-backend/issues/182

Was due to the wrong behaviour of JSON::XS in combination with perls before 5.20 and locales where numbers are encoded with a comma instead of a dot.

The folowing test code shows the problem:

```
use strict;
use warnings;
use utf8;
use 5.10.1;

use POSIX qw[setlocale LC_ALL];
use JSON::XS;
use Data::Dumper;

my $number = 1.2324534564657567;

my $json1 = encode_json({ number => $number});
say Dumper($json1);

setlocale( LC_ALL, "fr_FR.UTF-8" );

my $json2 = encode_json({ number => $number});
say Dumper($json2);
````

With the following very wrong result as numbers must always be encoded with a dot in the JSON specification:

```
perl TestJOSN_Locale_Bug.pl 
$VAR1 = '{"number":1.23245345646576}';

$VAR1 = '{"number":1,23245345646576}';
```

This bug shouldn't be present anymore in the code because we set the encoding back to en_US before the engine JSON encoding is done but we should still drop the usage of JSON::XS in favor of JSON::PP everywhere in the engine code as this issue could resurface with some future unrelated modifications of the code.
JSON::PP does this encoding right and the JSON processing we are doing in zonemaster doesnt really jusitfy the use of JSON::XS for performance.

**Comment by vlevigneron** [comment]
fixed by dotse/zonemaster-engine#262

**Comment by mtoma** [comment]
It was reported upstram by Calle a few years back, did not check recently if it was fixed.

**Comment by mtoma** [comment]
Can you provide a use case where the difference in processing time between JSON::XS vs JSON::PP would make any real difference in the zonemaster context?

**Comment by pawal** [comment]
There is a lot of json usage throughout the engine, that I mostly use. A simple performance indicator on the difference between PP and XS is this, which runs through encoding and decoding json a large number of iterations:

JSON::PP:
```
real	1m28.337s
user	1m22.356s
sys	0m0.164s
```

JSON::XS:
```
real	0m1.552s
user	0m1.544s
sys	0m0.008s
```

The difference is astronomical.

The same goes for Net::IP::XS mentioned in other issues...

matsduf:
[...]

PR #262, opened 2016-12-13
Use JSON::PP instead of JSON and JSON::XS

mattias-p:
Use JSON::PP everywhere and JSON and JSON::XS nowhere. Because of #261.


vlevigneron:
Fixes dotse/zonemaster-engine#261

PR #266, opened 2016-12-19
- Fixes dotse/zonemaster-engine#263

vlevigneron:
Version numbering that works with perl.prov and is consistent with PBP recommendations and last CPAN version module syntax.


vlevigneron:
Since dotse/zonemaster-cli#35 has been merged, we need to merge this one to keep consistent. And it generates conflicts each time a pull request is merged before because of the new way to deal with versions numbers.

PR #270, opened 2016-12-26
Fixes dotse/zonemaster-engine#253

vlevigneron:
@matsduf While it is not a real bug, I have added comments in .po files files to ensure that the content of the files is less ambiguous. If you consider it's useful, you can add swedish translation.
The "issue" here is that the key of translation file is the "msgid" which is the same for all these messages. The TAG is present in the file to be sure we don't forget one of them.


matsduf:
> While it is not a real bug, I have added comments in .po files files to ensure that the content of the files is less ambiguous. 

I agree that it is (was) not a bug. It was an inconsistency. One possibility is to only have the changed messages in en.po, the other is to have all. We should not have something in between.

PR #282, opened 2017-02-10
- Fixes dotse/zonemaster-engine#264

vlevigneron:
This fix change the way "special" IPs properties are  loaded. It is no longer constant data in source code but we get CSV files from IANA repository.

Now, we have to find the better way to do the provisionning when IANA repository change.

The command used to get the lastest version of files is util/get_iana_address_spaces_infos.pl

### DIFF
--- a/lib/Zonemaster/Constants.pm
+++ b/lib/Zonemaster/Constants.pm
@@ -1,14 +1,18 @@
 package Zonemaster::Constants;
 
-use version; our $VERSION = version->declare("v1.2.0");
+use version; our $VERSION = version->declare("v1.2.1");
 
 use strict;
 use warnings;
 
+use Carp;
+
+use English qw( -no_match_vars ) ;
+
 use parent 'Exporter';
 use Zonemaster::Net::IP;
 use Text::CSV;
-use File::Share q{:all};
+use File::ShareDir qw[dist_dir dist_file];
 
 use Readonly;
 
@@ -91,7 +95,7 @@ sub _extract_iana_ip_blocks {
     my $csv = Text::CSV->new ({
       binary    => 1,
       auto_diag => 1,
-      sep_char  => ','
+      sep_char  => q{,}
     });
     my @files_details = (
         { name => q{iana-ipv4-special-registry.csv}, ip_version => $IP_VERSION_4 },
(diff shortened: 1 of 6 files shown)


util/get_iana_address_spaces_infos.pl
```diff
@@ -0,0 +1,20 @@
...
+use 5.14.2;
+use warnings;
+use strict;
+
+use LWP::Simple;
+use FindBin;
+
+my $iana_url = q{http://www.iana.org/assignments/};
+my $dest_dir = qq{$FindBin::Bin/../share/};
+my @files_details = (
```
matsduf: In Constants.pm dist_file() from File::Share is used to defined the location of these files. Why not here?

util/get_iana_address_spaces_infos.pl
```diff
@@ -0,0 +1,20 @@
...
+    { name => q{iana-ipv4-special-registry.csv}, url => $iana_url.q{/iana-ipv4-special-registry/iana-ipv4-special-registry-1.csv}, ip_version => 4 },
+    { name => q{iana-ipv6-special-registry.csv}, url => $iana_url.q{/iana-ipv6-special-registry/iana-ipv6-special-registry-1.csv}, ip_version => 6 },
+);
+
+foreach my $file_details ( @files_details ) {
+    getstore(${$file_details}{url}, $dest_dir.${$file_details}{name});
+}
```
matsduf: For me it looks like there is a risk that the files are overwritten with incomplete files. Is that incorrect?

lib/Zonemaster/Constants.pm
```diff
@@ -77,47 +80,46 @@ Readonly our $SOA_RETRY_MINIMUM_VALUE       => 3_600;      # 1 hour
...
+        { name => q{iana-ipv4-special-registry.csv}, ip_version => $IP_VERSION_4 },
+        { name => q{iana-ipv6-special-registry.csv}, ip_version => $IP_VERSION_6 },
+    );
+
+    foreach my $file_details ( @files_details ) {
+        my $first_line = 1;
+        next unless ${$file_details}{ip_version} == $ip_version;
+        my $data_location = dist_file('Zonemaster', ${$file_details}{name});
+        open(my $data, '<:encoding(utf8)', $data_location);
+        while (my $fields = $csv->getline( $data )) {
```
vlevigneron: Well, it's an internal method that is called with $IP_VERSION_4 and $IP_VERSION_6 parameters value. I'll add a test, but I see no case where a developper could make a mistake and pass the 'make test step.

You are right about missing files, that's why they are in the repository. I give more details in next comment.

util/get_iana_address_spaces_infos.pl
vlevigneron: This little "tool" is not "water-proof" yet. I mean, we have to decide how to update files on a regular basis for instance. That's why files are in repository. Usage of this tool is not safe yet. I plan to add usage of temporary files to check consistency between old and new files and things like that. But I think it can be done in a second phase. The purpose of the issue, I guess, was to be able to use data from IANA files, we should add an issue about provisionning of new data.

util/get_iana_address_spaces_infos.pl
vlevigneron: We have to decide if this tool is just used during installation phase. For the moment, it works like that and it is supposed to be launched before the installation (just in case if we want to replace files that are provided in the repository).

lib/Zonemaster/Constants.pm
matsduf: Yes, but what error message will be emitted? And is that good?

util/get_iana_address_spaces_infos.pl
matsduf: In Config.pm you use &File::Share::dist_file() to construct the path. Here in you use "$FindBin::Bin/../share/". Can we be sure that we always get the same result? Why not use &File::Share::dist_file() in both cases?

util/get_iana_address_spaces_infos.pl
matsduf: I think the script should verify that the update was OK before the file is replaced.

lib/Zonemaster/Constants.pm
```diff
@@ -1,12 +1,15 @@
...
-use version; our $VERSION = version->declare("v1.1.2");
+use version; our $VERSION = version->declare("v1.2.0");
 
 use strict;
 use warnings;
 
 use parent 'Exporter';
 use Zonemaster::Net::IP;
+use Text::CSV;
+use File::Share q{:all};
```
matsduf: In Zonemaster/Config.pm we use another module, "File::ShareDir". Doesn't that do the same thing? The fewer dependencies we have the better.

PR #283, opened 2017-02-12
- Fixes dotse/zonemaster-engine#127

vlevigneron:


sandoche2k:
@matsduf please verify

share/fr.po
```diff
@@ -43,23 +43,23 @@ msgstr "Aucune réponse des serveurs de noms sur une requête de type \"A\"."
...
-msgstr "Le serveur de noms {ns} renvoie un enregistrement de type \"A\" en réponse à une requête pour {dname}."
+msgstr "Le serveur de noms {ns} renvoie un enregistrement de type \"A\" en réponse à une requête pour '{dname}'."
 
 #: BASIC:HAS_GLUE
 msgid  "Nameserver for zone {pname} listed these nameservers as glue: {nsnlist}."
-msgstr "Le serveur de noms pour la zone {pname} renvoie la liste de serveurs de noms suivants: {nsnlist}."
+msgstr "Le serveur de noms pour la zone '{pname}' renvoie la liste de serveurs de noms suivants: {nsnlist}."
 
 #: BASIC:HAS_NAMESERVERS
 msgid  "Nameserver {ns} listed these servers as glue: {nsnlist}."
```
matsduf: Here:
`"Nameserver '{ns}' listed these servers as glue: {nsnlist}."`

I think the style could be that an NS list at the end after colon could be without quoting. How should the style be? Should the style be consistent for all languages?

share/fr.po
```diff
@@ -87,15 +87,15 @@ msgstr "IPv6 est activé, il est possible de réaliser une requête de type \"{r
 
 #: BASIC:NO_A_RECORDS
 msgid  "Nameserver {ns} did not return A record(s) for {dname}."
-msgstr "Le serveur de noms {ns} ne renvoie pas d'enregistrement de type \"A\" en réponse à une requête pour {dname}."
+msgstr "Le serveur de noms {ns} ne renvoie pas d'enregistrement de type \"A\" en réponse à une requête pour '{dname}'."
 
 #: BASIC:NO_DOMAIN
 msgid  "Nameserver for zone {pname} responded with NXDOMAIN to query for glue."
-msgstr "Le serveur de noms pour la zone {pname} retourne un code NXDOMAIN à une requête de type \"NS\" pour la zone testée."
+msgstr "Le serveur de noms pour la zone '{pname}' retourne un code NXDOMAIN à une requête de type \"NS\" pour la zone testée."
```
matsduf: Here:

`"Le serveur de noms pour la zone '{pname}' retourne un code NXDOMAIN à une requête de type 'NS' pour la zone testée."`

share/fr_FR.UTF-8.po
```diff
@@ -43,23 +43,23 @@ msgstr "Aucune réponse des serveurs de noms sur une requête de type \"A\"."
 
 #: BASIC:DOMAIN_NAME_LABEL_TOO_LONG
```
matsduf: Same in this file.

share/fr.po
vlevigneron: The issue was about quoting "domain names", that's all. I agree it should be more consistent for "name servers" but it was not the purpose of the issue. And there are lots of occurences of "name servers" in the files. I guess we should create another issue for this case.

PR #284, opened 2017-02-12
- Fixes dotse/zonemaster-engine#260

vlevigneron:


vlevigneron:
Do not merge yet, I modified it too fast, I forgot a case.

matsduf:
Does not seem to work. Travis reports error, and when I test the code from your branch I get the same error:

```
$ prove -I lib t/Test-basic.t 
t/Test-basic.t .. 17/? 
#   Failed test 'svtbarn2.se gives NO_DOMAIN'
#   at t/Test-basic.t line 28.
```

vlevigneron:
@matsduf I removed different irrelevant tests cases and it passes now travis. I guess you can review and merge it if you agree with changes.

PR #285, opened 2017-02-13
- Fixes dotse/zonemaster-engine#278

vlevigneron:
Need to change CLI and Backend, but now the return from the CLI command is more consistent.

> zonemaster-cli 193.in-addr.arpa --ns pri.authdns.ripe.net --ns sns-pb.isc.org --ns tinnie.arin.net --ns sec3.apnic.net --ns abcdef --locale=en_US.utf8
Seconds Level     Message
======= ========= =======
   0.68 CRITICAL  Can not add a fake delegation for domain 193.in-addr.arpa to name server abcdef without IP address.
   6.79 CRITICAL  Not enough data about 193.in-addr.arpa was found to be able to run tests.


sandoche2k:
@matsduf Please take this pull request as priority to review.

sandoche2k:
@vlevigneron merging your pull request, i am not able to get the same results as yours above. Do we wait for the CLI modification to happen?

vlevigneron:
Yes that's correct. There will be a fix in the CLI after this merge in
the Engine. 

le 16 févr., sandoche2k a ?crit :
> @vlevigneron merging your pull request, i am not able to get the same results as yours above. Do we wait for the CLI modification to happen?
> 
> -- 
> https://github.com/dotse/zonemaster-engine/pull/285#issuecomment-280334283

-- 

matsduf:
@vlevigneron and @sandoche2k, I do not think this is the correct solution. I think the Engine should process this in the same way as if there was a real delegation pointing at a nameserver name that could not be resolved. The Engine should do as much as it can, i.e. test the other nameservers, the zone, DNSSEC etc and then in the report include an error for the nameserver name that cannot be resolved.

With this solution there could be other errors that Zonemaster does not report, because it does not do a complete test.

@aabdnn, please also comment.

(I also think that we should find another term for "fake delegation" in Zonemaster. It is not fake, it is just not yet implemented.)

vlevigneron:
OK, I thought this is what was expected. Stop to run in case of
undelegated configuration with a nameserver that can not be resolved.

It's really easy to change, so if everybody agrees on the fact that 
ZM should continue to process tests even if there is a mistaken
nameserver provided, I'll fix this pull request accordingly.

In the same way, CRITICAL level is no longer relevant if we apply this
change. ERROR level should be better.

le 16 févr., matsduf a ?crit :
> @vlevigneron and @sandoche2k, I do not think this is the correct solution. I think the Engine should process this in the same way as if there was a real delegation pointing at a nameserver name that could not be resolved. The Engine should do as much as it can, i.e. test the other nameservers, the zone, DNSSEC etc and then in the report include an error for the nameserver name that cannot be resolved.
> 
> With this solution there could be other errors that Zonemaster does not report, because it does not do a complete test.
> 
> @aabdnn, please also comment.
> 
> (I also think that we should find another term for "fake delegation" in Zonemaster. It is not fake, it is just not yet implemented.)
> 
> -- 
> https://github.com/dotse/zonemaster-engine/pull/285#issuecomment-280404042

-- 

aabdnn:
I agree with @matsduf. If the engine is given a set of name servers, and one of them doesn't even resolve, then it should not result in immediate failure. The tests should all continue to run, but the unresolvable name server's failure to resolve should be emitted as an ERROR.

matsduf:
The difference from today will be that the server that cannot be resolved will not be discarded. The fact that it cannot be resolved will be an error. It will be expected to be one of the NS in the zone, and if it is missing there, that will be an error.

sandoche2k:
@vlevigneron I do agree with the proposal from the wisemen :-) above. Can you take it as priority and update ASAP. Thanks.

vlevigneron:
It should be what you expect now.

matsduf:
@vlevigneron, as far as I can see, https://github.com/dotse/zonemaster-engine/pull/285/commits/4b317810c706e69eca1bbaed08c46aa2ced415e2 has already been merged. Should it really be included here?

vlevigneron:
Would you prefer I create a new PR from this commit 4b31781
<https://github.com/dotse/zonemaster-engine/pull/285/commits/4b317810c706e69eca1bbaed08c46aa2ced415e2>
with
all changes included.

2017-02-27 14:43 GMT+01:00 matsduf <notifications@github.com>:

matsduf:
@vlevigneron, there is no need for any PR for commit https://github.com/vlevigneron/zonemaster-engine/commit/08ea5fe11280b60c17f1ec355423a712069da44d. That commit is already merged.

lib/Zonemaster.pm
```diff
@@ -100,6 +100,27 @@ sub recurse {
 sub add_fake_delegation {
     my ( $class, $domain, $href ) = @_;
 
+    # Check fake delegation
+    foreach my $name ( keys %{$href} ) {
+        if ( not defined $href->{$name} or not scalar @{ $href->{$name} } ) {
```
matsduf: The control is that there is an reference to a non-empty array or not, but no control if the data is really a list of one or more IP addresses. What happens if the data so something else?

lib/Zonemaster.pm
```diff
@@ -100,6 +100,27 @@ sub recurse {
...
+                    Zonemaster->logger->add(
+                        FAKE_DELEGATION_NO_IP => { domain => $domain , ns => $name  }
+                    );
+		}
+            }
+        }  
+    }  
     my $parent = $class->zone( $recursor->parent( $domain ) );
     foreach my $ns ( @{ $parent->ns } ) {
         $ns->add_fake_delegation( $domain => $href );
```
matsduf: Here it looks like we are looking up the parent of the domain, and then resolving the delegation, but since we have a "fake delegation" that is not needed. Or happens here?

vlevigneron:
Sorry for the misunderstanding, I proposed to create a new PR from a
branch that already includes this commit (all commit of this PR would
remain n this new PR except
https://github.com/vlevigneron/zonemaster-engine/commit/08ea5fe11280b60c17f1ec355423a712069da44d).

le 27 févr., matsduf a ?crit :
> @vlevigneron, there is no need for any PR for commit https://github.com/vlevigneron/zonemaster-engine/commit/08ea5fe11280b60c17f1ec355423a712069da44d. That commit is already merged.
> 
> -- 
> https://github.com/dotse/zonemaster-engine/pull/285#issuecomment-282749977

-- 

vlevigneron:
le 27 févr., matsduf a ?crit :
> matsduf commented on this pull request.
> 
> The documentation of the method add_fake_delegation() should be updated since it is no longer mandatory to have the list of IP addresses.
>  
> This method adds some fake delegation information to the system. The arguments are a domain name, and a reference to a hash with delegation
>  information. The keys in the hash must be nameserver names, and the values references to lists of IP addresses for the corresponding nameserver.
> 
> 
> > @@ -100,6 +100,27 @@ sub recurse {
>  sub add_fake_delegation {
>      my ( $class, $domain, $href ) = @_;
>  
> +    # Check fake delegation
> +    foreach my $name ( keys %{$href} ) {
> +        if ( not defined $href->{$name} or not scalar @{ $href->{$name} } ) {
> 
> The control is that there is an reference to a non-empty array or not, but no control if the data is really a list of one or more IP addresses. What happens if the data so something else?

You are correct, there is a bug here (that already exists in the current
version of the code). You should create an issue that I will fix in
priority after we have merged the current PR.

> 
> > +                    FAKE_DELEGATION_IN_ZONE_NO_IP => { domain => $domain , ns => $name }
> +                );
> +            }
> +            else {
> +                my @ips = Net::LDNS->new->name2addr($name);
> +                if ( @ips ) {
> +                    push @{ $href->{$name} }, @ips;
> +                }
> +                else {
> +                    Zonemaster->logger->add(
> +                        FAKE_DELEGATION_NO_IP => { domain => $domain , ns => $name  }
> +                    );
> +		}
> +            }
> +        }  
> +    }  
>      my $parent = $class->zone( $recursor->parent( $domain ) );
>      foreach my $ns ( @{ $parent->ns } ) {
>          $ns->add_fake_delegation( $domain => $href );
> 
> Here it looks like we are looking up the parent of the domain, and then resolving the delegation, but since we have a "fake delegation" that is not needed. Or happens here?
> 

The fake delegation is set to the parent. This is the only part of the
previous code that remains, I guess it does what is expected.


> -- 
> https://github.com/dotse/zonemaster-engine/pull/285#pullrequestreview-24030485

-- 

mattias-p:
@vlevigneron Can you rebase this branch onto 4b31781 (current master)? That way it's easy to review only the relevant changes, and the discussion is kept in a single thread. I think that'd be very proper.

vlevigneron:
Sorry, I saw your comment too late, I did a revert of the commit.

le 27 févr., Mattias Päivärinta a ?crit :
> @vlevigneron Can you rebase this branch onto 4b31781 (current master)? That way it's easy to review only the relevant changes, and the discussion is kept in a single thread. I think that'd be very proper.
> 
> -- 
> https://github.com/dotse/zonemaster-engine/pull/285#issuecomment-282784839

-- 

lib/Zonemaster.pm
```diff
@@ -100,6 +100,27 @@ sub recurse {
...
     my ( $class, $domain, $href ) = @_;
 
+    # Check fake delegation
+    foreach my $name ( keys %{$href} ) {
+        if ( not defined $href->{$name} or not scalar @{ $href->{$name} } ) {
+            if ( Zonemaster::Zone->new( { name => $domain } )->is_in_zone( $name ) ) {
+                Zonemaster->logger->add(
+                    FAKE_DELEGATION_IN_ZONE_NO_IP => { domain => $domain , ns => $name }
+                );
+            }
```
matsduf: If there is an in-zone nameserver (nameserver name) without an IP address, will that name be excluded from further tests or will be kept and compared to the NS in the zone?

The latter is the correct behavior.

lib/Zonemaster.pm
```diff
@@ -100,6 +100,27 @@ sub recurse {
...
+            else {
+                my @ips = Net::LDNS->new->name2addr($name);
+                if ( @ips ) {
+                    push @{ $href->{$name} }, @ips;
+                }
+                else {
+                    Zonemaster->logger->add(
+                        FAKE_DELEGATION_NO_IP => { domain => $domain , ns => $name  }
+                    );
```
matsduf: If there is a nameserver (nameserver name) that cannot be resolved to any IP address, will that name be excluded from further tests or will be kept and compared to the NS in the zone?

The latter is the correct behavior.

lib/Zonemaster.pm
```diff
@@ -123,6 +144,13 @@ sub add_fake_ds {
     return;
 }
 
+sub can_continue {
+    my ( $class ) = @_;
+
+    return 1;
+    
+}
```
matsduf: What is the meaning of can_continue()? It always returns true.

share/en.po
```diff
@@ -865,6 +865,14 @@ msgstr "Followed a fake delegation."
 msgid  "Name server {ns} not adding fake delegation for domain {domain} to itself."
 msgstr "Name server {ns} not adding fake delegation for domain {domain} to itself."
 
+#: SYSTEM:FAKE_DELEGATION_IN_ZONE_NO_IP
+msgid  "Name server {ns} not adding fake delegation for domain {domain}. In zone name server {ns} needs an IP address."
+msgstr "Name server {ns} not adding fake delegation for domain {domain}. In zone name server {ns} needs an IP address."
+
+#: SYSTEM:FAKE_DELEGATION_NO_IP
+msgid  "Can not add a fake delegation for domain {domain} to name server {ns} without IP address."
+msgstr "Can not add a fake delegation for domain {domain} to name server {ns} without IP address."
```
matsduf: From the two error messages I get the impression that the problematic nameservers (nameserver names) have been excluded from further tests. Is that correct or wrong? (Also see comment to lib/Zonemaster.pm.)

mattias-p:
@vlevigneron Oh, but 4b31781 is already included in the master branch. I believe it is a good fix and this PR shouldn't revert it. At the very least reset your branch to 4e82f417 and force push to update this PR. If you'd also rebase your branch onto 4b31781 before you force push, that'd be even better.

matsduf:
@vlevigneron, the following section needs update since IP addresses are no longer mandatory.

```
 =item add_fake_delegation($domain, $data)
 
 This method adds some fake delegation information to the system. The arguments are a domain name, and a reference to a hash with delegation
 information. The keys in the hash must be nameserver names, and the values references to lists of IP addresses for the corresponding nameserver.
 
 Example:
 
     Zonemaster->add_fake_delegation(
         'lysator.liu.se' => {
             'ns.nic.se'  => [ '212.247.7.228',  '2a00:801:f0:53::53' ],
             'i.ns.se'    => [ '194.146.106.22', '2001:67c:1010:5::53' ],
             'ns3.nic.se' => [ '212.247.8.152',  '2a00:801:f0:211::152' ]
         }
    );

```

mattias-p:
@vlevigneron I'll be happy to help you out with git if it's giving you trouble.

lib/Zonemaster.pm
vlevigneron: @matsduf no the nameserver will be kept and used to compare in relevant tests.

lib/Zonemaster.pm
vlevigneron: @matsduf no the nameserver will be kept and used to compare in relevant tests.

share/en.po
vlevigneron: Messages error reflect the previous behaviour, I'll change them.

vlevigneron:
Yes, we have to add the case when list of IP addresses can be empty.

le 28 févr., matsduf a ?crit :
> @vlevigneron, the following section needs update since IP addresses are no longer mandatory.
> 
> ```
>  =item add_fake_delegation($domain, $data)
>  
>  This method adds some fake delegation information to the system. The arguments are a domain name, and a reference to a hash with delegation
>  information. The keys in the hash must be nameserver names, and the values references to lists of IP addresses for the corresponding nameserver.
>  
>  Example:
>  
>      Zonemaster->add_fake_delegation(
>          'lysator.liu.se' => {
>              'ns.nic.se'  => [ '212.247.7.228',  '2a00:801:f0:53::53' ],
>              'i.ns.se'    => [ '194.146.106.22', '2001:67c:1010:5::53' ],
>              'ns3.nic.se' => [ '212.247.8.152',  '2a00:801:f0:211::152' ]
>          }
>     );
> 
> ```
> 
> -- 
> https://github.com/dotse/zonemaster-engine/pull/285#issuecomment-283041660

-- 

sandoche2k:
@vlevigneron and @mattias-p thanks for your updates to close this issue.

lib/Zonemaster/Test.pm
```diff
@@ -1,6 +1,6 @@
 package Zonemaster::Test;
 
-use version; our $VERSION = version->declare("v1.0.2");
+use version; our $VERSION = version->declare("v1.1.1");
```
matsduf: Should we really increase version from 1.0.x to 1.1.x? If so, why should we go to 1.1.1 instead of 1.1.0?

lib/Zonemaster/Translator.pm
```diff
@@ -81,26 +81,28 @@ sub translate_tag {
...
+        "POLICY_DISABLED"               => "The module {name} was disabled by the policy.",
+        "UNKNOWN_METHOD"                => "Request to run unknown method {method} in module {module}.",
+        "UNKNOWN_MODULE"                => "Request to run {method} in unknown module {module}. Known modules: {known}.",
+        "SKIP_IPV4_DISABLED"            => "IPv4 is disabled, not sending query to {ns}.",
+        "SKIP_IPV6_DISABLED"            => "IPv6 is disabled, not sending query to {ns}.",
+        "FAKE_DELEGATION"               => "Followed a fake delegation.",
+        "ADDED_FAKE_DELEGATION"         => "Added a fake delegation for domain {domain} to name server {ns}.",
+        "FAKE_DELEGATION_TO_SELF"       => "Name server {ns} not adding fake delegation for domain {domain} to itself.",
+        "FAKE_DELEGATION_IN_ZONE_NO_IP" => "Name server {ns} not adding fake delegation for domain {domain}. In zone name server {ns} needs an IP address.",
+        "FAKE_DELEGATION_NO_IP"         => "Can not add a fake delegation for domain {domain} to name server {ns} without IP address.",
```
matsduf: The three messages above says "not adding fake delegation", "not adding fake delegation" and "can not add a fake delegation". Those messages look incorrect.

share/en.po
matsduf: I think the messages need update. Also see further up.

lib/Zonemaster.pm
```diff
@@ -1,6 +1,6 @@
 package Zonemaster;
 
-use version; our $VERSION = version->declare("v1.0.17");
+use version; our $VERSION = version->declare("v1.1.1");
```
vlevigneron: Yes, it as been increased to 1.1.0 in commit https://github.com/vlevigneron/zonemaster-engine/commit/27d34d16a7e2b413eab6ca3ceb7d0caccd4092a9. The file has been modified again in commit https://github.com/vlevigneron/zonemaster-engine/commit/373068dad94765d86283c6ac3f571f7c334f615e, then the version number has been increased again.

lib/Zonemaster/Test.pm
vlevigneron: Yes, it as been increased to 1.1.0 in commit https://github.com/vlevigneron/zonemaster-engine/commit/27d34d16a7e2b413eab6ca3ceb7d0caccd4092a9. The file has been modified again in commit https://github.com/vlevigneron/zonemaster-engine/commit/373068dad94765d86283c6ac3f571f7c334f615e, then the version number has been increased again.

lib/Zonemaster/Translator.pm
vlevigneron: OK, I'll modify the messages in a second.

PR #326, opened 2017-10-18
Remove unused internal package version numbers

mattias-p:


matsduf:
@vlevigneron, we have a model for incrementing the version on the main module and the component. What is the model for version number on the individual modules? Do you just use the version number to keep track of what modules that have been updated?

I cannot see any harm of the version numbers, but it we need some explicit model.

mattias-p:
@vlevigneron, thanks for reviewing, and sorry I didn't know that. But I agree with @matsduf. Can you say something about how they're useful so we don't break that use case in the future just because we didn't understand it?

vlevigneron:
le 25 oct., matsduf a ?crit :
> @vlevigneron, we have a model for incrementing the version on the main module and the component. What is the model for version number on the individual modules? Do you just use the version number to keep track of what modules that have been updated?
> 
> I cannot see any harm of the version numbers, but it we need some explicit model.
> 

Yes, I understand that for the packaging, we take only the version
number of the main module. But when I fix issues on different
branches/repositories which are not all on the same version of the 
code, it's the fastest way for me to know if the modules are the same,
which one has the most recent features... And just, to know if a
specific module has changed between differents versions of the main
module...

If it breaks nothing, I would prefer to keep this granularity to track
module changes individually.

> -- 
> https://github.com/dotse/zonemaster-engine/pull/326#issuecomment-339394010

-- 

vlevigneron:
le 25 oct., Mattias Päivärinta a ?crit :
> @vlevigneron, thanks for reviewing, and sorry I didn't know that. But I agree with @matsduf. Can you say something about how they're useful so we don't break that use case in the future just because we didn't understand it?
> 

Is the reply I gave to Mats enough for you or do you need more details ?

> -- 
> https://github.com/dotse/zonemaster-engine/pull/326#issuecomment-339567498

-- 

sandoche2k:
@mattias please respond to Vincent, if you need more details.

mattias-p:
I don't see how the submodule version numbers would break anything in a technical sense. If the submodule version numbers don't follow our [component versioning principle](https://github.com/dotse/zonemaster/blob/master/docs/design/Versions%20and%20Releases.md) they might cause user confusion. So if we keep the submodule version numbers, we should update our [component versioning principle](https://github.com/dotse/zonemaster/blob/master/docs/design/Versions%20and%20Releases.md) to clarify for user what to expect from these version numbers as well.

On the other hand I think git is a better solution to the stated problem.

So what do you prefer @vlevigneron?
1) Maintaining submodule version numbers according to our [component versioning principle](https://github.com/dotse/zonemaster/blob/master/docs/design/Versions%20and%20Releases.md), or
2) Update our [component versioning principle](https://github.com/dotse/zonemaster/blob/master/docs/design/Versions%20and%20Releases.md) to clarify versioning of submodules, or
3) Remove submodule version numbers and track of changes using git?

I'm okay with any one of these three options.

matsduf:
If we decide to have versions on all perl modules (not just the main one) I need to know what I should do if I create a PR changing such a perl module. Should I also update the version number? If so, what is the versioning model? I need to know what version to set on the module.

sandoche2k:
I will move the discussion for 2018.1

mattias-p:
Won't fix. This evolved into dotse/zonemaster#535 instead.

PR #329, opened 2017-10-24
Fix links

mattias-p:
Updates the CPAN badge and the documentation link. (and the CC url that would 301)


mattias-p:
Wrong branch again...

PR #332, opened 2017-10-25
Update Installation.md

matsduf:
For locale to work under CentOS a newer version of Locale::TextDomain is required. To get that, skip the perl-libintl package with an older version (1.20) and install from CPAN (1.28).

### DIFF
--- a/docs/Installation.md
+++ b/docs/Installation.md
@@ -30,23 +30,34 @@ This instruction covers the following operating systems:
    sudo yum groupinstall "Development Tools"
    ```
 
-2) Install packages.
+2) Install Perl core from package.
 
    ```sh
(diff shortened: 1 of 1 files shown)


docs/Installation.md
```diff
@@ -30,23 +30,34 @@ This instruction covers the following operating systems:
...
 
+```sh
+   sudo cpan -i Locale::TextDomain
+   ```
+
+4) Install packages.
+
+   ```sh
+   sudo yum install perl-core perl-ExtUtils-MakeMaker perl-File-ShareDir perl-File-Slurp perl-IO-Socket-INET6 perl-JSON-PP perl-List-MoreUtils perl-Readonly perl-Time-HiRes perl-YAML libidn-devel perl-Devel-CheckLib openssl-devel perl-Test-Fatal
+   ```
```
mattias-p: If it's possible to group the yum dependencies together and install Locale::TextDomain from CPAN afterwards, I think that would be an improvement. If that's not possible the duplication of perl-core should be cleaned up.

PR #336, opened 2017-10-31
Clarify licensing

mattias-p:


Makefile.PL
```diff
@@ -9,7 +9,7 @@ name 'Zonemaster-Engine';
 # share/Makefile
 
 
-license 'http://www.opensource.org/licenses/BSD-3-Clause';
+license 'bsd';
```
matsduf: Why not 'https://opensource.org/licenses/BSD-2-Clause'?

mattias-p: Because the value of field is supposed to be a token, not a url.
http://search.cpan.org/~ether/Module-Install-1.18/lib/Module/Install.pod#license

PR #337, opened 2017-11-01
Release 2.0.1

mattias-p:


Changes
```diff
@@ -1,4 +1,15 @@
-Revision history for Perl module Zonemaster.
+Revision history for Perl distribution Zonemaster-Engine.
```
matsduf: We have different alternatives here... Should we be consistent?

mattias-p: Updated the heading to be consistent with your CLI PR.

PR #340, opened 2017-11-03
Installation fails on a new system without apt-get update

sandoche2k:


sandoche2k:
@matsduf I have verified second time.

docs/Installation.md
```diff
@@ -60,6 +60,10 @@ This instruction covers the following operating systems:
 1) Install dependencies from binary packages:
 
    ```sh
+   sudo apt-get update
```
matsduf: This should be under its own headline. This is not "install dependencies". E.g. "Resynchronize the package index files".

sandoche2k: Fixed by https://github.com/dotse/zonemaster-engine/pull/342

PR #347, opened 2017-11-17
fixes dotse/zonemaster-engine#343

vlevigneron:


sandoche2k:
@matsduf can you verify?

PR #355, opened 2017-12-07
- fixes dotse/zonemaster-engine#295

vlevigneron:


vlevigneron:
A new version of zonemaster-cli which support new features about fake delegations will follow...

lib/Zonemaster/Engine.pm
```diff
@@ -1,6 +1,6 @@
 package Zonemaster::Engine;
 
-use version; our $VERSION = version->declare("v2.0.1");
+use version; our $VERSION = version->declare("v2.0.2");
```
matsduf: When we release the updated Engine, based on this change, the version should be at least v2.1.0 or maybe even v3.0.0. Is the change "breaking"?

vlevigneron: No

vlevigneron:
I’m traveling today but I’ll improve that at the end of the afternoon.

Envoyé depuis iHeure

lib/Zonemaster/Engine/Recursor.pm
```diff
@@ -331,11 +382,24 @@ Internal method. Takes a packet and a recursion state and returns a list of ns o
 Takes a name and returns a (possibly empty) list of IP addresses for
 that name (in the form of L<Zonemaster::Engine::Net::IP> objects). When used
 internally by the recursor it's passed a recursion state as its second
-argument.
+argument. If the name has an entry in %fake_addresses_cache, these are these 
+IP addresses which are returned.
+
+=item add_fake_addresses()
```
mattias-p: Could you describe the method parameters?

lib/Zonemaster/Engine/TestMethods.pm
```diff
@@ -41,7 +50,14 @@ sub method3 {
 sub method4 {
     my ( $class, $zone ) = @_;
 
-    return $zone->glue;
+    if ( keys %Zonemaster::Engine::Recursor::fake_addresses_cache ) {
+        my $aref = [];
+        tie @$aref, 'Zonemaster::Engine::NSArray', keys %Zonemaster::Engine::Recursor::fake_addresses_cache;
+        return $aref;
```
mattias-p: Can you update the method documentation for method4 to describe the set of values that can be returned and what they mean? Perhaps a link to https://github.com/dotse/zonemaster/blob/master/docs/specifications/tests/Methods.md#method-4-obtain-glue-address-records-from-parent?

lib/Zonemaster/Engine/TestMethods.pm
```diff
@@ -29,41 +37,35 @@ sub method3 {
...
     else {
-        return $zone->ns;
+        return $zone->glue;
     }
 }
 
+sub method5 {
+    my ( $class, $zone ) = @_;
+
+    return $zone->ns;
```
mattias-p: Could you update the method documentation for method5 to describe the set of values it can return and what they mean? Perhaps link to https://github.com/dotse/zonemaster/blob/master/docs/specifications/tests/Methods.md#method-5-obtain-the-name-server-address-records-from-child?

lib/Zonemaster/Engine/TestMethods.pm
```diff
@@ -29,41 +37,35 @@ sub method3 {
...
-            next if not $p;
-            push @nsnames, $p->get_records_for_name( q{NS}, $zone->name );
-        }
-        @child_nsnames = uniq map { name( lc( $_->nsdname ) ) } @nsnames;
+    my $ns_aref = $zone->query_all( $zone->name, q{NS} );
+    foreach my $p ( @{$ns_aref} ) {
+        next if not $p;
+        push @nsnames, $p->get_records_for_name( q{NS}, $zone->name );
     }
+    @child_nsnames = uniq map { name( lc( $_->nsdname ) ) } @nsnames;
```
mattias-p: Could you update the method documentation for method3 to describe the set of values it can return and what they mean? Perhaps link to https://github.com/dotse/zonemaster/blob/master/docs/specifications/tests/Methods.md#method-3-obtain-name-servers-from-child?

mattias-p:
I don't quite grasp Github's management of review comments. At first I added comments declaring what items had been addressed, but that didn't end up looking right. So instead I removed the "has been removed" comments and edited the original comments with a bold footer. Not sure how to make this review process better.

matsduf:
See comments https://github.com/dotse/zonemaster-cli/pull/63#issuecomment-352000972 and https://github.com/dotse/zonemaster-cli/pull/63#issuecomment-352004780 on related PR dotse/zonemaster-cli#63.

vlevigneron:
@mattias-p Done unless for documentation of lib/Zonemaster/Engine/TestMethods.pm which is specified in https://github.com/dotse/zonemaster/blob/master/docs/specifications/tests/Methods.md . We should complete this link one day (we have plans to upgrade "Methods" algorithme, it could be done at this occasion).

lib/Zonemaster/Engine/Recursor.pm
```diff
@@ -331,11 +389,24 @@ Internal method. Takes a packet and a recursion state and returns a list of ns o
 Takes a name and returns a (possibly empty) list of IP addresses for
 that name (in the form of L<Zonemaster::Engine::Net::IP> objects). When used
 internally by the recursor it's passed a recursion state as its second
-argument.
+argument. If the name has an entry in %fake_addresses_cache, these are these 
+IP addresses which are returned.
+
+=item add_fake_addresses($data)
+
+Class method to create fake adresses for fake delegations from data provided.
```
mattias-p: Can you also describe what types of $data is acceptable by the method?
Oh, and mention how class members are affected.

mattias-p:
@vlevigneron Links from the test method API documentation of the test method specification documentation would be very helpful. With links to the specifications it's not necessary to describe the meaning of the return values yet again. But the data type of the return value is not obvious from the specifications.

lib/Zonemaster/Engine/Recursor.pm
vlevigneron: I know it could be more precise but It has the same level of documentation that others methods with the same kind of parameter. If we detail to much here it will be difficult to maintain both spec/code/doc...

mattias-p: The first thing I would change to make it easier to keep code/doc in sync is to move the documentation of each code element (e.g. methods) to right above the element itself.

Linking to specs from docs instead of writing the same thing again is a very good way of keeping doc/spec in sync.

And I don't think it's unreasonable to document the data types of parameters and return values.

vlevigneron:
Sorry guys for this mess. It took me more time than expected to modify this sensitive part of the code I had not changed for years. I found a lighter way to fix the issue. If this version works (it should be) I guess it can be finally merged.

lib/Zonemaster/Engine/Zone.pm
```diff
@@ -54,11 +54,24 @@ sub _build_glue_names {
 
 sub _build_glue {
     my ( $self ) = @_;
+    my @glue_names = @{ $self->glue_names };
+
+    if ( $Zonemaster::Engine::Recursor::fake_addresses_cache{$self->name->string} ) {
+        my @ns_list;
+        foreach my $ns ( @glue_names ) {
+            foreach my $ip ( @{ $Zonemaster::Engine::Recursor::fake_addresses_cache{$self->name->string}{$ns} } ) {
```
mattias-p: This line assumes that all glue names are present in the fake address cache. Is that intended?

vlevigneron: No, but in this context, it makes no difference for Perl. While useless, a test could help understand the code. I'll add it.

lib/Zonemaster/Engine/Zone.pm
```diff
@@ -55,12 +55,15 @@ sub _build_glue_names {
 sub _build_glue {
     my ( $self ) = @_;
     my @glue_names = @{ $self->glue_names };
+    my $zname = $self->name->string;
 
-    if ( $Zonemaster::Engine::Recursor::fake_addresses_cache{$self->name->string} ) {
+    if ( $Zonemaster::Engine::Recursor::fake_addresses_cache{$zname} ) {
```
mattias-p: Could you update the glue attribute documentation to mention how the fake address cache is consulted, and also clarify when the glue list is lazy and when it's not?

vlevigneron: updated

lib/Zonemaster/Engine/Zone.pm
```diff
@@ -332,8 +332,12 @@ A reference to an array of L<Zonemaster::Engine::Nameserver> objects for the
 domain, built by taking the list returned from L<glue_names()> and
 looking up addresses for the names. One element will be added to this
 list for each unique name/IP pair. Names for which no addresses could
-be found will not be in this list. The list is lazy-loading, so take
-care to only look at as many entries as you really need.
+be found will not be in this list. In this case, the list is lazy-loading, so take
+care to only look at as many entries as you really need. In case of 
+undelegated tests and fake delegation the IP associated with name servers
```
matsduf: Are "undelegated tests" and "fake delegation" the same thing or different things?

lib/Zonemaster/Engine/Zone.pm
```diff
@@ -332,8 +332,12 @@ A reference to an array of L<Zonemaster::Engine::Nameserver> objects for the
...
+for the tested zone will be the ones set by users (saved in
```
matsduf: When you write "name servers for the tested zone will be the ones set by users" I guess that you mean the in-bailiwick nameservers?

"In case of fake delegation (undelegated tests), the IP addresses associated with in-bailiwick nameservers will be the ones set by the user (saved in %Zonemaster::Engine::Recursor::fake_addresses_cache), instead of the ones found recursively. (For definition of "in-bailiwick, see RFC 7719.)

vlevigneron: The way namservers and there IP addresses are set for undelegated tests is explained in Zonemaster::Engine::add_fake_delegation and we will use these nameservers/IP addresses. I mean, that we will not do additional recurse queries to find addresses except the one done during add_fake_delegation method call. For isntance:

```
$ zonemaster-cli telia.com --ns dns2.telia.com --ns dns1.telia.com/145.239.76.199 --ns dns49.de.telia.net --show_module --locale en_US
Seconds Level     Module       Message
======= ========= ============ =======
   2.37 ERROR     SYSTEM       The fake delegation of domain telia.com includes an in-zone name server dns2.telia.com without mandatory glue (without IP address).
   0.54 ERROR     BASIC        Nameserver dns1.telia.com/145.239.76.199 did not return NS records. RCODE was REFUSED.
  12.13 WARNING   CONNECTIVITY All nameservers in the delegation have IPv6 addresses in the same AS (1299).
  12.34 NOTICE    CONSISTENCY  Parent has extra nameserver IP address(es) not listed at child (145.239.76.199).
  12.34 NOTICE    CONSISTENCY  Child has extra nameserver IP address(es) not listed at parent (204.70.57.242;81.228.10.67;81.228.11.67).
  12.41 NOTICE    DNSSEC       There are neither DS nor DNSKEY records for the zone.
  12.41 NOTICE    DNSSEC       The zone is not signed with DNSSEC.
  12.48 WARNING   DELEGATION   Nameserver dns1.telia.com response is not authoritative on UDP port 53.
  12.49 WARNING   DELEGATION   Nameserver dns1.telia.com response is not authoritative on TCP port 53.
  13.04 NOTICE    DELEGATION   Child has nameserver(s) not listed at parent (ns02.savvis.net).
  16.79 NOTICE    ZONE         SOA 'refresh' value (10800) is less than the recommended minimum (14400).
 ```

Comment (vlevigneron) [reply lib/Zonemaster/Engine/Zone.pm]: Another example with out of bailiwick nameservers:

```
$ zonemaster-cli 200.193.193.in-addr.arpa --ns ns.lucky.net --ns ns.gu.kiev.ua/145.239.76.199 --locale en_US
Seconds Level     Message
======= ========= =======
   0.56 ERROR     Nameserver ns.gu.kiev.ua/145.239.76.199 did not return NS records. RCODE was REFUSED.
  27.06 NOTICE    Parent has extra nameserver IP address(es) not listed at child (145.239.76.199).
  27.14 NOTICE    There are neither DS nor DNSKEY records for the zone.
  27.14 NOTICE    The zone is not signed with DNSSEC.
  27.15 WARNING   Nameserver ns.gu.kiev.ua response is not authoritative on UDP port 53.
  27.15 WARNING   Nameserver ns.gu.kiev.ua response is not authoritative on TCP port 53.
  28.01 NOTICE    Nameserver ns.lucky.net/193.193.193.100 allow zone transfer using AXFR.
  28.80 NOTICE    SOA 'refresh' value (3600) is less than the recommended minimum (14400).
  28.80 NOTICE    SOA 'retry' value (900) is less than the recommended minimum (3600).
  28.94 NOTICE    No target (MX, A or AAAA record) to deliver e-mail for the domain name.
```

PR #361, opened 2017-12-15
Profiles doc

mattias-p:
Partial fix for dotse/zonemaster#498

I've made a lot of assumptions here. Please check carefully.

The change in Overview.pod is completely unrelated. Just a couple of drive-by corrections.

### DIFF
--- a/lib/Zonemaster/Engine/Config.pm
+++ b/lib/Zonemaster/Engine/Config.pm
@@ -365,6 +365,10 @@ The number of times a query is sent before we give up. Can be set to zero, altho
 
 If set, queries that get truncated UDP responses will be automatically retried over TCP. Default not set.
 
+=head2 resolver.source
+
+The source address all resolver objects should use when sending queries, if one is set.
+
 =head2 net.ipv4
 
 If set, resolver objects are allowed to send queries over IPv4. Default set.
--- a/lib/Zonemaster/Engine/Config.pm
+++ b/lib/Zonemaster/Engine/Config.pm
@@ -256,52 +256,18 @@ Zonemaster::Engine::Config - configuration access module for Zonemaster::Engine
 
     Zonemaster::Engine->config->no_network(1); # Forbid network traffic
 
-    my $value = Zonemaster::Engine::Config->get->{key}{subkey}; # Not really recommended way to access config data
+    my $value = Zonemaster::Engine::Config->get->{key}{subkey}; # Not really recommended way to access profile data
 
-=head1 LOADING CONFIGURATION
+=head1 LOADING PROFILES
 
-Configuration data is loaded in several stages, each one overlaying the result
-from the previous one (that is, the later in the list take priority over the
-earlier). The first stage is hardcoded into the source code and loaded while it
-is being compiled, to make sure that there will always be some basic
-information available. Later, when the configuration object is first used, the
-system will look for a file named F<config.json> in each of a list of
-directories. If the file exists, is readable and contains proper JSON data, it
-will be loaded and overlaid on the current internal config. The directories
-are, in order from first checked to last:
-
-=over
-
-=item The L<Zonemaster::Engine> perl module installation directory
-
-This is where the installation process puts the default configuration. It is
-not meant to be modified by the user, and it will be overwritten when the
-module is upgraded (or reinstalled for any other reason). If you really need to
-know where it is, you can either check the log message left when loading it or
-run this command to find the path:
-
-    perl -MFile::ShareDir=dist_dir -E 'say dist_dir( "Zonemaster-Engine" )'
-
-=item /etc/zonemaster
-
-Intended to hold system-global configuration changes.
-
-=item /usr/local/etc/zonemaster
-
-Basically the same as the previous one, but for those who like to keep their
-locally installed software inside F</usr/local>.
-
-=item ~/.zonemaster
-
-That is, a F<.zonemaster> directory in the home directory of the current user.
-Intended, obviously, for configuration changes local to one particular user.
-
-=back
+Initial profile data for the effective profile is loaded from the default profile.
+The effective profile can be updated with custom profile data.
+The default profile is parsed from a JSON file called F<default.profile> in the Zonemaster-Engine L<dist_dir|File::ShareDir/dist_dir>.
 
 The possible contents of the JSON data is described further down in this manual
 page.
 
-=head1 METHODS FOR CONFIGURATION ITEMS
+=head1 PROFILE DATA ACCESS METHODS
 
 =over
 
@@ -337,38 +303,20 @@ Returns a reference to the list of ASN lookup domains.
 
 =back
 
-=head1 METHODS
+=head1 MANAGEMENT METHODS
 
 =over
 
 =item get()
 
-Returns a reference to a hash with configuration values.
+Returns the effective profile in the data structure described in the L<PROFILE DATA> section.
 
-=item policy()
+=item load_profile_file($path)
 
-Returns a reference to the current policy data. The format of that data is described further down in this document.
+Loads profile data from the given file and merges it into the effective profile.
 
-=item load_policy_file($filename)
-
-Load policy information from the given file and merge it into the pre-loaded
-policy. Information from the loaded file overrides the pre-loaded information
-when the same keys exist in both places.
-
-If the given name does not lead directly to a readable file, each of the usual
-directories will be checked if the name is there. If the plain name isn't, the
-suffix C<.json> will be appended and another try will be done. For example, a
-file F<$HOME/.zonemaster/Example.json> may be loaded by calling this method
-with the string C<"Example">.
-
-=item load_config_file($filename)
-
-Load configuration information from the given file and merge it into the pre-loaded config. Information from the loaded file overrides the pre-loaded information when the same keys exist in both places.
-
-=item load_module_policy($module)
-
-Loads policy data included in a test module. The argument must be the short
-form (without the initial C<Zonemaster::Engine::Test::>) and correctly capitalized.
+The given path must be a JSON file matching the L<PROFILE DATA> format.
+Data from the file overrides the effective profile when the same keys exist in both places.
 
 =item BUILD
 
@@ -377,15 +325,15 @@ Internal method only mentioned here to please L<Pod::Coverage>.
 =item should_run($name)
 
 Given a test case name, it returns true if that test case should be included in
-a test run according to the currently active policy or false if not.
+a test run according to the effective profile or false if not.
 
 =back
 
-=head1 CONFIGURATION DATA
+=head1 PROFILE DATA
 
-The configuration data is stored internally in a nested hash (possibly with arrays as values in places). As of this writing, the file format used is JSON.
+Profile data is represented as a nested hash (possibly with arrays as values in places).
 
-The interesting keys are as follows.
+The allowed keys are as follows.
 
 =head2 resolver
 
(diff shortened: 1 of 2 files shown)


mattias-p:
Travis blows up because I removed documentation for some methods I think don't make much sense after the profiles update.

lib/Zonemaster/Engine/Config.pm
```diff
@@ -377,67 +335,64 @@ Internal method only mentioned here to please L<Pod::Coverage>.
...
-
-The configuration data is stored internally in a nested hash (possibly with arrays as values in places). As of this writing, the file format used is JSON.
-
-The interesting keys are as follows.
-
-=head2 resolver
+=head1 PROFILE DATA
 
-=head3 defaults
+Profile data is represented as a nested hash (possibly with arrays as values in places).
```
matsduf: Suggestion:

"Profile data is represented as a nested hash with a value. The value can either be a scalar value or an array."

lib/Zonemaster/Engine/Config.pm
```diff
@@ -447,7 +402,7 @@ used, the rest are backups in case the earlier ones don't work.
 
 =head2 logfilter
 
-By using this key, the log level of messages can be set in a much more fine-grained way than by the policy file. The intended use is to remove known erroneous results. If you, for example, know that a certain name server is recursive and for some reason should be, you can use this functionality to lower the severity of the complaint about it to a lower level than normal.
+By using this key, the severity level of messages can be set in a much more fine-grained way than by the C<test_levels> item. The intended use is to remove known erroneous results. If you, for example, know that a certain name server is recursive and for some reason should be, you can use this functionality to lower the severity of the complaint about it to a lower level than normal.
```
matsduf: This line is too long. Please wrap at 70 characters.

lib/Zonemaster/Engine/Overview.pod
```diff
@@ -24,9 +24,12 @@ The simple-but-inflexible way is that all the methods in L<Zonemaster::Engine> t
 
  perl -MZonemaster::Engine -E 'say "$_" for Zonemaster::Engine->new->test_zone("example.org")'
```
matsduf: Running this testing iis.se produces 149 lines of logs. See next comment.

lib/Zonemaster/Engine/Overview.pod
```diff
@@ -24,9 +24,12 @@ The simple-but-inflexible way is that all the methods in L<Zonemaster::Engine> t
...
+The main drawbacks of this method are that there is no choice about what messages to see, and it's entirely synchronous.
+The code that started the test does not get a chance to do anything at all until the whole test has finished, which may be several minutes later.
 
-The get around those drawbacks there is the flexible-but-complex way, which consists of installing a callback that gets executed every time a message is logged. It's not that much more complicated, code-wise. The following example does roughly the same thing as the one above:
+To get around those drawbacks there is the flexible-but-complex way, which consists of installing a callback that gets executed every time a message is logged.
+It's not that much more complicated, code-wise.
+The following example does roughly the same thing as the one above:
 
  perl -MZonemaster::Engine -E 'Zonemaster::Engine->logger->callback(sub {say "$_[0]"}); Zonemaster::Engine->new->test_zone("example.org");'
```
matsduf: Running this testing iis.se produces 16698 lines of logs. See previous comment.

lib/Zonemaster/Engine/Config.pm
```diff
@@ -457,7 +412,7 @@ The data under the C<logfilter> key should be structured like this:
              "when"
                 Hash with conditions
              "set"
-                Level to set if all conditions match
+                Severity level to set if all conditions match
 
 The hash with conditions should have keys matching the attributes of the log entry that's being filtered (check the translation files to see what they are). The values for the keys should be either a single value that the attribute should be, or an array of values any one of which the attribute should be.
```
matsduf: The tree that starts on line 421 (in the updated version) starts with the key "SYSTEM", but it should rather start with "logfiler", shouldn't it? I.e.

```
logfilter {
        "SYSTEM": {
            "FILTER_THIS": [
                {
                    "when": {
                        "count": 1,
                        "type": ["this", "or"]
(...)
```

mattias-p:
@matsduf Do you have a particular reason for picking this very number for line width? Is it okay if I use the `fmt` tool (from coreutils)? `fmt` aims for 70 but has a hard maximum of 75.

I avoided wrapping my lines for ease of reviewing. A small change at the beginning of a long line wrapped paragraph will often cause the entire paragraph to reflow. Diff tools are line oriented and don't understand reflows very well, so they mark the entire paragraph as changed.

I suggest a compromise where lines are wrapped both at 70-ish characters as well as at sentences. How does that sound to you?

lib/Zonemaster/Engine/Overview.pod
mattias-p: What do you suggest I do about this?

lib/Zonemaster/Engine/Config.pm
mattias-p: I don't think so. 1) The line above it describes it as "a complete entry". An entry being assumed to mean a module-tag pair. 2) It matches the structure in the previous block quote.

matsduf: Well, then first show the high level structure and then a branch. As the text is now it is unclear. Maybe this should be in a separate file instead of in Config.pm.

lib/Zonemaster/Engine/Overview.pod
matsduf: See next comment...

lib/Zonemaster/Engine/Overview.pod
matsduf: It is, obviously, not true that "The following example does roughly the same thing as the one above" when there is such a difference in output.

matsduf:
I am fine with something in between 60 and 80. I just want the text to wrap when I look at it in Github's various rendering. It is hard to read when lines continue to the right of my screen.

lib/Zonemaster/Engine/Overview.pod
mattias-p: I don't agree that the statement is completely indefensible. One possible interpretation is that it performs the same queries but with slightly different filter settings.

mattias-p: I can do some digging to figure out exactly what's happening and improve the wording. Is that what you're asking?

lib/Zonemaster/Engine/Config.pm
mattias-p: I'll rework this part of the documentation.

lib/Zonemaster/Engine/Overview.pod
matsduf: Or consider removing the example.

mattias-p:
@matsduf I believe I've addressed all stated concerns. Can you take another look?

lib/Zonemaster/Engine/Config.pm
```diff
@@ -250,58 +250,38 @@ __PACKAGE__->meta->make_immutable;
...
-
-Basically the same as the previous one, but for those who like to keep their
-locally installed software inside F</usr/local>.
+A profile is a set of configuration options (a.k.a. profile data)
+regarding:
 
-=item ~/.zonemaster
+ * sets of tests to run
+ * test result severity levels
+ * DNS querying behavior
```
matsduf: * Set of tests to run.
* Test results serverity leves.
* Filtering of test results.
* IPv4/IPv6 connection.
* Network connection on/off.
* DNS querying behavior.
* ASN database setting

mattias-p: While I agree that profiles cover each one of the specifics you list, I don't think they belong in this particular list. My intention with this list is not to give a precise definition of what's included in profiles, but rather to convey a broad idea of what kinds of configurability to expect. The list is meant to be very short and the items are meant to be rubber-bandy. As such I believe that each one of the specifics you list are covered by the three present items.

lib/Zonemaster/Engine/Config.pm
```diff
@@ -250,58 +250,38 @@ __PACKAGE__->meta->make_immutable;
...
 
-=back
-
-The possible contents of the JSON data is described further down in this manual
-page.
+Zonemaster::Engine::Config reads profile data from disk in a JSON
+sub-format.
+The JSON sub-format is described below in the L<PROFILE DATA> section.
+The default profile is read from the file F<default.profile> located by
+L<dist_dir|File::ShareDir/dist_dir> for Zonemaster-Engine.
```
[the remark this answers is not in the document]
mattias-p: I believe not. That's why I didn't mention it.

lib/Zonemaster/Engine/Config.pm
```diff
@@ -250,58 +250,38 @@ __PACKAGE__->meta->make_immutable;
...
+ * sets of tests to run
+ * test result severity levels
+ * DNS querying behavior
 
-That is, a F<.zonemaster> directory in the home directory of the current user.
-Intended, obviously, for configuration changes local to one particular user.
+The I<effective profile> is a set of profile values that is consulted
+by various parts of Zonemaster::Engine.
+The I<default profile> is a set of profile values that is used to
+initialize the effective profile.
```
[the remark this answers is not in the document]
mattias-p: No, I'm referring to the idea of a set of default values for the effective profile.

lib/Zonemaster/Engine/Config.pm
```diff
@@ -250,58 +250,38 @@ __PACKAGE__->meta->make_immutable;
...
 
-=item ~/.zonemaster
+ * sets of tests to run
+ * test result severity levels
+ * DNS querying behavior
 
-That is, a F<.zonemaster> directory in the home directory of the current user.
-Intended, obviously, for configuration changes local to one particular user.
+The I<effective profile> is a set of profile values that is consulted
+by various parts of Zonemaster::Engine.
```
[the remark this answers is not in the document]
mattias-p: No. This part of the documentation is not about information flow. It's about what the information represents while it just sits there motionless. This is an object oriented perspective.

lib/Zonemaster/Engine/Config.pm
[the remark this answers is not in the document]
matsduf: I still do not understand.

lib/Zonemaster/Engine/Config.pm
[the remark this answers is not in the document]
matsduf: It must be clear for the reader where the values in the _effective profile_  comes from, and that the _effective profile_ profile that has been loaded at run-time.

lib/Zonemaster/Engine/Config.pm
```diff
@@ -250,58 +250,38 @@ __PACKAGE__->meta->make_immutable;
...
-from the previous one (that is, the later in the list take priority over the
-earlier). The first stage is hardcoded into the source code and loaded while it
-is being compiled, to make sure that there will always be some basic
-information available. Later, when the configuration object is first used, the
-system will look for a file named F<config.json> in each of a list of
-directories. If the file exists, is readable and contains proper JSON data, it
-will be loaded and overlaid on the current internal config. The directories
-are, in order from first checked to last:
+Zonemaster::Engine::Config provides methods for reading and updating
+the effective profile.
```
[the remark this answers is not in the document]
mattias-p: Not necessarily. I guess it depends on how it's implemented. At this time I'm trying to focus as much as possible on the interface and leave as much room as possible for different implementations.

lib/Zonemaster/Engine/Config.pm
```diff
@@ -250,58 +250,38 @@ __PACKAGE__->meta->make_immutable;
 
 =head1 NAME
 
-Zonemaster::Engine::Config - configuration access module for Zonemaster::Engine
+Zonemaster::Engine::Config - profile access module for Zonemaster::Engine
```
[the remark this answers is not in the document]
vlevigneron: @matsduf I would prefer to offer backward compatibility for the next revision of code. When we will decide that only profile should be used for configuration purposes we should remove all references to Config and replace it by Profile. So I guess it can be done in a future step.

mattias-p:
This PR was obsoleted by #376.

PR #375, opened 2018-02-06
Fix factual error

mattias-p:
The method is sync, not async.

Also: I wrapped the affected lines for easier reading in narrow views. And I put each affected sentence on a new line so diffs for trivial changes like this won't affect entire paragraphs in the future.

### DIFF
--- a/lib/Zonemaster/Engine/Overview.pod
+++ b/lib/Zonemaster/Engine/Overview.pod
@@ -29,7 +29,11 @@ messages to see, and it's entirely synchronous.
 The code that started the test does not get a chance to do anything at
 all until the whole test has finished, which may be several minutes later.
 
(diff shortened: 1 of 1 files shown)


lib/Zonemaster/Engine/Overview.pod
```diff
@@ -24,9 +24,16 @@ The simple-but-inflexible way is that all the methods in L<Zonemaster::Engine> t
...
+The main drawbacks of this method are that there is no choice about what
+messages to see, and it's entirely synchronous.
+The code that started the test does not get a chance to do anything at
+all until the whole test has finished, which may be several minutes later.
+
+To get around those drawbacks there is the flexible-but-complex way,
+which consists of installing a callback that gets executed every time
+a message is logged.
+It's not that much more complicated, code-wise.
+The following example does roughly the same thing as the one above:
```
matsduf: "The following example does roughly the same thing as the one above (besides more verbose printing to standard output):"

mattias-p: Please don't demand fixing of problems that are outside of the scope of the PR.

PR #376, opened 2018-02-06
Docs for profile class from scratch

mattias-p:
This is a second attempt to fix the Zonemaster Engine part of dotse/zonemaster#498. The first attempt was #361. The primary difference in this attempt is in the approach. Here a new module interface is created from scratch, while the previous PR tried to achieve the same goal by modifying the interface of an existing module. I believe the resulting interface in this PR is clearer than the one in the previous PR. If everyone agrees that the interface described here is better than that in #361, I'll close that PR.

My goal is to document Engine's behavior around profiles. I've approached this by designing an interface that I believe can be easily implemented in Engine, and easily used by both CLI and Backend to implement their respective profile behaviors.

@mtoma Can you review this API specifically looking at whether its a suitable basis for implementing dotse/zonemaster-backend#332.
@vlevigneron Can you review this API specifically looking at whether it would be easy to implement in the Engine?

Please raise anything you would like to have changed about it. E.g. if it's too vague or too specific in general or in some parts or some sense, or if it clashes somehow with how you guys would design it. Or even if you disagree with this whole approach.

### DIFF
--- a/docs/Profiles.md
+++ b/docs/Profiles.md
@@ -2,16 +2,14 @@
 
 ## Default profile
 
-N.B. At the time of writing there is no authoritative documentation for
-the default value of this property.
-For the time being, please refer to the file
-located by L<dist_file("Zonemaster-Engine",
-"default.profile")|File::ShareDir/dist_file>.
+At the time of writing there is no authoritative documentation for the
+default value of this property.
+For the time being, please refer to the file located by [dist_file(
+"Zonemaster-Engine", "default.profile" )].
 
 ## Creating profiles
 
-N.B. At the time of writing there is no authoritative documentation for
-the default value of this property.
+For a complete example of a profile JSON file, refer to the file located
+by [dist_file( "Zonemaster-Engine", "default.profile" )].
 
-located by L<dist_file("Zonemaster-Engine",
-"default.profile")|File::ShareDir/dist_file>.
+[dist_file( "Zonemaster-Engine", "default.profile" )]: https://metacpan.org/pod/File::ShareDir#dist_file
--- a/docs/Profiles.md
+++ b/docs/Profiles.md
@@ -2,8 +2,11 @@
 
 ## Default profile
 
+The default profile is documented in the [profile properties] section
+of the Zonemaster::Engine::Profile module.
+
 At the time of writing there is no authoritative documentation for the
-default value of this property.
+default value of the `test_levels` property.
 For the time being, please refer to the file located by [dist_file(
 "Zonemaster-Engine", "default.profile" )].
 
@@ -13,3 +16,4 @@ For a complete example of a profile JSON file, refer to the file located
 by [dist_file( "Zonemaster-Engine", "default.profile" )].
 
 [dist_file( "Zonemaster-Engine", "default.profile" )]: https://metacpan.org/pod/File::ShareDir#dist_file
+[Profile properties]: https://metacpan.org/pod/Zonemaster::Engine::Profile#PROFILE-PROPERTIES
--- a/docs/Profiles.md
+++ b/docs/Profiles.md
@@ -1,0 +1,17 @@
+# Profiles
+
+## Default profile
+
+N.B. At the time of writing there is no authoritative documentation for
+the default value of this property.
+For the time being, please refer to the file
+located by L<dist_file("Zonemaster-Engine",
+"default.profile")|File::ShareDir/dist_file>.
+
+## Creating profiles
+
+N.B. At the time of writing there is no authoritative documentation for
+the default value of this property.
+
+located by L<dist_file("Zonemaster-Engine",
+"default.profile")|File::ShareDir/dist_file>.
--- a/docs/Profiles.md
+++ b/docs/Profiles.md
@@ -7,13 +7,11 @@ of the Zonemaster::Engine::Profile module.
 
 At the time of writing there is no authoritative documentation for the
 default value of the `test_levels` property.
-For the time being, please refer to the file located by [dist_file(
-"Zonemaster-Engine", "default.profile" )].
+For the time being, please refer to [default.profile].
 
 ## Creating profiles
 
-For a complete example of a profile JSON file, refer to the file located
-by [dist_file( "Zonemaster-Engine", "default.profile" )].
+For a complete example of a profile JSON file, refer to [default.profile].
 
-[dist_file( "Zonemaster-Engine", "default.profile" )]: https://metacpan.org/pod/File::ShareDir#dist_file
+[default.profile]: ../share/default.profile
 [Profile properties]: https://metacpan.org/pod/Zonemaster::Engine::Profile#PROFILE-PROPERTIES
--- a/lib/Zonemaster/Engine/Profile.pm
+++ b/lib/Zonemaster/Engine/Profile.pm
@@ -28,20 +28,21 @@ be performed, and how the results are to be analyzed.
 For details on available properties see the L</PROFILE PROPERTIES>
 section.
 
-Here is an example that resets the effective profile to the default
-profile values.
+Here is an example for updating the effective profile with values from
+a given file and setting all properties not mentioned in the file to
+default values.
+For details on the file format see the L</JSON REPRESENTATION> section.
 
     use Zonemaster::Engine::Profile;
 
-    my $default = Zonemaster::Engine::Profile->default;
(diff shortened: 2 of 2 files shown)


lib/Zonemaster/Engine/Profile.pm
```diff
@@ -0,0 +1,331 @@
...
+
+Returns C<$self>.
+
+Dies if the given property name is invalid.
+
+Dies if the value is invalid for the given property.
+
+=head2 merge
+
+Merge this profile data of another profile into this one.
```
matsduf: "the profile data"?

lib/Zonemaster/Engine/Profile.pm
```diff
@@ -0,0 +1,331 @@
...
+
+    $profile1->merge( $profile2 );
+
+Properties from the other profile take precedence when the same property
+name exists in both profiles.
+The other profile object remains unmodified.
```
matsduf: "the other profile object" --> "the other profile properties"?

I find it unclear how merging is done. How do I merge a profile from a selected file on the "effective profile"? When mergining, which takes precedence over which?

lib/Zonemaster/Engine/Profile.pm
```diff
@@ -0,0 +1,331 @@
...
+=head1 PROFILE PROPERTIES
+
+Each property has a name and is either set or unset.
+If it is set it has a value that is valid for that specific property.
+Here is a listing of all the properties and their respective sets of
+valid values.
+
+=head2 resolver.defaults.usevc
+
+A boolean. If C<true>, only use TCP. Default C<false>.
```
matsduf: Default does not come from the code, but from the Engine default profile. Right? It should maybe be stated.

lib/Zonemaster/Engine/Profile.pm
```diff
@@ -0,0 +1,331 @@
...
+
+=head2 resolver.defaults.igntc
+
+A boolean. If C<false>, UDP queries that get responses with the C<TC>
+flag set will be automatically resent over TCP. Default C<false>.
+
+=head2 resolver.source
+
+The source address all resolver objects should use when sending queries,
+if one is set.
```
matsduf: If unset, the OS default address is used.

lib/Zonemaster/Engine/Profile.pm
```diff
@@ -0,0 +1,331 @@
...
+to run of all of its test cases.
+Test cases blacklisted here can still be run individually.
+The test cases C<basic00>, C<basic01> and C<basic02> cannot be blacklisted
+this way.
+The reason these particular test cases cannot be blacklisted is that part
+of their function is to verify that the given name can be tested at all.
+
+The keys of this hash are names of test cases from the test
+specifications.
+Only test cases mapped to C<false> are considered.
```
matsduf: "...are considered (blacklisted)".

lib/Zonemaster/Engine/Profile.pm
```diff
@@ -0,0 +1,331 @@
...
+Test cases mapped to C<true> are ignored.
```
matsduf: "Test cases not listed or mapped to C<true> are ignored (not blacklisted)."

mattias-p:
@matsduf I've addressed you're comments, though I acknowledge that the default values for _all_ profile properties should be clear from the documentation. This is not the case at present.

matsduf:
@mtoma and @vlevigneron, can you review this PR?

mattias-p:
There are two profile properties whose default values are still unspecified in this documentation: `asnroots` and `test_levels`.

lib/Zonemaster/Engine/Profile.pm
```diff
@@ -0,0 +1,355 @@
...
+modules emit.
+The values of the second level hashes are mapped to severity levels.
+The default severity level is C<DEBUG> for tags not found in the
+C<test_levels> item.
+
+N.B. At the time of writing there is no authoritative documentation for
+the default value of this property.
+For the time being, please refer to the file
+located by L<dist_file("Zonemaster-Engine",
+"default.profile")|File::ShareDir/dist_file>.
```
matsduf: Here we have a reference to the file containing the values for the Engine default profile.

1. It should be stated that this is not just some file, it is the default profile for Engine.
2. This should not be put in a foot note. This should be lifted up. This is an important file that not only lists the defaults, but also will work as a model for an alternative profile.

mattias-p: I don't see how this would be an improvement from an interface point of view. What use case does it serve? We shouldn't be adding things to the public API just for the sake of it. Keeping its surface area small helps maintenance.

I agree that the property values of the default profile should be documented. As a compromise since we don't have any documentation for the test levels at the moment I referred to this for documentation purposes. Once we have proper documentation for the test levels, that documentation should be pointed to instead.

matsduf: The use case is simply that the user can understand what role that file has, and also understand where the default configuration comes from. The default profile is an important file. I cannot see the problem of having it clearly documented here.

mattias-p: API-wise the existence of that file is completely irrelevant. It is a (potential) implementation detail.

The point of the API documentation is to describe the interface its behavior. It is out of place in the API documentation to describe implementation details. It's just noise and a burden of maintenance. People who are interested in the implementation are encouraged to look in the source.

matsduf: I cannot see it as noise. The burden is neglectable. The information will remain true as long as the code is not updated. As long as we do not have any other reasonable place for that information...

mattias-p: We can find a better place. I suggest the MANIFEST, it allows for comments.

But in any case I don't think this is the time for this discussion. We should be discussing the interface to be implemented - not the implementation.

matsduf: No, the MANIFEST file is not a good place. You can call it what you want, but this is not implementation, this is part of defining the interface in a broad sense.

mattias-p: No, the documentation is the authority. If the default() method returns anything else it's a bug.

sandoche2k: I agree with @matsduf that all information regarding the default profile should be here.

sandoche2k:
@mattias-p my advice will be to wait for approval from at least one reviewer. Provide a time frame for yourselves and if you do not get any approval from others, you merge the document.

docs/Profiles.md
```diff
@@ -0,0 +1,17 @@
+# Profiles
+
+## Default profile
+
+N.B. At the time of writing there is no authoritative documentation for
+the default value of this property.
+For the time being, please refer to the file
+located by L<dist_file("Zonemaster-Engine",
```
matsduf: This looks like POD formatting. It does not work in MD.

docs/Profiles.md
```diff
@@ -0,0 +1,17 @@
...
+"default.profile")|File::ShareDir/dist_file>.
+
+## Creating profiles
+
+N.B. At the time of writing there is no authoritative documentation for
+the default value of this property.
+
+located by L<dist_file("Zonemaster-Engine",
```
matsduf: This looks like POD formatting. It does not work in MD.

matsduf: I will approve when these to minor formatting errors are corrected.

docs/Profiles.md
```diff
@@ -0,0 +1,19 @@
...
+
+## Default profile
+
+The default profile is documented in the [profile properties] section
+of the Zonemaster::Engine::Profile module.
+
+At the time of writing there is no authoritative documentation for the
+default value of the `test_levels` property.
+For the time being, please refer to the file located by [dist_file(
+"Zonemaster-Engine", "default.profile" )].
```
matsduf: That link does not work. Add a link to the Github repository.

docs/Profiles.md
```diff
@@ -0,0 +1,19 @@
...
+
+## Creating profiles
+
+For a complete example of a profile JSON file, refer to the file located
+by [dist_file( "Zonemaster-Engine", "default.profile" )].
+
```
matsduf: That link does not work. Add a link to the Github repository.

docs/Profiles.md
mattias-p: It works for me.

PR #395, opened 2018-06-01
Delegation01: Rename messages

mattias-p:
Partial implementation of dotse/zonemaster#609. 
Refactors delegation01 method a bit, and renames a couple of messages.


mattias-p:
@sandoche2k can you review this?

PR #396, opened 2018-06-07
Delegation01: Add checks for child addresses

mattias-p:
Partial implementation of dotse/zonemaster#609.
Adds new checks that are specific to the in-zone name servers.


lib/Zonemaster/Engine/Test/Delegation.pm
```diff
@@ -170,6 +188,41 @@ sub delegation01 {
           info( NOT_ENOUGH_NS_CHILD => $child_nsnames_args );
     }
 
+    # Determine child NS addresses
+    my @child_addresses = map { $_->address } @{ Zonemaster::Engine::TestMethods->method4( $zone ) };
+    my @child_ipv4 = grep { $_->version == 4 } @child_addresses;
```
matsduf: As I interpret the code, `@child_ipv4` will hold two IP addresses if the zone is delegated to two nameserver, of which one resolves into two IPv4 addresses and one into no IPv4 address. Is that correct interpretation?

lib/Zonemaster/Engine/Test/Delegation.pm
```diff
@@ -170,6 +188,41 @@ sub delegation01 {
...
+    };
+    my $child_ipv6_args = {
+        count   => scalar( @child_ipv6 ),
+        minimum => $MINIMUM_NUMBER_OF_NAMESERVERS,
+        addrs   => join( q{;}, map { $_->short } sort { $a->binip cmp $b->binip } @child_ipv6 ),
+    };
+
+    if ( scalar( @child_ipv4 ) >= $MINIMUM_NUMBER_OF_NAMESERVERS ) {
+        push @results, info( ENOUGH_IPV4_NS_CHILD => $child_ipv4_args );
+    }
```
matsduf: The specification in dotse/zonemaster#609 refers to the number of name servers, not the number of IP addresses:

>  9. Count the number of name server names that resolve into at least one IPv4 
    address:
    1. If zero, emit *NO_IPV4_NS_CHILD*.
    2. If one, emit *NOT_ENOUGH_IPV4_NS_CHILD*.
    3. If two or more, emit *ENOUGH_IPV4_NS_CHILD*.

(From https://raw.githubusercontent.com/matsduf/zonemaster/3ac46bfbd45731b7dc2609ea21889f78ace03022/docs/specifications/tests/Delegation-TP/delegation01.md)

lib/Zonemaster/Engine/Test/Delegation.pm
mattias-p: Yes.

lib/Zonemaster/Engine/Test/Delegation.pm
mattias-p: I interpret the specification to be organized as series of "obtain data", "examine that data" cycles. Step 8 states that you should obtain a set of addresses. Apparently I didn't read step 9 carefully enough. More information than a mere set of addresses is needed to make the determination in step 9. Step 8 should be updated. Maybe we need a new method that returns a set of (name, address) pairs?

mattias-p:
Rebased resolving conflicts.

PR #398, opened 2018-06-07
Validation of some function arguments

mattias-p:
I didn't realize it that trailing dots were significant in a particular function argument. As a consequence I ended up debugging a faulty argument to this function for far too long. I don't want that to happen again (to me or anyone else) so I'm making it fail early instead of giving weird behaviors later on.

It seems a little weird that the domain argument should require the dot and that the href values should forbid the dot, but I'm following the precedence of `undelegated.t` which seems authoritative enough.


mattias-p:
#400 resolves this issue in the opposite direction. Apparently undelegated.t wasn't a good reference point.

PR #408, opened 2018-06-26
New test case Consistency06

mattias-p:
Implements zonemaster/zonemaster#619.


lib/Zonemaster/Engine/Test/Consistency.pm
```diff
@@ -102,6 +105,14 @@ sub metadata {
...
+        consistency06 => [
+            qw(
+              NO_RESPONSE
+              NO_RESPONSE_SOA_QUERY
+              ONE_SOA_MNAME
+              MULTIPLE_SOA_MNAMES
+              )
+        ],
     };
 } ## end sub metadata
```
matsduf: The messages should also be added to `sub translation{}` or is that deferred to a later step?

mattias-p: Good catch!

PR #413, opened 2018-07-02
Remove unused message and added used message in Basic.pm

matsduf:
Basic03: Removed message not used and added message used.

### DIFF
--- a/lib/Zonemaster/Engine/Test/Basic.pm
+++ b/lib/Zonemaster/Engine/Test/Basic.pm
@@ -104,7 +104,7 @@ sub metadata {
         ],
         basic03 => [
             qw(
-              NO_NAMESERVER_PREVENTS_WWW_A_TEST
+              NO_A_RECORDS
               HAS_A_RECORDS
               IPV4_DISABLED
               IPV6_DISABLED


lib/Zonemaster/Engine/Test/Basic.pm
```diff
@@ -104,7 +104,7 @@ sub metadata {
         ],
         basic03 => [
             qw(
-              NO_NAMESERVER_PREVENTS_WWW_A_TEST
```
mattias-p: Are you sure this removal is the right thing? It's emitted from all() in a context that's related to Basic03.

mattias-p:
Do you think you could update the PR title to be more descriptive?

lib/Zonemaster/Engine/Test/Basic.pm
matsduf: If you mean the following, it is a different message:

```perl
            push @results,
              info(
                HAS_NAMESERVER_NO_WWW_A_TEST => {
                    zname => $zone->name,
                }
              );
```
It is only mentioned once in the repo:

```sh
$ find . -type f -exec grep -Hi NO_NAMESERVER_PREVENTS_WWW_A_TEST {} \;
./lib/Zonemaster/Engine/Test/Basic.pm:              NO_NAMESERVER_PREVENTS_WWW_A_TEST
```

PR #414, opened 2018-07-02
Removed unused messages and added messages missing in policy.json

matsduf:
Removed message not used anywhere and added messages in use, but missing in policy.json.


matsduf:
share/policy.json does not exist anymore. The changes should be checked against share/profile.json.

matsduf:
Replaced by #464.

PR #433, opened 2018-08-23
Emit SOA_SERIAL for every serial

mattias-p:
Even when a single serial is detected.

This makes it easier to collect the serial numbers, as you only ever
need to look for a single message tag.

This should be a low-risk change. I even kept the serial argument for ONE_SOA_SERIAL in order to be conservative, but I'll gladly remove that on request.


mattias-p:
@sandoche2k I'm glad you like it. Could you either approve the PR or invite @bombsimon so he can approve it? Without formal approval from a contributor I'm unable to merge it.

PR #447, opened 2018-11-14
Profiles tweaks

mattias-p:
I've described some details about most of the changes in the respective commit messages.

I took the liberty to rename load() to from_json(). I think it's more fitting.

### DIFF
--- a/lib/Zonemaster/Engine/Profile.pm
+++ b/lib/Zonemaster/Engine/Profile.pm
@@ -170,7 +170,7 @@ profile JSON file.
 
 =head2 resolver.defaults.retrans
 
-A number. The number of seconds between retries. Default 3.
+A non-negative integer. The number of seconds between retries. Default 3.
 
 =head2 resolver.defaults.dnssec
 
--- a/lib/Zonemaster/Engine/Profile.pm
+++ b/lib/Zonemaster/Engine/Profile.pm
@@ -35,8 +35,9 @@ For details on the file format see the L</JSON REPRESENTATION> section.
 
     use Zonemaster::Engine::Profile;
 
-    my $foo     = Zonemaster::Engine::Profile->load( "/path/to/foo.profile" );
-    my $profile = Zonemaster::Engine::Profile->default;
+    my $profile_json = read_file( "/path/to/foo.profile" );
+    my $foo          = Zonemaster::Engine::Profile->from_json( $profile_json );
+    my $profile      = Zonemaster::Engine::Profile->default;
     $profile->merge( $foo );
     Zonemaster::Engine::Profile->effective->merge( $profile );
 
--- a/lib/Zonemaster/Engine/Profile.pm
+++ b/lib/Zonemaster/Engine/Profile.pm
@@ -166,7 +166,7 @@ profile JSON file.
 
 =head2 resolver.defaults.usevc
 
-A boolean. If C<true>, only use TCP. Default C<false>.
+1 or 0. If C<1>, only use TCP. Default C<0>.
 
 =head2 resolver.defaults.retrans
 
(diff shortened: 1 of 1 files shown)


lib/Zonemaster/Engine/Profile.pm
```diff
@@ -166,21 +161,21 @@ profile JSON file.
 
 =head2 resolver.defaults.usevc
 
-A boolean. If C<true>, only use TCP. Default C<false>.
+1 or 0. If C<1>, only use TCP. Default C<0>.
```
matsduf: Why change from true/false to 1/0?

lib/Zonemaster/Engine/Profile.pm
```diff
@@ -166,21 +161,21 @@ profile JSON file.
...
 
 =head2 resolver.defaults.retrans
 
-A number. The number of seconds between retries. Default 3.
+A non-negative integer. The number of seconds between retries. Default 3.
```
matsduf: Maybe a positive integer? Should zero be permitted?

matsduf:
@vlevigneron, I think the idea was that there should be a helper method to load from file to JSON string, and then use "from_json". That's an extra step i e.g. CLI, but simple to implement.

The helper method should verify that the result is valid JSON, but not necessarily valid policy.

lib/Zonemaster/Engine/Profile.pm
matsduf: An integer between (and including) 1 and 240 is reasonable. Default 3 is fine.

lib/Zonemaster/Engine/Profile.pm
```diff
@@ -166,21 +161,21 @@ profile JSON file.
...
 
 =head2 resolver.defaults.recurse
 
-A boolean. If C<true>, sets the RD flag in queries. Default C<false>.
+1 or 0. If C<1>, sets the RD flag in queries. Default C<0>.
 
-This should almost certainly be kept C<false>.
+This should almost certainly be kept C<0>.
 
 =head2 resolver.defaults.retry
```
matsduf: The resolver.defaults.retry should be a positive integer between (and including) 1 and 100. The default of 2 is fine.

lib/Zonemaster/Engine/Profile.pm
mattias-p: I was trying to take a shortcut here but I failed to consider the consequences with regard to JSON. I'm making an update that takes both Perl and JSON representations into consideration.

lib/Zonemaster/Engine/Profile.pm
mattias-p: Updating to positive integer.

mattias-p:
> The "load_policy_file" method should also be specified.

I was assuming that the load_policy_file() helper should live in Zonemaster::Engine::Util.

lib/Zonemaster/Engine/Profile.pm
mattias-p: Updating to integer between 1 and 255. LDNS uses uint8 and zero doesn't make sense.

lib/Zonemaster/Engine/Profile.pm
```diff
@@ -336,22 +334,19 @@ level for the tag.
 
 =head2 test_cases
 
-A hashref mapping test case names to booleans. Default C<{}>.
+An arrayref of names of test cases as listed in the L<test case specifications|
+https://github.com/zonemaster/zonemaster/tree/master/docs/specifications/tests>.
```
matsduf: "...as listed in the L<test case list|
https://github.com/zonemaster/zonemaster/tree/master/docs/specifications/tests/TestCases.md>"

File to be created before we release, see zonemaster/zonemaster#649

matsduf: I will create a separte PR on this change.

PR #449, opened 2018-11-19
Profiles tests

mattias-p:
A unit test mirroring the Profile pod. I've fixed it in two iterations and I believe it to be mostly correct now. 

It is a bit lax on validation of the complex property values. That should probably be tightened (including the pod) so we get the ability to give users prompt feedback on erroneous configurations.

### DIFF
--- a/t/profiles.t
+++ b/t/profiles.t
@@ -16,20 +16,20 @@ Readonly my $EXAMPLE_PROFILE_1 => q(
 {
   "resolver": {
     "defaults": {
-      "usevc": 1,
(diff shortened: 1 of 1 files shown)


vlevigneron:
I just pushed a new version of https://github.com/zonemaster/zonemaster-engine/pull/446, there is no need anymore to skip tests. All the tests succeed with this new version.

mattias-p:
This was merged into https://github.com/vlevigneron/zonemaster-engine/pull/1 instead.

PR #458, opened 2018-12-10
Make it possible to `use Zonemaster::Engine` in development context

### LINKED ISSUE #457: Make it possible to `use Zonemaster::Engine` in development context
Today you can't `use Zonemaster::Engine` unless it's installed. I.e. you cannot use `perl -I` to point out a development version of Engine. This is inconvenient when you want to develop Engine and CLI on the same machine.

If the Zonemaster-Engine distribution isn't installed, dist files should be sought under the source repo instead of the installed dist file directory.

mattias-p:
Fixes #457.


sandoche2k:
@matsduf @vlevigneron review pls.

mattias-p:
I don't like this approach. It's fragile and adds code to production environments that doesn't belong there. Also the implementation doesn't work.

PR #464, opened 2019-01-09
Updated profile.json files

matsduf:
Removed message not used anywhere and added messages in use, but missing in three profile.json files:

* share/profile.json
* share/profile_example_logfilter.json
* t/profiles/Test-all-levels.json

Replaces and updates #414.


matsduf:
@vlevigneron, please review this PR instead.

Issue #469, opened 2019-01-10
A Perl 5.20 specific issue in zonemaster-engine


mtoma:
The new profiles branch has an issue with some JSON value specific to the version of JSON in Perl 5.20 as can be seen in:
https://travis-ci.org/zonemaster/zonemaster-backend/builds/477783002

Error:  Property resolver.defaults.fallback is of type Boolean at /tmp/local-lib-pZVpnQ/lib/perl5/Zonemaster/Engine/Profile.pm line 225

The same code passes the tests with all other versions of Perl.

In previous tests the same error was showing up for ip_v4 and ip_v6 variables. They were set to 1, setting them to JSON::PP::true fixed the problem.

This is not feasible for the resolver.defaults.fallback variable as it is not used explicitly in the zonemaster-backend.

vlevigneron:
I'm not able to reproduce it.
```
> perl -v

This is perl 5, version 20, subversion 3 (v5.20.3) built for x86_64-linux
(with 1 registered patch, see perl -V for more detail)
[...]

> perl -I./lib ./t/test_DB_backend.pl PostgreSQL
ok 1 - Testing a supported database backend: PostgreSQL
ok 2 - use Zonemaster::Backend::RPCAPI;
ok 3 - An object of class 'Zonemaster::Backend::RPCAPI' isa 'Zonemaster::Backend::RPCAPI'
[...]
```

vlevigneron:
It seems that nobody can reproduce the issue which is still present in Travis. First I'll propose a patch that provide a more detailed output in case of die. It could help to understand what's wrong and what should be changed in the code to deal with that issue.

vlevigneron:
The issue should be fixed by https://github.com/zonemaster/zonemaster-engine/pull/472

sandoche2k:
@mtoma could be closed ?

PR #471, opened 2019-01-11
- Add more details in case of crash for profile properties.

vlevigneron:
In order to have more details to fix travis issue in Backend (https://travis-ci.org/zonemaster/zonemaster-backend/jobs/477783008) we need to have more details in case of crash.

Unless someone is able to reproduce the bug, we need this patch in order to fix  https://github.com/zonemaster/zonemaster-engine/issues/469

### DIFF
--- a/Makefile.PL
+++ b/Makefile.PL
@@ -32,6 +32,7 @@ requires 'MIME::Base64'          => 0;
 requires 'Clone'                 => 0;
 requires 'JSON'                  => 0;
 requires 'JSON::PP'              => 0;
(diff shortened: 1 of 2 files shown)


vlevigneron:
@matsduf We could decide to keep it but I know we avoid to add new dependencies (in this case Data::Dumper). Personaly, I would keep it but the first intention is to have it to fix the issue and if we do not want to keep it, remove it just after.

matsduf:
@vlevigneron, let us talk about it. I just wanted to raise the question.

PR #472, opened 2019-01-11
- Add true/false processing for Booleans.

vlevigneron:
Should fix backend issue about boolean properties in Profile.
Unit tests (t/profiles.t) has been adapted with new behaviour.

### DIFF
--- a/lib/Zonemaster/Engine/Profile.pm
+++ b/lib/Zonemaster/Engine/Profile.pm
@@ -1,6 +1,6 @@
 package Zonemaster::Engine::Profile;
 
-use version; our $VERSION = version->declare("v1.2.11");
+use version; our $VERSION = version->declare("v1.2.12");
 
 use 5.014002;
 use strict;
@@ -211,6 +211,7 @@ sub _set {
                             defined $value_type ? $value_type : q{UNDEF},
                             defined $value ? $value : q{[UNDEF]},
                             Data::Dumper::Dumper($value);
+			    #printf "%s : %s\n", $property_name, $data_details;
     # $value is a Scalar
     if ( ! $value_type  or $value_type eq q{SCALAR} ) {
         die "Property $property_name can not be undef" if not defined $value;
@@ -227,6 +228,12 @@ sub _set {
                 $value = JSON::PP::false;
             }
             elsif ( $from eq q{JSON} and $value_type and $value == JSON::PP::true ) {
+                $value = JSON::PP::true;
+            }
+            elsif ( $from eq q{JSON} and $value =~ /^0$/ ) {
+                $value = JSON::PP::false;
+            }
+            elsif ( $from eq q{JSON} and $value =~ /^1$/ ) {
                 $value = JSON::PP::true;
             }
             else {
--- a/lib/Zonemaster/Engine/Profile.pm
+++ b/lib/Zonemaster/Engine/Profile.pm
@@ -1,6 +1,6 @@
 package Zonemaster::Engine::Profile;
 
-use version; our $VERSION = version->declare("v1.2.14");
+use version; our $VERSION = version->declare("v1.2.15");
 
 use 5.014002;
 use strict;
@@ -217,10 +217,10 @@ sub _set {
 
         # Boolean
         if ( $profile_properties_details{$property_name}->{type} eq q{Bool} ) {
-            if ( $from eq q{DIRECT} and ($value =~ /^0$/ or $value eq q{false}) ) {
+            if ( $from eq q{DIRECT} and ($value eq q{0} or $value eq q{false}) ) {
                 $value = JSON::PP::false;
             }
-            elsif ( $from eq q{DIRECT} and ($value =~ /^1$/ or $value eq q{true}) ) {
+            elsif ( $from eq q{DIRECT} and ($value eq q{1} or $value eq q{true}) ) {
                 $value = JSON::PP::true;
             }
             elsif ( $from eq q{JSON} and $value_type and $value == JSON::PP::false ) {
--- a/lib/Zonemaster/Engine/Profile.pm
+++ b/lib/Zonemaster/Engine/Profile.pm
@@ -1,6 +1,6 @@
 package Zonemaster::Engine::Profile;
 
-use version; our $VERSION = version->declare("v1.2.13");
+use version; our $VERSION = version->declare("v1.2.14");
 
 use 5.014002;
 use strict;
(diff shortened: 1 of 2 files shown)


sandoche2k:
@matsduf @mtoma do you approve?

t/profiles.t
```diff
@@ -216,17 +222,19 @@ subtest 'from_json() dies on illegal paths' => sub {
 };
 
 subtest 'from_json() dies on illegal values' => sub {
-    dies_ok { Zonemaster::Engine::Profile->from_json( '{"resolver":{"defaults":{"usevc":0}}}' ); }
+    dies_ok { Zonemaster::Engine::Profile->from_json( '{"resolver":{"defaults":{"usevc":}}}' ); }
```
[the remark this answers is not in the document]
vlevigneron: This one is a typo, I'll fixed it.

lib/Zonemaster/Engine/Profile.pm
```diff
@@ -215,12 +215,12 @@ sub _set {
...
 
-        # Boolean (Should we accept (true|false) values ?
+        # Boolean
         if ( $profile_properties_details{$property_name}->{type} eq q{Bool} ) {
-            if ( $from eq q{DIRECT} and $value =~ /^0$/ ) {
+            if ( $from eq q{DIRECT} and ($value =~ /^0$/ or $value eq q{false}) ) {
                 $value = JSON::PP::false;
             }
-            elsif ( $from eq q{DIRECT} and $value =~ /^1$/ ) {
+            elsif ( $from eq q{DIRECT} and ($value =~ /^1$/ or $value eq q{true}) ) {
```
matsduf: Should really strings "0" and "1" be accepted as booleans?

vlevigneron: Yes. While we are more permissive for input data, we are strict and more homogenous for output data.

lib/Zonemaster/Engine/Profile.pm
```diff
@@ -215,12 +215,12 @@ sub _set {
     if ( ! $value_type  or $value_type eq q{SCALAR} ) {
         die "Property $property_name can not be undef" if not defined $value;
 
-        # Boolean (Should we accept (true|false) values ?
+        # Boolean
         if ( $profile_properties_details{$property_name}->{type} eq q{Bool} ) {
-            if ( $from eq q{DIRECT} and $value =~ /^0$/ ) {
+            if ( $from eq q{DIRECT} and ($value =~ /^0$/ or $value eq q{false}) ) {
```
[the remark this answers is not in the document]
vlevigneron: It's just another way to do this kind of test, but for the consistency I can use `eq ` syntax.

t/profiles.t
```diff
@@ -449,13 +465,14 @@ subtest 'set() dies if the given property name is invalid' => sub {
...
-    dies_ok { $profile->set( 'net.ipv6',                  'false' ); } 'checks type of net.ipv6';
-    dies_ok { $profile->set( 'no_network',                'false' ); } 'checks type of no_network';
+    dies_ok { $profile->set( 'resolver.defaults.usevc',    'folse' ); } 'checks type of resolver.defaults.usevc';
+    dies_ok { $profile->set( 'resolver.defaults.dnssec',   'folse' ); } 'checks type of resolver.defaults.dnssec';
+    dies_ok { $profile->set( 'resolver.defaults.recurse',  'folse' ); } 'checks type of resolver.defaults.recurse';
+    dies_ok { $profile->set( 'resolver.defaults.igntc',    'folse' ); } 'checks type of resolver.defaults.igntc';
+    dies_ok { $profile->set( 'resolver.defaults.fallback', 'folse' ); } 'checks type of resolver.defaults.fallback';
+    dies_ok { $profile->set( 'net.ipv4',                   'folse' ); } 'checks type of net.ipv4';
+    dies_ok { $profile->set( 'net.ipv6',                   'folse' ); } 'checks type of net.ipv6';
+    dies_ok { $profile->set( 'no_network',                 'folse' ); } 'checks type of no_network';
```
mattias-p: According to the documentation for `set()` any defined value should be accepted for boolean properties, and the value should be interpreted using the standard truthiness evaluation. So both the strings `"false"` and `"folse"` should in fact be accepted, and both should be interpreted to be true.

matsduf: For `set()` of boolean properties we could limit legal values to `1` for TRUE and `0` for FALSE. Then neither humans nor Perl would be confused.

mattias-p:
In Perl `"false"` evaluates to true no matter if it's confusing. If we create an API that makes an exception to this rule and interprets `"false"` as false, we'd only add to the confusion.

On the surface I like @matsduf's suggestion to only accept `1` and `0`, but on closer inspection this model also adds to the confusion. For example the following otherwise reasonable code would die since `eq` returns an empty string for false:
```perl
$profile->set( "no_network", "Alice" eq "Bob" );
```

I've really tried to think of a model for booleans in Perl that's free from surprises, but I've failed to come up with anything that's less confusing than just sticking with what Perl gives us.

FYI. Here's my usual go-to reference for truthiness is Perl: https://perlmaven.com/boolean-values-in-perl

t/profiles.t
mattias-p: In Perl `"false"` evaluates to true no matter if it's confusing. If we create an API that makes an exception to this rule and interprets `"false"` as false, we'd only add to the confusion.

On the surface I like @matsduf's suggestion to only accept `1` and `0`, but on closer inspection this model also adds to the confusion. For example the following otherwise reasonable code would die since `eq` returns an empty string for false:
```perl
$profile->set( "no_network", "Alice" eq "Bob" );
```

I've really tried to think of a model for booleans in Perl that's free from surprises, but I've failed to come up with anything that's less confusing than just sticking with what Perl gives us.

FYI. Here's my usual go-to reference for truthiness is Perl: https://perlmaven.com/boolean-values-in-perl

matsduf: I realize that my suggestion above was not optimal. We are discussing perl methods, not JSON. Using perl standard boolean is probably best, and not treating  the string "false" (or "falsk" (sv) or "faux" (fr)) as special. In the profile.json we should use JSON standard.

mattias-p: @vlevigneron The documentation and unit tests that I wrote regretfully disagree with each other on this point (the documentation being more permissive), so I guess it's not so clear which one is the goal. I'm arguing here that the documentation is a better goal than the unit test.

In particular the problem with interpreting `"false"` as false is that Perl defines it to be true, so then there would be two reasonable interpretations of this value, which is a source of confusion. If we do allow `"false"`, it should be interpreted the Perl way.

The `Profile` module implements two interfaces - one JSON interface and one Perl interface. One of its jobs is to isolate these two interfaces from each other. On the JSON side we use `true` and `false` because they are idiomatic there. On the Perl side we should use whatever is idiomatic in Perl.

That said I agree that we're getting derailed here. This PR started out to fix zonemaster/zonemaster-backend#403 but in the process of fixing that a discrepancy between documentation and unit test was discovered.

I'll accept this PR and then I'll make another one to bring things in line with the documentation.

vlevigneron:
@matsduf I updated the comment of PR about unit tests.

sandoche2k:
@vlevigneron merge

PR #473, opened 2019-01-11
Update resolver.source default value to match documentation

mattias-p:
The current develop branch implements resolver.source to determine its default value using an algorithm that assumes that the system has a hostname and that the hostname is resolvable to an address that is used by a network interface. If one of these assumptions is false, Engine is unable to send DNS requests with the default profile.

On the other hand, the documentation for resolver.source states that the default value should be `"os_default"` and the associated behavior should be to use the OS default.

This PR updates the implementation to match the documentation and adds corresponding unit tests for the Profile module.


sandoche2k:
@mattias-p merge?

PR #474, opened 2019-01-15
Align tests and impl for boolean properties with documentation

mattias-p:
There's been some confusion regarding what values to accept for booleans by the `Profile::set` method.

This PR updates the implementation and unit tests to match the documentation. In other words, any defined value is accepted, and the value is interpreted according to the Perl rules for boolean context.

Without this PR the string `"false"` is interpreted as false by `Profile::set` while at the same time being interpreted as true by Perl.


sandoche2k:
@mattias-p merge?

PR #487, opened 2019-01-24
- Fix zonemaster/zonemaster-backend#476

vlevigneron:
Don't know (now) how to fix this stringification issue ( use a ',' instead of '.' ). It works with that hack but needto be checks with various versions of Perl.


vlevigneron:
@mattias-p It's part of the plan to add Unit Test for that. It will be done later.

matsduf:
@vlevigneron, I really want to see documentation in the code explaining why the hack is there.

sandoche2k:
You can test in Centos via : http://149.202.56.242/domain_check

PR #490, opened 2019-01-25
installation: libmodule-install-perl is also required

fanf2:


sandoche2k:
If you are testing, please clone from the develop branch, all ZM modules, which has the recent updates w.r.t to EDNS and DNS flag day. We will be very grateful for your feedback.

matsduf:
@fanf2, did you install from CPAN or from source? I think module-install is only needed when you install from source, so if you run `sudo cpanm Zonemaster::LDNS` then you should not need it.

Please correct me if I am wrong.

matsduf:
Thank you for a detailed report. We will investigate.

sandoche2k:
@fanf2 is right. I was substituting earlier using cpan with inc::Module::Install.

mattias-p:
@fanf2 Thank you so much for contributing! And our apologies for taking so long.

PR #493, opened 2019-01-30
- Fix nameserver02 algorithm.

vlevigneron:
Related to Petr Špaček issue about flagday.rootcanary.net
Need review on translations missing for new error messages.


share/sv.po
```diff
@@ -801,6 +801,10 @@ msgstr "Namnservern {ns}/{address} tillåter zonöverföringar med AXFR."
 msgid  "AXFR not available on nameserver {ns}/{address}."
 msgstr "AXFR är inte tillgänglig från namnservern {ns}/{address}."
 
+#: NAMESERVER:BROKEN_EDNS_SUPPORT
+msgid  "Incorrect response from {ns}/{address} asking for {dname} when EDNS is used."
+msgstr "Incorrect response from {ns}/{address} asking for {dname} when EDNS is used."
```
matsduf: "Felaktigt DNS-svar från {ns}/{address} när {dname} efterfrågas med EDNS."

matsduf:
@vlevigneron and @sandoche2k, when I see the implementation I realize that we (or I) have, in the specification, lumped three different cases into one BROKEN_EDNS_SUPPORT. The implementation follows the specification, but I think that we should do an update before v2019.1 release. See zonemaster/zonemaster#672.

PR #496, opened 2019-02-06
Maintenance of dependencies

mattias-p:
Various updates related to dependencies. Mostly focused on CentOS. See commit messages for details.


mattias-p:
This is too messy. Breaking it down into smaller pieces.

PR #497, opened 2019-02-06
Maintenance of dependencies - part 1

mattias-p:
This will aid installation testing:
* By having fewer steps in the installation instruction,
* By making it easier to cut-and-paste the command for installing all CPAN dependencies except Zonemaster::LDNS (which is installed using a distribution tarball in this context), and
* By providing a slightly shorter command to cut-and-paste.

### DIFF
--- a/docs/Installation.md
+++ b/docs/Installation.md
@@ -30,25 +30,19 @@ This instruction covers the following operating systems:
    sudo yum groupinstall "Development Tools"
    ```
 
-2) Install cpan minus 
+2) Install binary packages:
 
    ```sh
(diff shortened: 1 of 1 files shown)


matsduf:
Do you want to simplify more? In FreeBSD inc::Module::Install is installed by package p5-Module-Install

mattias-p:
@matsduf, I've laid my eyes on that too. I intend to come back to that in a later PR. Depending on technical details we should either replace inc::Module::Install with p5-Module-Install (as you suggest) or remove it entirely from this installation instruction.

PR #498, opened 2019-02-07
Improve cache usage

vlevigneron:
- EDNS queries are now cached.
- upgrade all unit tests.
- It is also related to Memory Leaks issue zonemaster/zonemaster-backend#482
We are now ready to add new UT for last 'DNS flag day' tests.


vlevigneron:
@matsduf The mechanism itself has not changed. We only cache responses.

lib/Zonemaster/Engine/Nameserver.pm
```diff
@@ -215,16 +217,28 @@ sub query {
...
-        }
-        $p = $self->cache->data->{"$name"}{"\U$type"}{"\U$class"}{$dnssec}{$usevc}{$recurse}{$edns_size};
+    $md5->add( q{NAME}    , $name );
+    $md5->add( q{TYPE}    , "\U$type" );
+    $md5->add( q{CLASS}   , "\U$class" );
+    $md5->add( q{DNSSEC}  , $dnssec );
+    $md5->add( q{USEVC}   , $usevc );
+    $md5->add( q{RECURSE} , $recurse );
+    if ( $edns_special_case ) {
+        $md5->add( q{EDNS_VERSION}        , $href->{edns_details}{version} ? $href->{edns_details}{version} : 0 );
```
mattias-p: 1.  Have you considered documenting `edns_details`?
2.  Have you considered validating and/or normalizing the value of `$href->{edns_details}{version}` in a single place and/or documenting for which values the behavior is defined?

t/asn.t
```diff
@@ -17,8 +17,8 @@ is $asn1->[0], 15169, '8.8.8.8 is in 15169';
 is $prefix1->prefix, '8.8.8.0/24', '8.8.8.8 is in 8.8.8.0/24';
 
 my ( $asn2, $prefix2 ) = Zonemaster::Engine::ASNLookup->get_with_prefix( '91.226.36.46' );
-is $asn2->[0], 1257, '91.226.36.46 is in 1257';
-is $prefix2->prefix, '91.226.36.0/24', '91.226.36.46 is in 91.226.36.0/24';
+is $asn2->[0], 197564, '91.226.36.46 is in 197564';
+is $prefix2->prefix, '91.226.36.0/23', '91.226.36.46 is in 91.226.36.0/24';
```
mattias-p: These changes are here because the internet has changed, right?

matsduf:
With the same question (question section) but with different content in OPT in the query we will expect different responses. E.g. if the question is "example.com. SOA" and the EDNS version is 0 we will expect the SOA record in the response, but if the EDNS version is 1 we will expected the response to have RCODE FORMER. Will that work with caching?

vlevigneron:
@matsduf Yes, all EDNS parts are used to compute index of hash in cache, so similar queries with different EDNS content will have differents entries in the cache (wether or not the response is the same).

t/asn.t
vlevigneron: Yes, it is.

lib/Zonemaster/Engine/Nameserver.pm
vlevigneron: Yes we should add documentation, but it is not the purpose of that PR. If there is not already an issue for that, we should create one.

mattias-p: Personally I think it's much easier to update the code and documentation at the same time while I still have the thought process in my head. It's also much harder to make good code reviews when you don't have documentation that spells out the intention of the APIs. But go ahead and create an issue for the documentation if you feel that's better.

vlevigneron: I agree with you but in this case, this part of code should have been documented (`edns_details`) in previous release, when we added EDNS tests, amonst other things. `edns_details` had not been introduced in this PR. Documentation will be in a separate PR, that should be accepted without questions. And I was not sure that the current PR would be accepted.

matsduf:
@vlevigneron, what is "UT" in 
>     * We are now ready to add new UT for last 'DNS flag day' tests.

vlevigneron:
@matsduf UT => Unit Test

matsduf:
@vlevigneron, please create an issue for documenting the code. Please make the description of that issue explicit.

PR #502, opened 2019-02-12
 Maintenance of dependencies - part 2

mattias-p:
This will aid future maintenance of dependencies:
* By listing dependencies in alphabetical order,
* By removing unused modules,
* By removing core modules (except where we have minimum version requirements),
* By adding forgotten modules (all of `configure_requires`, `requires`, `test_requires` and `recommends`), and
* By preferring RPM to CPAN in the installation instruction.

Note that this PR only regards itself with the CentOS part of the installation instruction.

### DIFF
--- a/Makefile.PL
+++ b/Makefile.PL
@@ -29,7 +29,7 @@ requires 'Moose'                 => 2.0401;
 requires 'Net::IP'               => 1.26;
 requires 'Readonly'              => 0;
 requires 'Text::CSV'             => 0;
-requires 'Zonemaster::LDNS'      => 1.0;
+requires 'Zonemaster::LDNS'      => 2.0;
 
 test_requires 'Pod::Coverage'     => 0;
 test_requires 'Test::Differences' => 0;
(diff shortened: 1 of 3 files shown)


Makefile.PL
```diff
@@ -16,29 +16,25 @@ all_from 'lib/Zonemaster/Engine.pm';
...
 requires 'Readonly'              => 0;
-requires 'Locale::TextDomain'    => 1.23;
 requires 'Text::CSV'             => 0;
-requires 'MIME::Base64'          => 0;
-requires 'Clone'                 => 0;
-requires 'JSON'                  => 0;
-requires 'JSON::PP'              => 0;
-requires 'Data::Dumper'          => 0;
-requires 'Digest::MD5'           => 0;
+requires 'Zonemaster::LDNS'      => 1.0;
```
matsduf: Should rather be
```
requires 'Zonemaster::LDNS'      => 2.0000;
```

Issue #505, opened 2019-03-13
Test failure on questionable JSON serialization of profile


mattias-p:
```sh
$ perl -MJSON::PP -E 'say $JSON::PP::VERSION'
2.27105
$ PERL_DL_NONLAZY=1 /usr/bin/perl "-MExtUtils::Command::MM" "-e" "test_harness(0, 'inc', 'blib/lib', 'blib/arch')" t/profiles.t
t/profiles.t .. 26/29
        #   Failed test at t/profiles.t line 692.
        # +----+----------------------+--------------------+
        # | Elt|Got                   |Expected            |
        # +----+----------------------+--------------------+
        # |   0|{                     |{                   |
        # |   1|  resolver => {       |  resolver => {     |
        # |   2|    defaults => {     |    defaults => {   |
        # *   3|      retry => '123'  |      retry => 123  *
        # |   4|    }                 |    }               |
        # |   5|  }                   |  }                 |
        # |   6|}                     |}                   |
        # +----+----------------------+--------------------+
        # Looks like you failed 1 test of 1.

    #   Failed test 'resolver.defaults.retry'
    #   at t/profiles.t line 693.

        #   Failed test at t/profiles.t line 701.
        # +----+------------------------+----------------------+
        # | Elt|Got                     |Expected              |
        # +----+------------------------+----------------------+
        # |   0|{                       |{                     |
        # |   1|  resolver => {         |  resolver => {       |
        # |   2|    defaults => {       |    defaults => {     |
        # *   3|      retrans => '234'  |      retrans => 234  *
        # |   4|    }                   |    }                 |
        # |   5|  }                     |  }                   |
        # |   6|}                       |}                     |
        # +----+------------------------+----------------------+
        # Looks like you failed 1 test of 1.

    #   Failed test 'resolver.defaults.retrans'
    #   at t/profiles.t line 702.
    # Looks like you failed 2 tests of 16.

#   Failed test 'to_json() serializes each property'
#   at t/profiles.t line 758.
# Looks like you failed 1 test of 29.
t/profiles.t .. Dubious, test returned 1 (wstat 256, 0x100)
Failed 1/29 subtests

Test Summary Report
-------------------
t/profiles.t (Wstat: 256 Tests: 29 Failed: 1)
  Failed test:  27
  Non-zero exit status: 1
[...]

PR #506, opened 2019-03-14
Release 3.0.2

### LINKED ISSUE #507: Travis cannot build master branch with breaking changes in LDNS develop branch
Travis cannot build master branch with breaking changes in LDNS develop branch. This is a problem when we want to make a bug-fix release after introducing breaking changes in the LDNS develop branch. This problem was discovered in #506.

Reasonably the Engine master branch should depend on the LDNS master branch. I believe this could be solved by using the TRAVIS_BRANCH environment variable in the Travis config instead of hard-coding the develop branch. See https://docs.travis-ci.com/user/environment-variables/#default-environment-variables.

mattias-p:
With 3.0.1 a unit test fails if you have an older version of JSON::PP or if you have the PERL_JSON_PP_USE_B environment variable set. This fixes that (#505).

This also adds libmodule-install-perl to the Debian installation instruction. This change was mistakenly merged to the master branch instead of the develop branch.

Edit: I've added a commit that makes Travis use the LDNS master branch for PRs against the Engine master branch. I verified that it works using #508 (Engine master uses LDNS master) and #509 (Engine develop uses LDNS develop). Fixes #507.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,5 +11,5 @@ perl:
 before_install:
     - eval $(curl https://travis-perl.github.io/init) --auto
     - local-lib
-    - git clone --depth=1 --branch=develop https://github.com/zonemaster/zonemaster-ldns.git
+    - git clone --depth=1 --branch=$TRAVIS_BRANCH https://github.com/zonemaster/zonemaster-ldns.git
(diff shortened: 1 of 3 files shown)


mattias-p:
Travis can't build the Engine master branch against the LDNS develop branch because of breaking changes in LDNS.

Issue #504, opened 2019-03-13
Some nameservers are not resolved from Zonemaster, even though they were accessible


sandoche2k:
https://zonemaster.net/result/b2fe4826d3fc1244

For the domain "myriam-kassem.fr", zonemaster had a nameserver warning saying that NS (ns200.anycast.me/46.105.207.200) for the domain was not accessible. But, when a dig request was made from the ZM VPS, for the same NS, it was resolving. 

The issue was identified by @vlevigneron as an old code has been put in place to evict servers that were not responding.

sandoche2k:
Related to the BUG https://github.com/zonemaster/zonemaster-engine/issues/438

vlevigneron:
Well, here is what I found out.
We have a blacklisting mechanism that prevents query multiple times servers that do not respond.
The servers ns200.anycast.me/dns200.anycast.me are blacklisted because they answer REFUSED to a '.' query. It's their right.
The issue is that we have the zonemaster/zonemaster-engine#438 BUG that does not make a difference between unreachable and REFUSED responses.

IMHO, we have 3 options :
1/ Fix zonemaster/zonemaster-engine#438. As the problem is perhaps in LDNS and not in zonemaster code, it is not easy to fix in a short term.
2/ Inhibits for the moment blacklisting mechanism. Very easy but could slower processing in case of really broken zones.
3/ Adapt blacklisting mechanism. Did some tests, not sure it will be very effective.

sandoche2k:
@vlevigneron remove the blacklisting for now and close the issue. Also, create a new issue to put the blacklisting again.

matsduf:
Fixed by #511.

Issue #438, opened 2018-09-14
Cannot differentiate between REFUSED and no response


mattias-p:
Zonemaster::Engine::Nameserver::query returns `undef` for both DNS responses with status REFUSED and timeout (i.e. no response). In effect it's impossible for test cases to differentiate between those two conditions.

matsduf:
This requires a fix in LDNS.

mattias-p:
I cannot reproduce this. NAMESERVER01 emits "Erroneous response from nameserver" when it receives REFUSED responses.

PR #511, opened 2019-03-14
- Fixes zonemaster/zonemaster-engine#504

vlevigneron:
- Inhibits for the moment blacklisting mechanism. We need to fix zonemaster/zonemaster-engine#438 before considering to re-activate it.
- It could slower execution time of very broken zones.
- The code is still there, we just have added a flag to disable it (re-activitaion will be trivial).


matsduf:
@vlevigneron! The problem occurs when the server sends a REFUSED and we take that as no response. There is one Test Case that could trigger REFUSED and that is NAMESERVER01 because we query for a name completely outside the zone in question. It is normal for a authoritative name server to respond with REFUSED in that case.

I all other cases we query for names in or below the zone, and then the name server will not respond with REFUSED unless it is not configured with the zone, and then it will consistently respond with REFUSED and blacklisting should not be a problem.

Is it possible to turn blacklisting off for just NAMESERVER01?

Blacklisting is more complicated than above, there is potentially unwanted blacklisting unrelated to the problem of #438 and that is elaborated in #513.

vlevigneron:
@matsduf I agree with you in the fact that reply with a REFUSED can be a legitimate answer from a server. The problem here is that this is something that is never tested in all test cases we have (we never explicitely check REFUSED rcode returned). And this is exactly what happens, we query for '.', then the server answer with a REFUSED that we interpret as a no response , as a consequence we blacklist this server.

That's why it is more complicated than just this case that I propose to remove, and came back later when we will have fixed LDNS part or find a workaround that allow us to make a difference between nos response and REFUSED.

I can add a flag to allow the code to disable blacklisting on demand instead of a global switch. Is it what you propose for this patch ?

vlevigneron:
Here is my proposal (it will be pushed in minutes) : 
- I kept the global switch (in case we decide to completely remove blacklisting mechanism).
- I added an on demand flag that should be used for queries that are supposed to answer with legitimate REFUSED (like Nameserver07). That will prevent to add the server in the black list. That flag should be used for all queries of names completely outside the zone we are testing.

lib/Zonemaster/Engine/Constants.pm
```diff
@@ -67,6 +68,8 @@ Readonly our $ALGO_STATUS_OTHER        => 5;
 Readonly our $ALGO_STATUS_DELETE_DS    => 6;
 Readonly our $ALGO_STATUS_INDIRECT_KEY => 7;
 
+Readonly our $BLACKLISTING_ENABLED     => 1;
```
mattias-p: I don't think it makes much sense to have this as Readonly. Please consider either making it writable or removing it.

lib/Zonemaster/Engine/Nameserver.pm
```diff
@@ -393,9 +393,11 @@ sub _query {
             $msg =~ s/$trailing_info.*/\./;
             Zonemaster::Engine->logger->add( LOOKUP_ERROR =>
                   { message => $msg, ns => "$self", name => "$name", type => $type, class => $href->{class} } );
-            $self->blacklisted->{ $flags{usevc} }{ $flags{dnssec} } = 1;
-            if ( !$flags{dnssec} ) {
-                $self->blacklisted->{ $flags{usevc} }{ !$flags{dnssec} } = 1;
+            if ( not $href->{q{blacklisting_disabled}} ) {
```
mattias-p: Could you add some documentation for this new option?

vlevigneron: OK. Just after lunch.

lib/Zonemaster/Engine/Constants.pm
vlevigneron: It makes sense for me. It's a global switch that can not be modified. I remind you this is a temporary hack and the whole mechanism should be reviewed and modified accordingly when we will have find out whats wrong in zonemaster/LDNS interaction for "no responses"/REFUSED cases. This flag should be removed at this occasion.

lib/Zonemaster/Engine/Nameserver.pm
vlevigneron: Is it enough for you ?

mattias-p: Yeah, looks good :)

matsduf:
Nameserver01 must also have the "blacklist no" setting, and in both cases (Nameserver01 and Nameserver07) this is not only for now, but permanently.

vlevigneron:
@matsduf just did it for Nameserver01

PR #519, opened 2019-04-08
Dropped support for Ubuntu 14.04

mattias-p:
After dropping support for Ubuntu 14.04, none of the supported platforms use Perl 5.18.


sandoche2k:
@mattias-p merge?

PR #528, opened 2019-04-12
Fix a truncated msgid

mattias-p:
Fix the msgid for the NAMESERVER:NS_ERROR (of Nameserver12) is which is currently truncated.

### DIFF
--- a/lib/Zonemaster/Engine/Test/Nameserver.pm
+++ b/lib/Zonemaster/Engine/Test/Nameserver.pm
@@ -222,7 +222,7 @@ sub translation {
         NO_RESOLUTION              => 'No nameservers succeeded to resolve to an IP address.',
         NO_RESPONSE                => 'No response from {ns}/{address} asking for {dname}.',
         NO_UPWARD_REFERRAL         => 'None of the following nameservers returns an upward referral : {names}.',
-        NS_ERROR                   => 'Nameserver {ns}/{address} ',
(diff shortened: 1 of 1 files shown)


lib/Zonemaster/Engine/Test/Nameserver.pm
```diff
@@ -222,7 +222,7 @@ sub translation {
         NO_RESOLUTION              => 'No nameservers succeeded to resolve to an IP address.',
         NO_RESPONSE                => 'No response from {ns}/{address} asking for {dname}.',
         NO_UPWARD_REFERRAL         => 'None of the following nameservers returns an upward referral : {names}.',
-        NS_ERROR                   => 'Nameserver {ns}/{address} ',
+        NS_ERROR                   => 'Nameserver {ns}/{address} gave an erroneous response to a query with a set EDNS Z flag',
```
matsduf: NS_ERROR is outputted from several test cases. It must a more general wording:
```
The response from nameserver {ns}/{address} was incorrect.
```

mattias-p: That's close to useless from an end user perspective. Updating.

mattias-p: Created zonemaster/zonemaster#755.

matsduf: Also see #529 for another approach.

PR #531, opened 2019-04-16
English translations

mattias-p:
Here are English translations for newly introduced messages, and also a few drive-by tweaks of old ones. Since none of the other translation files refer to the new msgids yet, I took the liberty to update the msgids themselves. Having correct srings in the msgids makes it easier to make the other translations.

### DIFF
--- a/lib/Zonemaster/Engine/Test/DNSSEC.pm
+++ b/lib/Zonemaster/Engine/Test/DNSSEC.pm
@@ -348,13 +348,13 @@ sub metadata {
 sub translation {
     return {
         ADDITIONAL_DNSKEY_SKIPPED => "No DNSKEYs found. Additional tests skipped.",
-        ALGORITHM_DELETE_DS       => "The DNSKEY with tag {keytag} uses Delete DS algorithm number "
+        ALGORITHM_DELETE_DS       => "The DNSKEY with tag {keytag} uses the Delete DS algorithm number "
           . "{algorithm}/({description}).",
         ALGORITHM_DEPRECATED => "The DNSKEY with tag {keytag} uses deprecated algorithm number "
           . "{algorithm}/({description}).",
-        ALGORITHM_INDIRECT_KEY => "The DNSKEY with tag {keytag} uses algorithm number reserved for indirect keys "
+        ALGORITHM_INDIRECT_KEY => "The DNSKEY with tag {keytag} uses an algorithm number reserved for indirect keys "
           . "{algorithm}/({description}).",
-        ALGORITHM_NOT_ZONE_SIGN => "The DNSKEY with tag {keytag} uses algorithm number not meant for zone signing"
+        ALGORITHM_NOT_ZONE_SIGN => "The DNSKEY with tag {keytag} uses an algorithm number not meant for zone signing "
           . "{algorithm}/({description}).",
         ALGORITHM_OK => "The DNSKEY with tag {keytag} uses algorithm number {algorithm}/({description}), which is OK.",
         ALGORITHM_PRIVATE  => "The DNSKEY with tag {keytag} uses private algorithm number {algorithm}/({description}).",
@@ -405,7 +405,7 @@ sub translation {
         NO_KEYS_OR_NO_SIGS_OR_NO_SOA => "Cannot test SOA signatures, because we got {keys} DNSKEY records, "
           . "{sigs} RRSIG records and {soas} SOA records.",
         NO_NSEC3PARAM          => "{server} returned no NSEC3PARAM records.",
-        NO_RESPONSE_DNSKEY     => "Nameserver {ns}/{address} responded with no DNSKEY record(s).",
+        NO_RESPONSE_DNSKEY     => "Nameserver {ns}/{address} responded without DNSKEY record(s).",
         NO_RESPONSE            => "Nameserver {ns}/{address} did not respond.",
         NOT_SIGNED             => "The zone is not signed with DNSSEC.",
         NSEC3_COVERS_NOT       => "NSEC3 record does not cover {name}.",
--- a/lib/Zonemaster/Engine/Test/Delegation.pm
+++ b/lib/Zonemaster/Engine/Test/Delegation.pm
@@ -112,8 +112,8 @@ sub metadata {
 sub translation {
     return {
         ARE_AUTHORITATIVE    => "All these nameservers are confirmed to be authoritative : {nsset}.",
-        CHILD_DISTINCT_NS_IP => "All the IP addresses used by the nameservers in child are unique.",
-        CHILD_NS_SAME_IP     => "IP {address} in child refers to multiple nameservers ({nss}).",
+        CHILD_DISTINCT_NS_IP => "All the IP addresses used by the nameservers in the zone are unique.",
+        CHILD_NS_SAME_IP     => "IP {address} in the zone refers to multiple nameservers ({nss}).",
         DEL_DISTINCT_NS_IP   => "All the IP addresses used by the nameservers in parent are unique.",
         DEL_NS_SAME_IP       => "IP {address} in parent refers to multiple nameservers ({nss}).",
         DISTINCT_IP_ADDRESS  => "All the IP addresses used by the nameservers are unique",
@@ -146,11 +146,11 @@ sub translation {
           . "Lower limit set to {minimum}.",
         NO_IPV4_NS_CHILD => "Child lists no nameserver that resolves to an IPv4 address. If any were present, "
           . "the minimum allowed would be {minimum}.",
-        NO_IPV4_NS_DEL => "Delegation lists no nameserver that resolves to an IPv4 address. If any were present, "
+        NO_IPV4_NS_DEL => "Parent lists no nameserver that resolves to an IPv4 address. If any were present, "
           . "the minimum allowed would be {minimum}.",
         NO_IPV6_NS_CHILD => "Child lists no nameserver that resolves to an IPv6 address. If any were present, "
           . "the minimum allowed would be {minimum}.",
-        NO_IPV6_NS_DEL => "Delegation lists no nameserver that resolves to an IPv6 address. If any were present, "
+        NO_IPV6_NS_DEL => "Parent lists no nameserver that resolves to an IPv6 address. If any were present, "
           . "the minimum allowed would be {minimum}.",
         NS_RR_IS_CNAME          => "Nameserver {ns} {address_type} RR point to CNAME.",
         NS_RR_NO_CNAME          => "No nameserver point to CNAME alias.",
--- a/lib/Zonemaster/Engine/Test/Delegation.pm
+++ b/lib/Zonemaster/Engine/Test/Delegation.pm
@@ -146,11 +146,11 @@ sub translation {
           . "Lower limit set to {minimum}.",
         NO_IPV4_NS_CHILD => "Child lists no nameserver that resolves to an IPv4 address. If any were present, "
           . "the minimum allowed would be {minimum}.",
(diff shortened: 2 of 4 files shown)


lib/Zonemaster/Engine/Test/Delegation.pm
```diff
@@ -112,45 +112,45 @@ sub metadata {
...
     return {
         ARE_AUTHORITATIVE    => "All these nameservers are confirmed to be authoritative : {nsset}.",
-        CHILD_DISTINCT_NS_IP => "All the IP addresses used by the nameservers in child are unique.",
-        CHILD_NS_SAME_IP     => "IP {address} in child refers to multiple nameservers ({nss}).",
-        DEL_DISTINCT_NS_IP   => "All the IP addresses used by the nameservers in parent are unique.",
-        DEL_NS_SAME_IP       => "IP {address} in parent refers to multiple nameservers ({nss}).",
+        CHILD_DISTINCT_NS_IP => "All the IP addresses used by the nameservers in the zone are unique.",
+        CHILD_NS_SAME_IP     => "IP {address} in the zone refers to multiple nameservers ({nss}).",
+        DEL_DISTINCT_NS_IP   => "All the IP addresses used by the nameservers in the delegation are unique.",
+        DEL_NS_SAME_IP       => "IP {address} in the delegation refers to multiple nameservers ({nss}).",
```
matsduf: I think we should consider being more explicit, e.g.:

```
CHILD_DISTINCT_NS_IP => "All the IP addresses used by the nameservers in the zone (the domain name tested) are unique.",
CHILD_NS_SAME_IP     => "IP {address} in the zone (the domain name tested) refers to multiple nameservers ({nss}).",
DEL_DISTINCT_NS_IP   => "All the IP addresses used by the nameservers in the delegation of the domain name tested are unique.",
DEL_NS_SAME_IP       => "IP {address} in the delegation of the domain name tested refers to multiple nameservers ({nss}).",
```

Is it a problem if the message is long? What is the limit? Could it be several sentences?

What do you think, @sandoche2k?

mattias-p: We should make an effort to keep the messages short and concise but having a having a hard limit on message length doesn't make much sense.

Zone and delegation are very central concepts. About one quarter of the messages refer to them. Repeating the phrase "the domain tested" in all of those messages doesn't make much sense to me.

sandoche2k: I agree with @mattias-p. Short is better but with no restrictions. Long messages are never read :-)

lib/Zonemaster/Engine/Test/DNSSEC.pm
```diff
@@ -348,13 +348,13 @@ sub metadata {
 sub translation {
     return {
         ADDITIONAL_DNSKEY_SKIPPED => "No DNSKEYs found. Additional tests skipped.",
-        ALGORITHM_DELETE_DS       => "The DNSKEY with tag {keytag} uses Delete DS algorithm number "
+        ALGORITHM_DELETE_DS       => "The DNSKEY with tag {keytag} uses the Delete DS algorithm number "
```
sandoche2k: the => a

lib/Zonemaster/Engine/Test/Delegation.pm
```diff
@@ -112,45 +112,45 @@ sub metadata {
 sub translation {
     return {
         ARE_AUTHORITATIVE    => "All these nameservers are confirmed to be authoritative : {nsset}.",
-        CHILD_DISTINCT_NS_IP => "All the IP addresses used by the nameservers in child are unique.",
-        CHILD_NS_SAME_IP     => "IP {address} in child refers to multiple nameservers ({nss}).",
-        DEL_DISTINCT_NS_IP   => "All the IP addresses used by the nameservers in parent are unique.",
-        DEL_NS_SAME_IP       => "IP {address} in parent refers to multiple nameservers ({nss}).",
+        CHILD_DISTINCT_NS_IP => "All the IP addresses used by the nameservers in the zone are unique.",
```
sandoche2k: Child zone seems to be appropriate for me

lib/Zonemaster/Engine/Test/Delegation.pm
sandoche2k: Parent zone seems to be appropriate to me

lib/Zonemaster/Engine/Test/DNSSEC.pm
mattias-p: Really? There's only one algorithm number that means Delete DS, so in my opinion "the" is better.

lib/Zonemaster/Engine/Test/Delegation.pm
mattias-p: I like "the zone" and "the delegation" better. "The zone" is the one we're testing and "the delegation" is its delegation.
If we're using the "parent"/"child" terminology, it's clear that we're dealing with two zones, but it's (slightly) less obvious which ones.

mattias-p: I'd like to point out that RFC 7199 was obsoleted by [RFC 8499](https://tools.ietf.org/html/rfc8499), but that doesn't invalidate the point you're making.

I agree that the use of "delegation" in my suggestion doesn't fit with the definitions in the RFCs. I'll revert to "child"/"parent".

mattias-p:
@matsduf, @sandoche2k, @vlevigneron Please check again.

lib/Zonemaster/Engine/Test/Delegation.pm
matsduf: It is correct that [RFC 8499] defines "delegation" as "[t]he process by which a separate zone is created" but the RFC itself uses "delegation" to also mean the DNS records that creates the delegation, e.g. "(...) the same as the origin of the zone that contains the **delegation** to the name server (...)" (my emphasis).

My conclusion is that there is no problem to use "delegation" to refer the the DNS records that create the delegation. It is used in that sense over and over again by DNS experts.

[RFC 8499]: https://tools.ietf.org/html/rfc8499

lib/Zonemaster/Engine/Test/Delegation.pm
```diff
@@ -112,8 +112,8 @@ sub metadata {
 sub translation {
     return {
         ARE_AUTHORITATIVE    => "All these nameservers are confirmed to be authoritative : {nsset}.",
-        CHILD_DISTINCT_NS_IP => "All the IP addresses used by the nameservers in child are unique.",
-        CHILD_NS_SAME_IP     => "IP {address} in child refers to multiple nameservers ({nss}).",
+        CHILD_DISTINCT_NS_IP => "All the IP addresses used by the nameservers in the zone are unique.",
```
matsduf: It is ambigues to refer to "the zone" when we have two zones. I suggest "the zone for the domain being tested". The argument is that it is too long, but complex things cannot always be explained in a short sentence.

lib/Zonemaster/Engine/Test/Delegation.pm
```diff
@@ -146,11 +146,11 @@ sub translation {
           . "Lower limit set to {minimum}.",
         NO_IPV4_NS_CHILD => "Child lists no nameserver that resolves to an IPv4 address. If any were present, "
           . "the minimum allowed would be {minimum}.",
-        NO_IPV4_NS_DEL => "Delegation lists no nameserver that resolves to an IPv4 address. If any were present, "
+        NO_IPV4_NS_DEL => "Parent lists no nameserver that resolves to an IPv4 address. If any were present, "
```
matsduf: "Parent" --> "Parent zone" or "Parent zone (parent domain)"

lib/Zonemaster/Engine/Test/Delegation.pm
```diff
@@ -146,11 +146,11 @@ sub translation {
...
           . "the minimum allowed would be {minimum}.",
         NO_IPV6_NS_CHILD => "Child lists no nameserver that resolves to an IPv6 address. If any were present, "
           . "the minimum allowed would be {minimum}.",
-        NO_IPV6_NS_DEL => "Delegation lists no nameserver that resolves to an IPv6 address. If any were present, "
+        NO_IPV6_NS_DEL => "Parent lists no nameserver that resolves to an IPv6 address. If any were present, "
```
matsduf: "Parent" --> "Parent zone" or "Parent zone (parent domain)"

lib/Zonemaster/Engine/Test/Delegation.pm
mattias-p: I created issue zonemaster/zonemaster#760 to create a glossary for technical terms. I suggest that "the zone" is added as a term to the glossary with the definition "the zone for the domain being tested". To contrast this we could define "the parent zone" to mean "the parent zone for the domain being tested", or possibly "the parent zone of the zone for the domain being tested", if you prefer to be explicit.

lib/Zonemaster/Engine/Test/Delegation.pm
mattias-p: Reverting to old wording.

lib/Zonemaster/Engine/Test/Delegation.pm
mattias-p: Reverting to old wording.

lib/Zonemaster/Engine/Test/DNSSEC.pm
```diff
@@ -405,7 +405,7 @@ sub translation {
         NO_KEYS_OR_NO_SIGS_OR_NO_SOA => "Cannot test SOA signatures, because we got {keys} DNSKEY records, "
           . "{sigs} RRSIG records and {soas} SOA records.",
         NO_NSEC3PARAM          => "{server} returned no NSEC3PARAM records.",
-        NO_RESPONSE_DNSKEY     => "Nameserver {ns}/{address} responded with no DNSKEY record(s).",
+        NO_RESPONSE_DNSKEY     => "Nameserver {ns}/{address} responded without DNSKEY record(s).",
```
matsduf: "without DNSKEY" --> "without any DNSKEY"

lib/Zonemaster/Engine/Test/Delegation.pm
matsduf: "All the IP addresses used by the nameservers in the zone are unique." --> "The IP addresses of the nameservers listed in the zone of the tested domain are unique."

lib/Zonemaster/Engine/Test/Delegation.pm
```diff
@@ -112,8 +112,8 @@ sub metadata {
 sub translation {
     return {
         ARE_AUTHORITATIVE    => "All these nameservers are confirmed to be authoritative : {nsset}.",
-        CHILD_DISTINCT_NS_IP => "All the IP addresses used by the nameservers in child are unique.",
-        CHILD_NS_SAME_IP     => "IP {address} in child refers to multiple nameservers ({nss}).",
+        CHILD_DISTINCT_NS_IP => "All the IP addresses used by the nameservers in the zone are unique.",
+        CHILD_NS_SAME_IP     => "IP {address} in the zone refers to multiple nameservers ({nss}).",
```
matsduf: "IP {address} in the zone refers to multiple nameservers ({nss})." --> "Of the nameservers listed in the zone of the tested domain there are multiple such ({nss}) having the IP address {address}."

mattias-p:
@matsduf and @sandoche2k, please go through your review comment and resolve the conversations where appropriate.

matsduf:
> @matsduf and @sandoche2k, please go through your review comment and resolve the conversations where appropriate.

Done.

mattias-p:
@matsduf I didn't mean to make a re-request for your review. I mis-clicked.

mattias-p:
I'll start over from scratch.

Issue #451, opened 2018-12-04
Avoid installing dependencies from CPAN in installation instruction


mattias-p:
For CentOS we're relying on CPAN for some runtime dependencies. This increases the burden of packaging Engine for CentOS.
The dependencies in question are:
* Locale::TextDomain - [Locale::TextDomain] 1.20 (perl-libintl) is available in CentOS but Engine requires 1.23. We should lower the requirement and add workarounds as needed.
* Mail::RFC822::Address - We should look into the possibility of migrating to [Email::Address] or [MailTools] (available from CentOS) or [Email::Valid] (available from EPEL).
* Text::CSV - Available from EPEL.

The mentioned potential replacements are available for all the other supported OSes.

For reference:
* [CentOS packages database]
* [Extra Packages for Enterprise Linux (EPEL)]

**Edit:** Lots of work has been done on dependencies since this issue was created and the old description was badly out of date, so I updated it to match the current state of things. (2019-04-08)

[CentOS packages database]: http://mirror.centos.org/centos/7/os/x86_64/Packages/
[Email::Address]: https://metacpan.org/pod/Email::Address
[Email::Valid]: https://metacpan.org/pod/Email::Valid
[Extra Packages for Enterprise Linux (EPEL)]: https://ftp.acc.umu.se/mirror/fedora/epel/7/x86_64/Packages/p/
[Locale::TextDomain]: https://metacpan.org/pod/Locale::TextDomain
[MailTools]: https://metacpan.org/pod/MailTools

matsduf:
We get translation issues with lower versions of Locale::TextDomain. 1.23 was the lowest version that we verified that is works. I might be enough with 1.21, but 1.20 is too low.

matsduf:
p5-Net-IP-XS version 0.21 is available on FreeBSD as a package.

mattias-p:
Thanks @matsduf! I didn't realize the FreeBSD Ports Search was case sensitive.

matsduf:
I suggest that this issue is renamed to cover all four OSs and be structed in the same way as the start of the description of zonemaster/zonemaster-backend#449. In that way we can synchronize between the OSs.

matsduf:
Related to packaging of FreeBSD.

PR #547, opened 2019-05-21
Install all runtime dependencies from binary packages

mattias-p:
With this PR we're satisfying all of Engine's runtime dependencies with OS packages, fixing #451.

* Text::CSV is taken from EPEL on CentOS,
* Mail::Address::RFC822 is replaced with Email::Valid, and
* the minimum required version for Locale::TextDomain is lowered to 1.20.

Note that this is for 2019.2.


Makefile.PL
```diff
@@ -16,20 +16,20 @@ all_from 'lib/Zonemaster/Engine.pm';
...
-requires 'Text::CSV'             => 0;
-requires 'Zonemaster::LDNS'      => 2.0;
+requires 'Clone'              => 0;
+requires 'Email::Valid'       => 0;
+requires 'File::ShareDir'     => 1.00;
+requires 'File::Slurp'        => 0;
+requires 'IO::Socket::INET6'  => 2.69;
+requires 'JSON::PP'           => 0;
+requires 'List::MoreUtils'    => 0;
+requires 'Locale::TextDomain' => 1.20;
```
matsduf: We have previously concluded that we need version 1.23 of Locale::TextDomain for translation to work in all environments.

mattias-p:
I'll test Debian, @matsduf has promised to test on FreeBSD, @sandoche2k has promised to test on CentOS and Ubuntu.

matsduf:
Issue for FreeBSD in #556. Waiting for this to be merged.

mattias-p:
I tested this on Debian 8 and Debian 9 and it works on both.

PR #550, opened 2019-05-27
For travis testing. Not to be merged.

matsduf:
Just a meaningless comment added.


matsduf:
Verified that it works after release.

PR #555, opened 2019-06-03
A util that prints recorded data in dig format

mattias-p:
I wrote this some time ago. It's quite useful in a range of situations, e.g. debugging.

~MANIFEST.SKIP should probably be updated also.~ MANIFEST.SKIP does not need to be updated.

### DIFF
--- a/util/data2dig
+++ b/util/data2dig
@@ -1,0 +1,109 @@
+#!/usr/bin/env perl
+use strict;
+use warnings;
+use feature 'say';
+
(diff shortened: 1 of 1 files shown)


mattias-p:
I agree that changes in the data format is a problem. However, that is a problem for all of Zonemaster::Engine and not specific to `data2dig`. This tool doesn't do its own parsing, but instead just acts as a front-end to whatever Zonemaster::Engine library you're using.

Issue #546, opened 2019-05-19
Translation is not done by default on FreeBSD 11


matsduf:
This issue has been found on FreeBSD version 11.2-RELEASE-p10 (but it does not exist on tested version 12.0-RELEASE-p3) when running on develop branch of Zonemaster-Engine (3d0c148) and Zonemaster-CLI ([fa4a3c6]), respectively. They have both the same versions of p5-Locale-Msgfmt (0.15_2), p5-Locale-libintl (1.31) and perl5 (5.28.1_1).

If translation works, and no translation has been requested, then the messages should be translated to English. In most cases that means no change, but in a few cases the "msgid" from the code is change to a different "msgstr" by the translation for English. Two such examples (from [en.po]):
```
#: ADDRESS:NAMESERVERS_IP_WITH_REVERSE
msgid  "Reverse DNS entry exist for all Nameserver IP addresses."
msgstr "Reverse DNS entry exists for each Nameserver IP address."
```
```
#: ADDRESS:NAMESERVER_IP_PTR_MATCH
msgid  "All reverse DNS entry matches name server name."
msgstr "Every reverse DNS entry matches name server name."
```

The following two tests are expected to give the same translated messages, but they do not.
```
# zonemaster-cli --level info zonemaster.net --no-ipv6 --show_module --test Address      
(...)
  10.83 INFO      ADDRESS      Reverse DNS entry exist for all Nameserver IP addresses.
  10.85 INFO      ADDRESS      All reverse DNS entry matches name server name.
```
```
# zonemaster-cli --level info zonemaster.net --no-ipv6 --show_module --test Address --locale en_US.UTF-8
(...)
  10.87 INFO      ADDRESS      Reverse DNS entry exists for each Nameserver IP address.
  10.89 INFO      ADDRESS      Every reverse DNS entry matches name server name.
```
When running then commands on a 12.0-RELEASE-p3 system translation is done i both cases, as expected.

Translation as such works also for other languages on the 11.2-RELEASE-p10 system, e.g.:
```
# zonemaster-cli --level info zonemaster.net --no-ipv6 --show_module --test Address --locale sv_SE.UTF-8
 Sekund Nivå      Modul        Meddelande
======= ====      ============ ==========
   0.00 INFO      SYSTEM       Använder version v3.0.2 av Zonemasters testmotor.
   4.94 INFO      ADDRESS      Alla namnserveradresser finns i den publikt routningsbara addressrymden.
[...]

PR #562, opened 2019-06-13
Partial fix for translation on FreeBSD 11

mattias-p:
The first commit is a partial fix for #546. It makes it possible to select translation for a single locale, but once you do that, translations stuck on that locale and you cannot switch to another. This is good enough for applications like Zonemaster CLI, but not Zonemaster Backend RPCAPI (which I believe reuses the same process for multiple requests).

The remaining commits are just other improvements I came up with while working on #546.

### DIFF
--- a/lib/Zonemaster/Engine/Translator.pm
+++ b/lib/Zonemaster/Engine/Translator.pm
@@ -76,6 +76,10 @@ sub translate_tag {
         return $entry->string;
     }
 
+    # Partial workaround for FreeBSD 11. It works once, but then translation
+    # gets stuck on that locale.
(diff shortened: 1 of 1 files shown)


mattias-p:
I'd say these are low-risk changes. They should have no negative effect on compliant platforms. Just for the sake of it I verified that translations still work as expected on both Debian 8 and 9.

mattias-p:
@matsduf I'm posing the same questions here that I posed in zonemaster/zonemaster-backend#526: What changes are you requesting? Do you think that no fix is better than this partial fix as currently implemented? I've spent a fair amount of time on this already and intend to shift my focus to other things.

PR #565, opened 2019-06-19
Use Net::IP::XS when available

### LINKED ISSUE #293: Make use of Net::IP::XS if available
For performance reasons, please use Net::IP::XS if it is available to the Perl version running.

There has been a serious performance degradation since introducing Net::IP instead of Net::IP::XS.

mattias-p:
Fixes #293. This cuts Travis time by approximately 30%.

### DIFF
--- a/lib/Zonemaster/Engine/Net/IP.pm
+++ b/lib/Zonemaster/Engine/Net/IP.pm
@@ -3,43 +3,79 @@ package Zonemaster::Engine::Net::IP;
 use version; our $VERSION = version->declare("v0.0.6");
 
 use strict;
-no strict 'refs';
 use warnings;
 
-use vars qw(@ISA @EXPORT @EXPORT_OK %EXPORT_TAGS);
-
-require Exporter;
-
-@ISA = qw(Exporter);
-
-my $p_class;
-
-eval {
-# Still does not work that way...
-#    require Net::IP::XS;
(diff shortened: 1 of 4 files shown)


lib/Zonemaster/Engine/Net/IP.pm
```diff
@@ -3,45 +3,81 @@ package Zonemaster::Engine::Net::IP;
...
+    my $self = shift;
+    return $self->{_inner}->reverse_ip();
+}
+
+sub version {
+    my $self = shift;
+    return $self->{_inner}->version();
+}
+
+sub ip_is_ipv6 {
```
vlevigneron: Could you explain why ip_is_ipv6 implementation is different from other methods ? Or am I missing something ?

mattias-p: `ip_is_ipv6` is part of the procedural interface while the others are part of the OO interface.

mattias-p: Do you think the documentation could be improved to make that clearer? If you like I could put the method documentations just above their respective implementations, and the POD headings would follow along. That way the code would be more clearly organized with different kinds of things under different kinds of headings as well.

vlevigneron: No, it's fine for me, I should have read more closely Net::IP and Net::IP::XS documentations.

PR #572, opened 2019-06-28
Remove Example test module

### LINKED ISSUE #554: Remove Example test module
#456 removed the Example entry from the default profile, but the Example module itself still exists. As far as I know it is of limited use and I don't believe it is being maintained. It should be removed.

mattias-p:
Fixes #554.


mattias-p:
Apparently the unit tests had a dependency on the Example test module for checking the POLICY_DISABLED tag. I updated the unit tests so that the POLICY_DISABLED test relies on the Address test module instead.

Please review this new commit as well.

Issue #557, opened 2019-06-05
Make translation to Danish complete (Engine)


matsduf:
@haarbo, I hope that you can help us with translation in Zonemaster-Engine too.

Please update [da-po] (see zonemaster/zonemaster-cli#61).

You will find several items where msgstr is in English. Please translate those to Danish. In some of those case msgid and msgstr are different. Always base the translation on msgstr.

This issue is a superset of #518 and replaces it.

[da-po]: https://github.com/zonemaster/zonemaster-engine/blob/master/share/da.po

matsduf:
Resolved by #575.

PR #575, opened 2019-07-08
Translate into da

haarbo:
Translated missing magic into danish


matsduf:
This PR resolved #557.

PR #580, opened 2019-08-12
Make test module auto-detection less dynamic

mattias-p:
This should fix a problem with Travis failing for zonemaster-backend.

The problem caused by Travis installing a newer version of Z::E being on top of an older version without properly cleaning up older version. A module named Z::E::Test::Example existed in the older version but was removed in the newer one. In addition to this there was an API change causing the old version of Z::E::Test::Example to be incompatible with the new version of Z::E.

The problem is solved by auto-detecting the Z::E::Test modules once and for all at configure-time instead of looking for any modules under the Z::E::Test namespace at run-time.

As a drive-by fix I also added to remove the necessity to install Z::E in @inc in development mode.


lib/Zonemaster/Engine/Test.pm
```diff
@@ -12,16 +12,21 @@ use Zonemaster::Engine::Test::Basic;
...
 use POSIX qw[strftime];
 
 my @all_test_modules;
 
-@all_test_modules =
-  sort { $a cmp $b }
-  map { my $f = $_; $f =~ s|^Zonemaster::Engine::Test::||; $f }
-  grep { $_ ne 'Zonemaster::Engine::Test::Basic' } useall( 'Zonemaster::Engine::Test' );
+BEGIN {
+    @all_test_modules = split /\n/, read_file( dist_file( 'Zonemaster-Engine', 'modules.txt' ) );
```
matsduf: Is this really the best solution? Isn't better to create a perl module that lists the modules? I feel that the code would be simpler and more direct.

mattias-p: I just want to fix a bug here. As such I'm trying to change the design as little as possible.

Feel free to create an issue to make the module listing completely static.

sandoche2k:
@vlevigneron to review

PR #608, opened 2019-10-03
- Fixes zonemaster/zonemaster-engine#102

### LINKED ISSUE #102: args.ns should only be a name server name
args.ns should always be only the nameserver name, now it is often a mix of ns/address.
If there is an address, split it out to args.address.

Any change must also update the logging string so that the address is not forgotten,

"Nameserver {ns} did not..." -> "Nameserver {ns}/{address} did not..."

This also affects all the translation strings.

vlevigneron:
Very few cases left before this fix. Now {ns} refers only to a name server (without address) and {nss} refers to a list of name servers.

Resolves #102.

### DIFF
--- a/lib/Zonemaster/Engine/Test/Basic.pm
+++ b/lib/Zonemaster/Engine/Test/Basic.pm
@@ -1,6 +1,6 @@
 package Zonemaster::Engine::Test::Basic;
 
-use version; our $VERSION = version->declare("v1.0.11");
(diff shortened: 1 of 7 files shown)


matsduf:
Is the interpretation that "ns" refer to a single name server (by name), and "nss" to a list of one or more name servers (by name).

I think it would make sense if we have a dictionary so that we use these variables consistently.

PR #623, opened 2019-10-14
Update and correct sv.po

matsduf:
* Repaired the header.
* Added missing trailing space in some cases.
* Translated untranslated.

### DIFF
--- a/share/sv.po
+++ b/share/sv.po
@@ -692,7 +692,7 @@ msgid ""
 "zone({zone_addresses})"
 msgstr ""
 "Delegeringen listar en IP-adress ({parent_addresses}) för en namnserver, "
-"som är 'in-bailiwick', som misstämmer med eventuell motsvarande adresspost "
+"som är 'in-bailiwick', som inte stämmer med eventuell motsvarande adresspost "
 "i dotterzonen ({zone_addresses})."
 
 #. EDNS_VERSION_ERROR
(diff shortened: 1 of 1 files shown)


share/sv.po
```diff
@@ -693,6 +691,9 @@ msgid ""
 "at parent ({parent_addresses}) and any equivalent address record in child "
 "zone({zone_addresses})"
 msgstr ""
+"Delegeringen listar en IP-adress ({parent_addresses}) för en namnserver "
+"som är 'in-bailiwick' som misstämmer eventuell motsvarande adressenpost "
```
mattias-p: "misstämmer" doesn't seem right.

matsduf: Corrected this, and the message with "out-of-bailiwick", that is very similar.

mattias-p: You didn't replace "misstämmer". Is that even a real word? My dictionary says not. The closest thing being "misstämning" which means "olustig stämning". Since connotations and signaling seems to be high on the agenda for the test messages, I think we should look for a better word.

matsduf: Well, Google translate agrees with me, [misstämma]. Yes, it is a real word in real use that works very well in the context. It has nothing to do will "misstämning". It means "stämmer inte". I will update to make more comfortable for you.

[misstämma]: https://translate.google.com/#view=home&op=translate&sl=sv&tl=en&text=misst%C3%A4mmer

matsduf:
@mattias-p, please review again.

share/sv.po
mattias-p: Are you using Google Translate as an authority on language?

matsduf: @mattias-p, Swedish is my native language. I have met the word "misstämma" in the meaning "inte stämma" many times. It is a word in use even if you have never met it.

PR #624, opened 2019-10-14
Updated da.po

matsduf:
* All fuzzy translated replaced by English
* Preparation for translation to Danish

When this is merged, an issue for translation to Danish will be created.

### DIFF
--- a/share/da.po
+++ b/share/da.po
@@ -1,11 +1,9 @@
-#. WRONG_SOA
-#: ../lib/Zonemaster/Engine/Test/Zone.pm:266
 msgid ""
 msgstr ""
 "Project-Id-Version: 1.0.0\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2019-10-04 19:08+0200\n"
-"PO-Revision-Date: 2019-07-08\n"
+"POT-Creation-Date: 2019-10-11 17:58+0200\n"
+"PO-Revision-Date: 2019-10-14 16:39+0200\n"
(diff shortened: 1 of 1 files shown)


share/da.po
```diff
@@ -130,15 +128,15 @@ msgstr "Alle navneservernes IP-adresser er unikke."
 
 #. CHILD_DISTINCT_NS_IP
 #: ../lib/Zonemaster/Engine/Test/Delegation.pm:127
-#, fuzzy
+#,
```
mattias-p: May I ask what tool(s) you used to make these updates? Just curious.

share/da.po
```diff
@@ -600,12 +578,12 @@ msgid "IP {address} refers to multiple nameservers ({nss})."
 msgstr "IP-adressen {address} refererer til flere navneservere ({nss})."
 
 #. IPV4_DISABLED
+#: ../lib/Zonemaster/Engine/Test/Nameserver.pm:288
```
mattias-p: I'm bothered by how these line numbers jump around. They're clearly not meant to be checked in. I'll create an issue to replace them with some other module/test case annotations.

share/da.po
matsduf: Just emacs.

share/da.po
mattias-p: Created #625.

share/da.po
mattias-p: PO mode or vanilla?

matsduf: PO mode. I felt that the risk errors was lower.

PR #631, opened 2019-10-18
Replace module references

### LINKED ISSUE #625: Replace source references in checked-in PO files
We should not keep source code references (i.e. `#:` comments) in checked-in PO files. They drive diffs all over the PO files making it harder to work with them. See https://github.com/zonemaster/zonemaster-engine/pull/624/files#diff-0c771b7df6621e822f3bf53df2f6fc37R583.

Today the source references serve the purpose of annotating what modules each message tag belongs to. The source references must not be removed without some other means of annotating the relation between message tags and modules.

When designing the alternative annotation, consider the possibility of referring to test case rather than test modules. Keep in mind that the system messages are related to a module, but they are not related to any test case.

mattias-p:
Fixes #625 the way we discussed on the meeting.

### DIFF
--- a/lib/Zonemaster/Engine/Test/Connectivity.pm
+++ b/lib/Zonemaster/Engine/Test/Connectivity.pm
@@ -88,83 +88,83 @@ sub metadata {
 
 Readonly my %TAG_DESCRIPTIONS => (
     NAMESERVERS_IPV4_WITH_UNIQ_AS => sub {
-        __x    # NAMESERVERS_IPV4_WITH_UNIQ_AS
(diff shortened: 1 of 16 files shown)


lib/Zonemaster/Engine/Test/Connectivity.pm
```diff
@@ -178,7 +178,7 @@ sub version {
 }
 
 ###
-### Tests
+### CONNECTIVITY:Tests
```
mattias-p: I'll fix this.

PR #637, opened 2019-10-18
Clean-up of en.po

matsduf:
* Removed "fuzzy"
* Made msgstr equal msgid unless they are set to be different
* Removed left-over msgid/msgstr

### DIFF
--- a/share/en.po
+++ b/share/en.po
@@ -2,8 +2,8 @@ msgid ""
 msgstr ""
 "Project-Id-Version: 1.0.0\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2019-10-17 18:33+0200\n"
-"PO-Revision-Date: 2017-12-18\n"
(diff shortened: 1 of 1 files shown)


share/en.po
```diff
@@ -503,12 +503,12 @@ msgstr "IP {address} in parent refers to multiple nameservers ({nss})."
...
 msgstr "IP {address} refers to multiple nameservers ({nss})."
 
-#. CONSISTENCY:IPV4_DISABLED
+#. NAMESERVER:IPV4_DISABLED
 #. CONNECTIVITY:IPV4_DISABLED
 #. BASIC:IPV4_DISABLED
 #. DELEGATION:IPV4_DISABLED
-#. NAMESERVER:IPV4_DISABLED
 #. DNSSEC:IPV4_DISABLED
+#. CONSISTENCY:IPV4_DISABLED
```
mattias-p: At least we got rid of the line number annoyance, but we should have some solution to this jumping around as well.

PR #639, opened 2019-10-18
Remove non-determinism in file generation

### LINKED ISSUE #638: Remove non-determinism from file generation
Today the modules.txt and the extracted comments in PO files aren't deterministically genereated. This results in bigger diffs than necessary. That should be fixed.

mattias-p:
Fixes #638.


mattias-p:
I removed the PO file changes and resolve the conflict by rebase.

Issue #646, opened 2019-10-21
Messages to changes in DNSSEC10 test case


vlevigneron:
DS_ALGORITHM_MISSING
DS_ALGORITHM_OK
It has been badly mixed in https://github.com/zonemaster/zonemaster-engine/pull/627

MIXED_NSEC_NSEC3
TEST_ABORTED
NO_NSEC_NSEC3
"{ns/address}" instead of "{ns}/{address}"

PR #644, opened 2019-10-21
Fixes zonemaster-engine French translation

vlevigneron:
- Fixes zonemaster/zonemaster-engine#599
- Fixes zonemaster/zonemaster-engine#551
- Fixes zonemaster/zonemaster-engine#607
Applies only for fr.po.
Some DNSSEC messages missing, will be added soon

PR #650, opened 2019-10-22
- Fixes some DNSSEC10 and DNSSEC14 messages (+ french translation)

vlevigneron:
- Fixes zonemaster/zonemaster-engine#646
- Add missing french translation in zonemaster/zonemaster-engine#644

### DIFF
--- a/lib/Zonemaster/Engine/Test/DNSSEC.pm
+++ b/lib/Zonemaster/Engine/Test/DNSSEC.pm
@@ -1,6 +1,6 @@
 package Zonemaster::Engine::Test::DNSSEC;
 
-use version; our $VERSION = version->declare("v1.1.8");
+use version; our $VERSION = version->declare("v1.1.9");
 
 ###
 ### This test module implements DNSSEC tests.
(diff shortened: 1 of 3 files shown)


matsduf:
@vlevigneron, it is not only DNSSEC10 and DNSSEC14 as the headline says. At least DNSSEC01 is also involved.

lib/Zonemaster/Engine/Test/DNSSEC.pm
```diff
@@ -529,11 +529,11 @@ Readonly my %TAG_DESCRIPTIONS => (
     },
     DS_ALGORITHM_MISSING => sub {
         __x    # DNSSEC:DS_ALGORITHM_MISSING
-          "{ns}/{address} returned a DS record created by algorithm {algorithm_number}/{algorithm_mnemonic}, which is OK. The DS record is for the DNSKEY record with keytag {keytag} in zone {zone}.", @_;
+          "{ns}/{address} returned no DS record created by algorithm {algorithm_number}/{algorithm_mnemonic} for zone {zone}, which is required.", @_;
```
matsduf: DNSSEC01. "algorithm_number" here is always 2, so it could be a constant instead to clarify that.

PR #657, opened 2019-10-25
Empty

mattias-p:
Checking again.


mattias-p:
Travis works as expected after merging fix.

Issue #658, opened 2019-10-28
Consistency05 reports error on irrelevant difference


matsduf:
We get the following output from a test of a domain with Consistency05:
```
$ zonemaster-cli --test consistency/consistency05 --no-ipv6 --level INFO Verkkotunnus1.fi
Seconds Level     Message
======= ========= =======
   0.00 INFO      Using version v3.0.3 of the Zonemaster engine.
   4.23 ERROR     CONSISTENCY:OUT_OF_BAILIWICK_ADDR_MISMATCH parent_addresses=ns2.z.fi./162.219.55.2;ns2.z.fi./2620:10a:80ec::2; zone_addresses=ns2.z.FI./162.219.55.2;ns2.z.FI./2620:10a:80ec::2
   5.21 ERROR     CONSISTENCY:OUT_OF_BAILIWICK_ADDR_MISMATCH parent_addresses=ns1.z.fi./162.219.54.2;ns1.z.fi./2620:10a:80eb::2; zone_addresses=ns1.z.FI./162.219.54.2;ns1.z.FI./2620:10a:80eb::2
```

1. The message is not about differences in owner name, only in differences in IP address with the same owner name.
2. Difference in case, `fi` vs `FI` is irrelevant.

A longer log file is available.

vlevigneron:
@matsduf  I'm on it...

matsduf:
@vlevigneron, it was a strange behavior of one of the servers that responede with `z.FI` even though the question was `z.fi`. And the behavior was only if there was no EDNS in the query.

matsduf:
PR #659 will resolve this issue.

vlevigneron:
@matsduf A new test case to take care of ???

matsduf:
> @matsduf A new test case to take care of ???

There is [Nameserver08] and [Nameserver09], but they check the behavior of the name servers for the zone itself, not the name servers of the parent zone as in this case.

[Nameserver08] and [Nameserver09] need update, however.

[Nameserver08]: https://github.com/zonemaster/zonemaster/blob/develop/docs/specifications/tests/Nameserver-TP/nameserver08.md
[Nameserver09]: https://github.com/zonemaster/zonemaster/blob/develop/docs/specifications/tests/Nameserver-TP/nameserver09.md

matsduf:
Resolved by #659.

PR #659, opened 2019-10-28
Fix case sensitivity in Consistency05

vlevigneron:
Fixes zonemaster/zonemaster-engine#658


vlevigneron:
@matsduf I agree, we should check and identify this test cases and try to configure some specific zone for the unit tests (If we can reproduce the zone issue).

Issue #670, opened 2019-11-21
Test-dnssec.t fails


mattias-p:
When running the unit tests in the current develop branch (6986f7a) this failure is encountered:

```
===(    1428;155  162/?  23/? )=========================================
#   Failed test 'afnic.fr gives ALL_ALGO_SIGNED'
#   at t/Test-dnssec.t line 19.
#   Failed test 'afnic.fr does not give RRSIG_BROKEN'
#   at t/Test-dnssec.t line 30.
===(    1452;162  178/?  31/? )=========================================# Looks like you failed 2 tests of 189.
t/Test-dnssec.t ........... Dubious, test returned 2 (wstat 512, 0x200)
Failed 2/189 subtests 
	(less 2 skipped subtests: 185 okay)
```

I've seen this on Travis and also locally on Debian 10 when installing the preliminary dist tarball using `cpanm`.

mattias-p:
The mentioned message tags are emitted by DNSSEC13.

mattias-p:
I tried to re-record the data and that makes these particular tests pass, but it also breaks a couple of others tests in Test-dnssec.t:
```
...
not ok 150 - dnssec08-dnskey-not-signed.zut-root.rd.nic.fr gives DNSKEY_NOT_SIGNED
#   Failed test 'dnssec08-dnskey-not-signed.zut-root.rd.nic.fr gives DNSKEY_NOT_SIGNED'
#   at t/Test-dnssec.t line 19.
not ok 151 - dnssec08-dnskey-not-signed.zut-root.rd.nic.fr gives DNSKEY_SIGNATURE_NOT_OK
#   Failed test 'dnssec08-dnskey-not-signed.zut-root.rd.nic.fr gives DNSKEY_SIGNATURE_NOT_OK'
#   at t/Test-dnssec.t line 19.
...
not ok 164 - afnic.fr does not give NO_RESPONSE
#   Failed test 'afnic.fr does not give NO_RESPONSE'
#   at t/Test-dnssec.t line 30.
...
```

mattias-p:
@vlevigneron Could you take a look at this?

vlevigneron:
Yes I know, I'm working on it. I'm on my way to "fix" broken zones.

I'll propose a PR today for that.

le 22 nov., Mattias Päivärinta a ?crit :
> I tried to re-record the data and that makes these particular tests pass, but it also breaks a couple of others tests in Test-dnssec.t:
> ```
> ...
[...]

PR #671, opened 2019-11-22
Pass DNSSEC tests

vlevigneron:
Fixes issue zonemaster/zonemaster-engine#670
New Test-dnssec.data with refreshed zones


vlevigneron:
@mattias-p It seems there is still a bug. I do not meet that on my own station.

t/Test-dnssec.t
```diff
@@ -195,7 +195,7 @@ zone_gives_not( 'dnssec07', $zone, [qw{NEITHER_DNSKEY_NOR_DS DNSKEY_BUT_NOT_DS D
 
 # dnssec08
 
-$zone = Zonemaster::Engine->zone( 'dnssec08-dnskey-not-signed.zut-root.rd.nic.fr' );
+$zone = Zonemaster::Engine->zone( 'dnssec08-dnskey-signature-not-ok-broken.zut-root.rd.nic.fr' );
```
mattias-p: This looks curious to me. Before the change the zone name and the checked tags seemed to match up, but after the change they don't seem to match up anymore. Are you sure this is right?

mattias-p:
By the way, I like the cleanup you did in the first commit, but don't see how it could change the behavior.

t/Test-dnssec.t
vlevigneron: I did some modification on the zone because it is used for different tests cases.
The dnssec08-dnskey-signature-not-ok-broken zone has 2 signatures for DNSKEY RRset, one is expiree, one is broken. In the other hand dnssec08-dnskey-not-signed.zut-root.rd.nic.fr is a signed zone but with no RRSIG records for DNSKEY RRset. I removed broken RRSIG records from dnssec08-dnskey-not-signed.zut-root.rd.nic.fr to have broken easier to manage.

vlevigneron:
@mattias-p For the clean-up, it's a consequence of what I told during our F2F meeting. I have fixed my perltidy and my perlcritic. That's why just after the release I'll send a PR with lot's of clean-up.

matsduf:
The issue is tagged with milestone v2019.2 so I tagged this with the same milestone.

vlevigneron:
No problem @matsduf, I'll wait for @mattias-p re-review.

matsduf:
@vlevigneron, I think this could be merged.

vlevigneron:
@matsduf Done !

Issue #672, opened 2019-11-25
Fatal error: Can't call method "has_rrs_of_type_for_name"


matsduf:
I have installed Engine on commit f862416 on FreeBSD 11.3 and 12.1. On top of that I have installed CLI. I get 

```
Fatal error in Zonemaster::Engine::Test::Zone: Can't call method "has_rrs_of_type_for_name" on an undefined value at /usr/local/lib/perl5/site_perl/Zonemaster/Engine/Test/Zone.pm line 600.
```
in both installations.

```
# zonemaster-cli --no-ipv6  errors.com
Seconds Level     Message
======= ========= =======
  52.51 NOTICE    Nameserver ns1.digimedia.com has an IP address (23.21.242.88) with mismatched PTR result (ec2-23-21-242-88.compute-1.amazonaws.com.).
  52.58 NOTICE    Nameserver ns2.digimedia.com has an IP address (23.21.243.119) with mismatched PTR result (ec2-23-21-243-119.compute-1.amazonaws.com.).
 123.91 NOTICE    Child lists no nameserver that resolves to an IPv6 address. If any were present, the minimum allowed would be 2.
 124.04 NOTICE    Delegation lists no nameserver that resolves to an IPv6 address. If any were present, the minimum allowed would be 2.
 124.65 ERROR     Parent has nameserver(s) not listed at the child (ns3.digimedia.com).
 124.96 NOTICE    There are neither DS nor DNSKEY records for the zone.
 124.99 NOTICE    The zone is not signed with DNSSEC.
 127.93 NOTICE    Nameserver ns1.digimedia.com/23.21.242.88 allow zone transfer using AXFR.
 128.31 NOTICE    Nameserver ns2.digimedia.com/23.21.243.119 allow zone transfer using AXFR.
 128.77 NOTICE    Nameserver ns3.digimedia.com/54.241.0.203 allow zone transfer using AXFR.
 137.46 WARNING   Domain name MX (0.0.0.0) within a 'numeric only' TLD (0).
 138.15 NOTICE    SOA 'refresh' value (10800) is less than the recommended minimum (14400).
 138.77 CRITICAL  Fatal error in Zonemaster::Engine::Test::Zone: Can't call method "has_rrs_of_type_for_name" on an undefined value at /usr/local/lib/perl5/site_perl/Zonemaster/Engine/Test/Zone.pm line 600.
```

vlevigneron:
@matsduf I'm on it.

[...]

PR #673, opened 2019-11-25
Fix crash when testing "errors.com"

vlevigneron:
Fixes zonemaster/zonemaster-engine#672


vlevigneron:
Travis will be OK when zonemaster/zonemaster-engine#671 will be merged.

Issue #674, opened 2019-11-25
Syntax.pm emits errors on IPv6 when "no-ipv6" has been selected


matsduf:
I am testing from a computer with IPv4 only and have selected `--no-ipv6` but Syntax.pm still emits messages on not being able to reach servers on IPv6.

```
# zonemaster-cli --no-ipv6 --show_module  ed25519.nl
Seconds Level     Module       Message
======= ========= ============ =======
  84.82 WARNING   SYNTAX       No response from mango.plexis.eu/2a01:7c8:aaae:3e2::1 asking for ed25519.nl.
  85.24 WARNING   SYNTAX       No response from pdns-public-ns1.powerdns.com/2a03:b0c0:2:d0::4af:b001 asking for ed25519.nl.
  85.27 WARNING   SYNTAX       No response from pdns-public-ns2.powerdns.com/2604:a880:1:20::132:5001 asking for ed25519.nl.
  87.32 NOTICE    ZONE         SOA 'refresh' value (10800) is less than the recommended minimum (14400).
```

vlevigneron:
@matsduf I'm on it

vlevigneron:
Resolved by #676

PR #676, opened 2019-11-25
Fix Syntax IPv4/IPv6 

vlevigneron:
- Fixes zonemaster/zonemaster-engine#674


vlevigneron:
Travis will be OK when zonemaster/zonemaster-engine#671 will be merged.

Issue #680, opened 2019-12-24
Engine fails to install in CentOS 8


sandoche2k:
```
t/Test-dnssec-more.t ...... ok

#   Failed test 'afnic.fr gives ALL_ALGO_SIGNED'
#   at t/Test-dnssec.t line 19.

#   Failed test 'afnic.fr does not give RRSIG_BROKEN'
#   at t/Test-dnssec.t line 30.
# Looks like you failed 2 tests of 189.
t/Test-dnssec.t ...........
Dubious, test returned 2 (wstat 512, 0x200)
Failed 2/189 subtests
        (less 2 skipped subtests: 185 okay)
t/Test-dnssec05-A.t ....... ok
```

sandoche2k:
Vincent  will you merge this issue since Mats will be back next week only. I will test and get back to you.

vlevigneron:
Review it, I will merge it.

le 26 déc., sandoche2k a ?crit :
> Vincent  will you merge this issue since Mats will be back next week only. I will test and get back to you.
> 
> -- 
> You are receiving this because you were assigned.
> Reply to this email directly or view it on GitHub:
> https://github.com/zonemaster/zonemaster-engine/issues/680#issuecomment-569021106

-- 
	Vincent Levigneron  A.F.N.I.C.  Vincent.Levigneron@afnic.fr

matsduf:
Resolved by #681

PR #681, opened 2019-12-26
- The skipped test is not skipped because it fails but because we sho…

vlevigneron:
Fixes zonemaster/zonemaster-engine#680


matsduf:
Issue #722 to revert, fix or remove has been created.

PR #683, opened 2020-01-02
Updated CentOS Install Instruction

sandoche2k:


docs/Installation.md
```diff
@@ -24,31 +24,53 @@ This instruction covers the following operating systems:
...
+
+   ```sh
+   sudo yum install cpanminus libidn-devel openssl-devel perl-Clone perl-core perl-Devel-CheckLib perl-File-ShareDir perl-File-Slurp perl-libintl perl-IO-Socket-INET6 perl-JSON-PP perl-List-MoreUtils perl-Module-Find perl-Moose perl-Net-IP perl-Pod-Coverage perl-Test-Differences perl-Test-Exception perl-Test-Fatal perl-Test-Pod perl-Text-CSV perl-YAML
+   ```
 
+5) *Only* for CentOS8, Install:
+   
    ```sh
-   sudo yum install cpanminus libidn-devel openssl-devel perl-Clone perl-core perl-Devel-CheckLib perl-Email-Valid perl-File-ShareDir perl-File-Slurp perl-libintl perl-IO-Socket-INET6 perl-JSON-PP perl-List-MoreUtils perl-Module-Find perl-Moose perl-Net-IP perl-Pod-Coverage perl-Readonly-XS perl-Test-Differences perl-Test-Exception perl-Test-Fatal perl-Test-Pod perl-Text-CSV perl-YAML
+   sudo yum install perl-MailTools
```
mattias-p: We're not using MailTools so we don't need to install it.

And please add Email::Valid to the list of CPAN packages. It's included in Makefile.PL so it should be included in the installation instruction as well.

sandoche2k: perl-Email-Valid is removed from CentOS8

mattias-p: Yes, that's why we need to install it from CPAN instead.

We're not using MailTools (correct me if I'm wrong) and therefore we should not install it.

mattias-p:
Oh, and you may want to change the target branch of this PR.

docs/Installation.md
sandoche2k: Doesn't work in CentOS8. There are lot of dependencies that need to be installed. Hence got it rectified by installing l perl-MailTools

matsduf: @sandoche2k, do you say that is does not work to install Email::Valid with cpanm?

matsduf: @sandoche2k, do you say that is does not work to install Email::Valid with cpanm? And with your update, perl-Email-Valid is no longer installed (as binary package) for CentOS 7. Is that really correct?

mattias-p: > [...] perl-Email-Valid is no longer installed (as binary package) for CentOS 7. Is that really correct?

I'd say it is correct. Knowing that a module lacks binary packaging for some OS versions is more important than knowing exactly which OS versions those are. (You don't need to know very often and when you do it's quick enough to figure it out from the `cpanm` lines and the package databases for the respective OS versions.) Also, simplicity of instruction is important.

matsduf: > I'd say it is correct. Knowing that a module lacks binary packaging for some OS versions is more important than knowing exactly which OS versions those are. (You don't need to know very often and when you do it's quick enough to figure it out from the `cpanm` lines and the package databases for the respective OS versions.) Also, simplicity of instruction is important.

With that argument we could ignore trying to find binary packages for the installation instructions and just install everything from CPAN.

I disagree. Knowing that some version of the OS can be installed with binary packages only is important information for understanding the possibility of packaging the Zonemaster components.

mattias-p: > With that argument we could ignore trying to find binary packages for the installation instructions and just install everything from CPAN.

You're not following.

> Knowing that some version of the OS can be installed with binary packages only is important information for understanding the possibility of packaging the Zonemaster components.

I agree with this statement.

While the list of dependencies lacking binary packaging for some OS version is short, the full list of dependencies is quite long. Going through all of them to figure out which ones lack binary packaging is quite a chore. Going though only the shortlist is quick. If you don't think so, let me do it. If you like I can show you how I do it and you can do it yourself.

Maintaining a common shortlist of packages that lack binary packaging for a group of OS versions is easier than maintaining exact sets of packages for each OS version. Also, following an installation instruction with fewer branches is easier for the user. Moreover, both of these use cases are much more common than the use case of finding out the exact set of modules that are not packaged for a specific OS version.

sandoche2k:
@matsduf I think we can remove perl-Email-Valid from CentOS7 also. Got to verify with another new CentOS7 install

docs/Installation.md
sandoche2k: With CPANM, there is no need for installing packages from CPAN
```
[centos@ip-172-31-29-204 ~]$ sudo cpanm Zonemaster-Engine-v3.0.3.tar.gz 
--> Working on Zonemaster-Engine-v3.0.3.tar.gz
Fetching file:///home/centos/Zonemaster-Engine-v3.0.3.tar.gz ... OK
Configuring Zonemaster-Engine-3.000003 ... OK
==> Found dependencies: Email::Valid
--> Working on Email::Valid
Fetching http://www.cpan.org/authors/id/R/RJ/RJBS/Email-Valid-1.202.tar.gz ... OK
Configuring Email-Valid-1.202 ... OK
Building and testing Email-Valid-1.202 ... OK
Successfully installed Email-Valid-1.202
```

matsduf: @sandoche2k, when you install things with cpanm then you install packages from CPAN.

mattias-p: > With CPANM, there is no need for installing packages from CPAN

This is technically correct.

However there are a couple of other use cases we need to consider. Namely 1) finding the set of dependencies without binary packaging for a given OS version, and 2) updating a set of binary packages in the installation instruction. Figuring out the details for these to use cases from first principles is rather cumbersome, so we need to maintain some supporting documentation.

As you point out, having the non-binary-packaged dependencies serves no purpose at all for the installation use case. However we're also using the installation instruction as supporting documentation to support the two use cases mentioned above. Whether this is a good idea or not is debatable, but it's how we've been doing things up until now. We could remove the cpanm commands for dependencies in the installation instruction, but then we have to replace them with some other documentation to support these two other use cases, and we have to somehow integrate the maintenance of that documentation in our processes.

N.B. The list of CPAN-installed dependencies could possibly support some other use case(s) as well, but these two are the ones that have turned up so far in this discussion.

sandoche2k: MailTools is needed for Mail::Address during Engine Installation
```
--> Working on Mail::Address
Fetching http://www.cpan.org/authors/id/M/MA/MARKOV/MailTools-2.21.tar.gz ... OK
Configuring MailTools-2.21 ... OK
Building and testing MailTools-2.21 ... OK
Successfully installed MailTools-2.21
Building and testing Email-Valid-1.202 ... OK
Successfully installed Email-Valid-1.202
--> Working on Readonly
Fetching http://www.cpan.org/authors/id/S/SA/SANKO/Readonly-2.05.tar.gz ... OK
```

sandoche2k:
Updated the installation instruction. Removed all installation from CPAN packages.

docs/Installation.md
mattias-p: @sandoche2k, the Mail::Address module is part of the MailTools package. See https://metacpan.org/pod/Mail::Address. I wonder who's using that. I'm pretty sure it's not us. Maybe it's a transitive dependency? In that case I think we should let cpanm install it without any mention in the installation instruction.

mattias-p:
> Updated the installation instruction. Removed all installation from CPAN packages.

Please put the CPAN packages back. As discussed above we really need them for maintenance purposes. If you really don't want them in the installation instruction you'll have to move them to a separate document, but you can't simply remove them. Keep in mind that the maintenance burden will be higher if they're in a separate document, so I would really advice to keep them in the installation instruction.

matsduf:
> @matsduf I think we can remove perl-Email-Valid from CentOS7 also. Got to verify with another new CentOS7 install

Since Email::Valid is used I think we should keep it as binary package for CentOS 7.

sandoche2k:
@mattias-p We did have this discussion earlier regarding CPAN and CPANM. At that time, it was mentioned that once we use CPANM we will remove the packages from installation instructions. 
I don't see the logic of keeping the CPANM packages for maintenance purpose. While installing Zonemaster components from CPANM, there are many dependency packages being installed. Do you want to find all of them and install prior to installing Zonemaster components? 
I think we are wasting time here. My objective is to simplify installation instructions.

mattias-p:
> We did have this discussion earlier regarding CPAN and CPANM. At that time, it was mentioned that once we use CPANM we will remove the packages from installation instructions.

This is not how I recall things. Both `cpan` and `cpanm` will install dependencies. If we wanted to remove dependencies from the installation instruction we should have been able to do that right away.

> While installing Zonemaster components from CPANM, there are many dependency packages being installed. Do you want to find all of them and install prior to installing Zonemaster components?

No, in principle I do not think it's important that all of them are installed before we run the `cpanm` commands to install the Zonemaster components.

What I do think is important is that we're using binary packages when they are available. Since `cpanm` is unable to resolve these binary package dependencies we do need to list those in the instruction. You seem to agree with this, but let me state perspective. I see two major benefits. Installation is faster and quality control is better.

Since we have the list of binary dependencies in the instruction we need to maintain it, and this is where the `cpanm` command for packages comes in. The purpose of the command is not to have all dependencies installed before the `cpanm Zonemaster::*` command. Its purpose is to serve as supporting documentation for the maintenance of the binary packages list. It also gives us a hint on where we should consider migrating to a different dependency that does have binary packaging. Doing this eases the burden of having Zonemaster binary packaged itself.

It's possible but quite tedious to figure out which dependencies aren't met by the binary packages list - we have so many dependencies and the mapping between Perl module name and binary package name is non-trivial. Having the `cpanm` command for dependencies in the instruction is *really* helpful for this purpose.

I'm not trying to be difficult. I'm just trying to avoid a bunch of tedious, boring and error prone work.

sandoche2k:
@mattias-p 
>  I see two major benefits. Installation is faster and quality control is better.
I do not see the logic. Either one installs all dependencies using CPANM for ZM components and then install ZM components or we let CPANM to do that. This is what is the purpose of tools like CPANM as I understand: http://www.cpan.org/modules/INSTALL.html. 

> Since we have the list of binary dependencies in the instruction we need to maintain it
Binary dependencies are not removed.

matsduf:
@sandoche2k, we have followed the current model for some time now. Direct dependencies are explicitly installed either with a binary package (when available) or from CPAN (with cpanm). We have collected as many binary packages as possible, and the ones installed from CPAN have also been reminders of what to look for as binary packages.

We are now late in the release cycle, and I suggest that we do not change anything now. Please let us discuss your suggestion when the release is ready, and we can adjust for the next release.

perl-Email-Valid should then be restored as binary package for CentOS 7.

mattias-p:
> I do not see the logic.

Please, read my comment again. In this paragraph I'm quite clearly speaking of binary packages. And I believe we're in perfect agreement on this point.

>> Since we have the list of binary dependencies in the instruction we need to maintain it
>
> Binary dependencies are not removed.

The list of binary dependencies is not something that is written once and never touched again. It needs to be kept up to date over time. Knowing what packages to remove is trivial - you know you have to do it when your package manager says a package isn't available, but that seldom happens. Knowing what packages to add is what you have to work for. I've been doing that work pretty much since I joined the Zonemaster project. From experience I can tell you that depending on how you do it, it can be a lot of work. Having the list of dependencies installed from CPAN as supporting documentation makes it so much easier.

sandoche2k:
Then what is the use of tools like apt-get, yum, cpan etc.. if you want to
install explicitly all packages..
My objection is for an user to do less..
Tomorrow if we package Zonemaster are you going to install all dependencies
or let the package manager to do its job.

matsduf:
> Then what is the use of tools like apt-get, yum, cpan etc.. if you want to install explicitly all packages.. My objection is for an user to do less..

There will always be some binary packages to install, and then you use either `apt`, `yum` or `pkg` depending on OS (never more than one). There will always be the Zonemaster packages to install from CPAN with `cpanm`. 

With the model we have now I think we have more control, but we could change for next release.

mattias-p:
> Then what is the use of tools like apt-get, yum, cpan etc.. if you want to install explicitly all packages..

With regard to OS package managers (binary packages), again, I see two major benefits. Installation is faster and quality control is better. Both of these are immediately beneficial to the user.
With regard to `cpanm`, the immediate benefit is to the maintainers. Having this list helps maintainers to grow the list of OS-managed packages (binary packages), which in turn is beneficial to the user.

> My objection is for an user to do less..

Cutting and pasting two commands more is a small price to pay for faster install times and fewer malfunctions.

> Tomorrow if we package Zonemaster are you going to install all dependencies or let the package manager to do its job.

Yes. Once we have OS-managed packages for Zonemaster we should recommend that people use those. Listing dependencies in such an instruction would be pointless. I agree with that.

sandoche2k:
Can be closed if PR #720 is closed

docs/Installation.md
```diff
@@ -24,31 +24,48 @@ This instruction covers the following operating systems:
...
 ### Installation on CentOS
 
-1) Install the [EPEL 7][EPEL] repository:
+1) *Only* for CentOS8, enable powertools:
+
+   ```sh
+   sudo yum config-manager --set-enabled PowerTools
+   ```
+
+2) Install the [EPEL][EPEL] repository:
```
mattias-p, line 33: `[EPEL][EPEL]` is redundant. `[EPEL]` is equivalent but looks better.

docs/Installation.md
```diff
@@ -24,31 +24,48 @@ This instruction covers the following operating systems:
...
+3) Make sure the development environment is installed:
 
    ```sh
    sudo yum groupinstall "Development Tools"
    ```
 
-3) Install binary packages:
-
+4) Install binary packages:
+
```
mattias-p, line 46: Could you remove this whitespace? Unnecessary whitespace is a cause of unnecessary merge conflicts.

docs/Installation.md
```diff
@@ -24,31 +24,48 @@ This instruction covers the following operating systems:
...
    sudo yum groupinstall "Development Tools"
    ```
 
-3) Install binary packages:
-
+4) Install binary packages:
+   
    ```sh
-   sudo yum install cpanminus libidn-devel openssl-devel perl-Clone perl-core perl-Devel-CheckLib perl-Email-Valid perl-File-ShareDir perl-File-Slurp perl-libintl perl-IO-Socket-INET6 perl-JSON-PP perl-List-MoreUtils perl-Module-Find perl-Moose perl-Net-IP perl-Pod-Coverage perl-Readonly-XS perl-Test-Differences perl-Test-Exception perl-Test-Fatal perl-Test-Pod perl-Text-CSV perl-YAML
+   sudo yum install cpanminus libidn-devel openssl-devel perl-Clone perl-core perl-Devel-CheckLib perl-File-ShareDir perl-File-Slurp perl-libintl perl-IO-Socket-INET6 perl-JSON-PP perl-List-MoreUtils perl-Module-Find perl-Moose perl-Net-IP perl-Pod-Coverage perl-Test-Differences perl-Test-Exception perl-Test-Fatal perl-Test-Pod perl-Text-CSV perl-YAML perl-MailTools
```
mattias-p, line 48: Please remove perl-MailTools. We don't use it. If necessary grep for MailTools in the source code.

docs/Installation.md
```diff
@@ -24,31 +24,48 @@ This instruction covers the following operating systems:
 
 ### Installation on CentOS
 
-1) Install the [EPEL 7][EPEL] repository:
+1) *Only* for CentOS8, enable powertools:
```
mattias-p, line 27: Have you considered adding a space between CentOS and 8 to make it more readable?

docs/Installation.md
```diff
@@ -24,31 +24,48 @@ This instruction covers the following operating systems:
...
    ```sh
-   sudo yum install cpanminus libidn-devel openssl-devel perl-Clone perl-core perl-Devel-CheckLib perl-Email-Valid perl-File-ShareDir perl-File-Slurp perl-libintl perl-IO-Socket-INET6 perl-JSON-PP perl-List-MoreUtils perl-Module-Find perl-Moose perl-Net-IP perl-Pod-Coverage perl-Readonly-XS perl-Test-Differences perl-Test-Exception perl-Test-Fatal perl-Test-Pod perl-Text-CSV perl-YAML
+   sudo yum install cpanminus libidn-devel openssl-devel perl-Clone perl-core perl-Devel-CheckLib perl-File-ShareDir perl-File-Slurp perl-libintl perl-IO-Socket-INET6 perl-JSON-PP perl-List-MoreUtils perl-Module-Find perl-Moose perl-Net-IP perl-Pod-Coverage perl-Test-Differences perl-Test-Exception perl-Test-Fatal perl-Test-Pod perl-Text-CSV perl-YAML perl-MailTools
    ```
 
-4) Install packages from CPAN:
+5) Install packages from CPAN:
 
    ```sh
    sudo cpanm Locale::Msgfmt Module::Install Module::Install::XSUtil Test::More
```
mattias-p, line 54: Please add Email::Valid and Readonly::XS. We use those.

docs/Installation.md
mattias-p, line 27: Have you considered adding a space between CentOS and 8 to improve readability?

docs/Installation.md
mattias-p, line 33: `[EPEL][EPEL]` can be expressed more concisely as `[EPEL]`.

docs/Installation.md
mattias-p, line 48: Please remove perl-MailTools. We don't use it. Verify by grepping in the source code for `Mail::`.

docs/Installation.md
mattias-p, line 54: Please add Email::Valid and Readonly::XS. We use those.

docs/Installation.md
mattias-p, line 46: Could you remove this whitespace?

docs/Installation.md
sandoche2k: 👍

docs/Installation.md
sandoche2k: We discussed it in #683. Not able to install in CentOS 8 without this package.

docs/Installation.md
sandoche2k: Please refer to the discussion in https://github.com/zonemaster/zonemaster-engine/pull/683

docs/Installation.md
sandoche2k: Do not see any additional whitespace

docs/Installation.md
sandoche2k: 👍

docs/Installation.md
matsduf: But this *IS* PR #683. Or what do you refer to?

docs/Installation.md
```diff
@@ -24,31 +24,48 @@ This instruction covers the following operating systems:
...
    ```
 
-5) Install Zonemaster::LDNS and Zonemaster::Engine:
+6) Install Zonemaster::LDNS and Zonemaster::Engine for *CentOS7*:
+
+   ```sh
+   sudo cpanm Zonemaster::LDNS --configure-args="--no-ed25519"
+   ```
```
matsduf, line 61: I suggest a note such as "Please note that DNSSEC Algorithm 15 is not supported on CentOS7"

docs/Installation.md
sandoche2k: There was another https://github.com/zonemaster/zonemaster-engine/pull/720. I will close that

docs/Installation.md
```diff
@@ -24,31 +24,49 @@ This instruction covers the following operating systems:
...
    sudo yum groupinstall "Development Tools"
    ```
 
-3) Install binary packages:
-
+4) Install binary packages:
+   
    ```sh
-   sudo yum install cpanminus libidn-devel openssl-devel perl-Clone perl-core perl-Devel-CheckLib perl-Email-Valid perl-File-ShareDir perl-File-Slurp perl-libintl perl-IO-Socket-INET6 perl-JSON-PP perl-List-MoreUtils perl-Module-Find perl-Moose perl-Net-IP perl-Pod-Coverage perl-Readonly-XS perl-Test-Differences perl-Test-Exception perl-Test-Fatal perl-Test-Pod perl-Text-CSV perl-YAML
+   sudo yum install cpanminus libidn-devel openssl-devel perl-Clone perl-core perl-Devel-CheckLib perl-File-ShareDir perl-File-Slurp perl-libintl perl-IO-Socket-INET6 perl-JSON-PP perl-List-MoreUtils perl-Module-Find perl-Moose perl-Net-IP perl-Pod-Coverage perl-Test-Differences perl-Test-Exception perl-Test-Fatal perl-Test-Pod perl-Text-CSV perl-YAML perl-MailTools
```
matsduf, line 48: What does perl-MailTools provide? `Email::Valid` The latter is required. At this stage I will approve, but I think we should review the installation document again after the release.

docs/Installation.md
mattias-p: There are three SPACE (U+0020) characters on line 46. Github highlights them in a darker shade of green.

mattias-p:
I believe Travis is failing because of a bug in Travis. I usually resolve the problem by adding an empty commit to the pull request. (You can create an empty commit with `git commit --allow-empty -mEmpty`)

docs/Installation.md
sandoche2k: 👍

docs/Installation.md
sandoche2k: I don't see any difference. Anyway, we will update next time

PR #685, opened 2020-03-13
List IPv6 addresses in IPv6 message

quinot:
Fixes issue mentioned on #523: IPv4 addresses are listed instead of IPv6 addresses in the [NOT_]ENOUGH_IPV6_NS_CHILD messages.

Issue #545, opened 2019-05-18
Incorrect message from the Delegation module


matsduf:
When there are too few IPv6 name servers the Delegation module creates the following  message referring to IPv4 addresses:
```
   4.96 ERROR     Child does not list enough (1) nameservers that resolve to IPv6 addresses (46.253.202.186;79.99.1.90;82.102.5.100). Lower limit set to 2.
```

The error was found on an installation based on the develop branch on commit 3d0c148 and was found in the following test:
```
# zonemaster-cli pearldragon.se --level info --no-ipv6 --test delegation
Seconds Level     Message
======= ========= =======
   0.00 INFO      Using version v3.0.2 of the Zonemaster engine.
   1.78 INFO      Parent lists enough (3) nameservers (kif.hildingsson.se;lakridstrollet.regexp.se;ns.narnia.pp.se). Lower limit set to 2.
   4.95 INFO      Child lists enough (3) nameservers (kif.hildingsson.se;lakridstrollet.regexp.se;ns.narnia.pp.se). Lower limit set to 2.
   4.96 INFO      Child lists enough (3) nameservers that resolve to IPv4 addresses (46.253.202.186;79.99.1.90;82.102.5.100). Lower limit set to 2.
   4.96 ERROR     Child does not list enough (1) nameservers that resolve to IPv6 addresses (46.253.202.186;79.99.1.90;82.102.5.100). Lower limit set to 2.
   5.43 INFO      Delegation lists enough (3) nameservers that resolve to IPv4 addresses (46.253.202.186;79.99.1.90;82.102.5.100). Lower limit set to 2.
   5.43 ERROR     Delegation does not list enough (1) nameservers that resolve to IPv6 addresses (2a02:750:7::4c6). Lower limit set to 2.
(...)
```

vlevigneron:
The fix has already been merged.

matsduf:
Fixed by #552.

PR #686, opened 2020-03-18
Add test case for listing IPv6 addresses in IPv6 message

quinot:
Regression test for #545
As discussed on #685


matsduf:
@vlevigneron, should this be included in v2019.2 or wait for the release after that?

sandoche2k:
@vlevigneron can you approve and merge ?

Issue #699, opened 2020-04-02
Test.pm needs update


matsduf:
[Test.pm] (here linked to develop branch) refer to `Mail::RFC822::Address` which is not used anywhere in the Engine code.

```
info( DEPENDENCY_VERSION => { name => 'Mail::RFC822::Address', version => $Mail::RFC822::Address::VERSION } );
```

Could that be removed? Anything else in that list?

[Test.pm]: https://github.com/zonemaster/zonemaster-engine/blob/develop/lib/Zonemaster/Engine/Test.pm

mattias-p:
Good digging. Yeah, Mail::RFC822::Address was replaced with Email::Valid in #547 for packaging reasons.

This entire list of dependencies should match what's listed in Makefile.PL as `requires` and `recommends`.

matsduf:
Resolved by #700

matsduf:
Resolved by #700.

PR #700, opened 2020-04-02
Remove useless dependency

vlevigneron:
- Fixes zonemaster/zonemaster-engine#699


vlevigneron:
@mtas the goal was to remove Mail::RFC822::Address dependency but as perlcritic raised severity 4 warnings about package import, I chose to fix that too. But it seems it has Travis drawbacks so I revert some part of the initial commit.

PR #708, opened 2020-04-15
Clarified the instructions for the translators

matsduf:
I have added some more instructions to the translators.

### DIFF
--- a/docs/Translation.pod
+++ b/docs/Translation.pod
@@ -76,12 +76,12 @@ The first step in updating the translations is to generate a new template file
 (F<Zonemaster-Engine.pot>).
 In practice you rarely need to think about generating it as it is generally
 performed as an implicit intermediate step.
-If you do want to generate it, the command is C<make extract-pot> 
+If you do want to generate it, the command is C<make extract-pot>
 (C<gmake extract-pot> on FreeBSD).
 
 The translated strings are maintained in files named C<{language_code}.po>
 (currently F<en.po>, F<sv.po>, F<da.po> or F<fr.po>).
-Execute C<make update-po> (C<gmake update-po> on FreeBSD) to update these 
+Execute C<make update-po> (C<gmake update-po> on FreeBSD) to update these
 files with new message ids from the
 source code (C<make extract-pot> will be infoked behind the scenes).
 This should only be necessary to do when a developer has added or changed a test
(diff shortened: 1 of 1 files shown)


docs/Translation.pod
```diff
@@ -133,16 +133,33 @@ C<make update-po> (C<gmake update-po> on FreeBSD).
...
 
+=item * Inspect every every message tagged as "fuzzy". Update F<msgstr> if needed
+and remove the "fuzzy" mark.
+
+=item * Search for messages that have empty F<msgstr> and add a translation. At
+the end of the file there could be lines starting with "#~" and those could have
+matching translations.
+
+=item * Any remaining lines at the end of the file starting with "#~" could be
+removed.
```
mattias-p, line 147: IMHO it would be a further improvement to use gettext terminology for the respective entry types. I.e. "fuzzy entry", "untranslated entry" and "obsolete entry".

matsduf: Good suggestion. I've worked it in. I also added a short section on tools and removed trailing spaces.

docs/Translation.pod
```diff
@@ -101,50 +101,81 @@ create one, contact the Zonemaster work group for instructions, either by
...
+The ".po" can be edited with a plain text editor, but then it is important
+to keep the database structur of the file.
+
+There are tools that makes editing of the ".po" files easier. When using
+those, the ".po" file is handled rather as a database then a plain file.
+
+There is an add-on to Emacs, L<https://www.emacswiki.org/emacs/PoMode>,
+which makes updating and searching in the ".po" file easier and more robust.
+
+F<This section needs expansion with more details.>
```
mattias-p, line 122: Can't we just link to https://www.gnu.org/software/gettext/manual/html_node/PO-Mode.html#PO-Mode?

matsduf: None of the references are good, but I changed. It is surprisingly hard to install po-mode under emacs. But I found "GNOME Translation Editor" which feels nice for a graphical environment. @mattias-p, please re-review.

mattias-p: The links to PO mode and GNOME Translation Editor look good to me.

I'm not sure what more details are needed here. Could we replace this note with an issue with a clearer scope?

matsduf:
@sandoche2k, please review.

Issue #697, opened 2020-03-30
share/Makefile not compatible with FreeBSD "make"


matsduf:
The [share/Makefile] (in current develop branch) is not compatible with FreeBSD `make`(which is *not* GNU Make). The make file contains the GNU Make macros `$(shell ...)` and `$(wildcard ...)` which are not supported by FreeBSD `make`.

The [share/Makefile] is not only used for creating distribution files, as documented in the [ReleaseProcess] document, but is also use for translation work, as documented in the [Translation] document.

A work-around could be to use FreeBSD `gmake` (which *is* GNU Make), but then it should be documented.

All links to documents are to the develop branch of that repository.

[share/Makefile]: https://github.com/zonemaster/zonemaster-engine/blob/develop/share/Makefile
[ReleaseProcess]: https://github.com/zonemaster/zonemaster/blob/develop/docs/internal-documentation/maintenance/ReleaseProcess.md
[Translation]: https://github.com/zonemaster/zonemaster-engine/blob/develop/docs/Translation.pod

mattias-p:
Has this worked before? The `$(shell)` function is new since last release but `$(wildcard)` has been there since 2014. How the `share/Makefile` is invoked has changed over time.

matsduf:
It has probably not worked before either.

matsduf:
PR #726 will resolve this issue.

PR #726, opened 2020-05-05
Updates to translation tooling

mattias-p:
The main improvement here is POSIX compliance of share/Makefile, which means we don't need to use gmake on FreeBSD. Documentation is updated accordingly.

Along with it comes various clean ups related to translation tooling.

### DIFF
--- a/docs/Translation.pod
+++ b/docs/Translation.pod
@@ -60,7 +60,7 @@ To be written.
 
 Install the following:
 
-C<pkg install gettext git>
+C<apt install gettext git>
 
 =head4 FreeBSD
 
--- a/docs/Translation.pod
+++ b/docs/Translation.pod
@@ -48,7 +48,7 @@ Make sure the message tag comments are properly added and up to date.
 
 =head3 Software preparation
 
-For the steps below you need to work on a computer with Git, Perl and Xgettext.
+For the steps below you need to work on a computer with Git, Perl and Gettext.
 Select what OS you want to work on. Other OSs will also work, but you will
 have to find instructions elsewhere.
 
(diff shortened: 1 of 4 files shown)


docs/Translation.pod
```diff
@@ -48,41 +48,41 @@ Make sure the message tag comments are properly added and up to date.
...
+=head4 CentOS
+
+To be written.
+
+=head4 Debian
 
 Install the following:
 
-C<pkg install git-lite devel/p5-Locale-XGettext devel/p5-Locale-Msgfmt devel/gmake>
+C<pkg install gettext git>
```
matsduf, line 63: This line is for Debian. "sudo apt-get install"?

mattias-p:
> I propose that a comment is inserted in Makefile that is should be kept in Posix format to be compliant with both Gnu and BSD.

I'll try to think of a way to have it automatically checked.

> When (if) this PR is merged, then ReleaseProcess document should also be updated.

You mean adding an explicit step to manually verify share/Makefile is POSIX still compliant?

docs/Translation.pod
mattias-p: Good catch. I'll update it.

matsduf:
> > When (if) this PR is merged, then ReleaseProcess document should also be updated.
> 
> You mean adding an explicit step to manually verify share/Makefile is POSIX still compliant?

No, simpler than that. The [ReleaseProcess] document also have special instructions for FreeBSD med `gmake` etc. That should be cleaned.

[ReleaseProcess]: https://github.com/zonemaster/zonemaster/blob/master/docs/internal-documentation/maintenance/ReleaseProcess.md

mattias-p:
Good point. I didn't realize.

mattias-p:
The .mo files directly in share help `make` in its target resolution. The ones below share/locale are input files for `make dist` and `make distcheck`. I'll make an update to explicitly exclude the .mo files directly in share from the manifest.

mattias-p:
Can this be merged now or are we still making bug fix releases off of the develop branch?

matsduf:
I suggest that we wait until Thursday 2020-05-28 before we merge anything else to the develop branch.

matsduf:
Issue #697 also relates to this.

Issue #727, opened 2020-05-12
Zonemaster crasches in DNSSEC13


matsduf:
When running Zonemaster (version v2019.2), it crashed during test av [DNSSEC13]. Here as running `zonemaster-cli` but it has been verified in Zonemaster-Backend too.

```
$ zonemaster-cli ip.se --no-ipv6 --test DNSSEC/dnssec13 --show_module --raw --level INFO
   0.01 INFO      SYSTEM       SYSTEM:GLOBAL_VERSION version=v3.0.3
Use of uninitialized value in subroutine entry at /usr/local/lib/x86_64-linux-gnu/perl/5.26.1/Zonemaster/LDNS/RR/RRSIG.pm line 12.
   3.79 ERROR     DNSSEC       DNSSEC:RRSIG_BROKEN address=162.219.54.128; keytag=5134; ns=ns1.anycast.iis.se; rrtype=DNSKEY
Segmentation fault (core dumped)
```
(Disregard `--no-ipv6`.)

The consequence for `zonemaster-cli` is that the test breaks and ends with "core dump".

The consequence for Zonemaster-Backend is that the specific testagent process that runs the test will crash, but the primary `testagent` process will continue. The test will never be completed, but neither the `testagent` function nor the `RPC-API` function will be affected.

[DNSSEC13]: https://github.com/zonemaster/zonemaster/blob/704829668719ff61ed7d99da597c1f4241bb4c8a/docs/specifications/tests/DNSSEC-TP/dnssec13.md

matsduf:
Resolved by #728.

matsduf:
The bug was triggered when the zone had a RRSIG referring to absent DNSKEY.

PR #728, opened 2020-05-12
Remove empty elements from @keys to resolve issue #727

matsduf:
Resolves issue #727.

`Zonemaster::LDNS::RR::RRSIG::verify()` requires that incoming parameter `$keys` is a reference to a non-empty array of DNSKEYs. Under some circumstances,  `@keys` in Zonemaster::Engine::Test::DNSSEC::dnssec13() is created as an array of an empty key, and is then validated as non-empty and used in a call to `Zonemaster::LDNS::RR::RRSIG::verify()`.

This update removes any empty element from `@keys` after it is created.

### DIFF
--- a/lib/Zonemaster/Engine/Test/DNSSEC.pm
+++ b/lib/Zonemaster/Engine/Test/DNSSEC.pm
@@ -5,7 +5,7 @@ use 5.014002;
 use strict;
 use warnings;
 
-use version; our $VERSION = version->declare( "v1.1.16" );
+use version; our $VERSION = version->declare( "v1.1.17" );
 
 ###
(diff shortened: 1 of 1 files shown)


lib/Zonemaster/Engine/Test/DNSSEC.pm
```diff
@@ -1880,6 +1880,14 @@ sub dnssec13 {
 
             foreach my $sig ( @sigs ) {
                 my @keys = ($keytags{$sig->keytag});
```
vlevigneron, line 1882: I guess the following line would be enough:
my @keys = grep { $_ } ($keytags{$sig->keytag});

PR #736, opened 2020-05-19
Remove the DISABLED feature

### LINKED ISSUE #454: Special key DISABLED in test_levels
The `test_levels` option supports a special key `DISABLED`. It's somewhat similar to the `test_cases` option but works on the the module level. The `DISABLED` key is undocumented and to my knowledge the only place it is used is in the "disabled" "Example" in policy.json/profile.json.

If this feature is worth its weight, it should be documented.

See also #456.

mattias-p:
This is a proposal to fix #454 by removing the DISABLED feature.

This is a breaking change.

This also includes a drive by update to the installation instruction where a dependency was forgotten.


docs/Installation.md
```diff
@@ -43,15 +43,15 @@ This instruction covers the following operating systems:
...
+
    ```sh
    sudo yum install cpanminus libidn-devel openssl-devel perl-Clone perl-core perl-Devel-CheckLib perl-File-ShareDir perl-File-Slurp perl-libintl perl-IO-Socket-INET6 perl-JSON-PP perl-List-MoreUtils perl-Module-Find perl-Moose perl-Net-IP perl-Pod-Coverage perl-Test-Differences perl-Test-Exception perl-Test-Fatal perl-Test-Pod perl-Text-CSV perl-YAML perl-MailTools
    ```
 
 5) Install packages from CPAN:
 
    ```sh
-   sudo cpanm Locale::Msgfmt Module::Install Module::Install::XSUtil Test::More
+   sudo cpanm Email::Valid Locale::Msgfmt Module::Install Module::Install::XSUtil Test::More
```
matsduf, line 54: Is this change connected to the DISABLE feature?

mattias-p: No, it's the drive-by update of the installation instruction that I mentioned in the description.

matsduf: Why not in a separate PR since this PR does not change anything to Installation.md? That will make the Changes file clearer.

mattias-p: I don't believe it's warranted to mention drive-by changes in the Changes. If I thought it was worth a mention in the Changes file I wouldn't have called it a drive-by change. Nor would I have included it in this PR.

matsduf: @mattias-p, you usually argue that we should keep changes separate and clean. When we look back at the changes of the installation document there would be an unrelated pull request.

mattias-p:
Please @matsduf. I do not argue that updates to the installation instruction should go into the change log, and even if I ever have, tū quoque is not a valid argument.

This change does not warrant a mention in the change log. If you want to know all changes for the installation instruction for a given release `git log PREVIOUS..VERSION Changes` will happily tell you. If you to know which PR those commits were merged with, git will answer that as well.

docs/Installation.md
mattias-p: Please @matsduf. I do not argue that updates to the installation instruction should go into the change log, and even if I ever have, tū quoque is not a valid argument.

This change does not warrant a mention in the change log. If you want to know all changes for the installation instruction for a given release git log PREVIOUS..VERSION Changes will happily tell you. If you to know which PR those commits were merged with, git will answer that as well.

Issue #734, opened 2020-05-15
Zone10 does not ignore case when comparing domain names


matsduf:
In [Zone10], the owner name of the SOA record is compared with the zone name tested. The comparison should be done after neutralizing any case difference, which is not done in the implementation.

_This issue does not only hit zones (domains) with capital letters in owner name of the SOA record in the zone file. It hits any domain if you enter the domain name with case setting than the SOA record has, e.g. testing "Afnic.fr" will give the same error._

```
$ zonemaster-cli --show_module --raw arpa --test ZONE/zone10
   1.54 ERROR     ZONE         ZONE:WRONG_SOA address=199.9.14.201; name=arpa.; ns=b.root-servers.net; owner=ARPA.
   1.57 ERROR     ZONE         ZONE:WRONG_SOA address=2001:500:200::b; name=arpa.; ns=b.root-servers.net; owner=ARPA.
   1.61 ERROR     ZONE         ZONE:WRONG_SOA address=192.33.4.12; name=arpa.; ns=c.root-servers.net; owner=ARPA.
   1.68 ERROR     ZONE         ZONE:WRONG_SOA address=2001:500:2::c; name=arpa.; ns=c.root-servers.net; owner=ARPA.
   1.83 ERROR     ZONE         ZONE:WRONG_SOA address=192.112.36.4; name=arpa.; ns=g.root-servers.net; owner=ARPA.
   6.88 ERROR     ZONE         ZONE:WRONG_SOA address=2001:500:12::d0d; name=arpa.; ns=g.root-servers.net; owner=ARPA.
   7.11 ERROR     ZONE         ZONE:WRONG_SOA address=2001:dc3::35; name=arpa.; ns=m.root-servers.net; owner=ARPA.
   7.14 ERROR     ZONE         ZONE:WRONG_SOA address=202.12.27.33; name=arpa.; ns=m.root-servers.net; owner=ARPA.
```

With some more details:

```
$zonemaster-cli --show_module --raw arpa --level DEBUG3 --test ZONE/zone10
   0.00 DEBUG     SYSTEM       SYSTEM:START_TIME string=2020-05-15 20:59:51 +0200; time_t=1589569191
   0.00 DEBUG     SYSTEM       SYSTEM:TEST_ARGS args=Zonemaster::Engine::Zone=HASH(0x56360be59e88); method=zone10; module=ZONE
[...]

PR #737, opened 2020-05-20
- Fixes case issue when comparing domain name in Zone10

vlevigneron:
Fixes zonemaster/zonemaster-engine#734

### DIFF
--- a/lib/Zonemaster/Engine/Test/Zone.pm
+++ b/lib/Zonemaster/Engine/Test/Zone.pm
@@ -5,7 +5,7 @@ use 5.014002;
 use strict;
 use warnings;
 
-use version; our $VERSION = version->declare( "v1.0.11" );
+use version; our $VERSION = version->declare( "v1.0.12" );
 
 use Zonemaster::Engine;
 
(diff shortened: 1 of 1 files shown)


lib/Zonemaster/Engine/Test/Zone.pm
```diff
@@ -699,14 +699,14 @@ sub zone10 {
                         }
                       );
                 }
-                elsif ( $soa[0]->owner ne $name->fqdn ) {
+                elsif ( lc( $soa[0]->owner ) ne lc( $name->fqdn ) ) {
```
mattias-p, line 702: I note that `lc` can use either ASCII rules or Unicode rules depending on context. Are we sure that either ASCII rules will always be applied here, or that `LATIN SMALL LETTER I` is lowercase of `LATIN CAPITAL LETTER I` even in Turkish locales?

Also, case insensitivity only applies to ASCII labels, right? `lc` operates character by character, so this will fail to catch some errors for some binary labels. But maybe we don't care about binary labels?

matsduf: No, in Turkish locale lower case of `LATIN CAPITAL LETTER I` is `LATIN SMALL LETTER DOTLESS I`.

It is little bit unclear, but in practice we disregard binary labels. It is also unclear about IDN labels, but if they were entered as U-labels they are converted to A-labels and can here be treated as ASCII labels.

It is good that you raise the issue, and it should be handled, but this fix is to the better rather than the worse.

matsduf:
@mattias-p, can you approve? We should maybe create some issues to remind us to straighten out some of the important issues that you raise?

matsduf:
@vlevigneron, please merge.

PR #745, opened 2020-05-29
Remove en.po

### LINKED ISSUE #704: Remove the need for en.po
`msgid` is the text message in English that comes from the Perl modules. For French, Swedish and Danish that is translated to a `msgstr` in the target language in the designated `.po` file. There is also a `en.po` for "translation" from English to English. In most cases the `msgstr` is identical to the `msgid`, i.e. nothing is changed. In a number of cases, the `msgstr` contains an improved or corrected version of the `msgid`. The correct process is to correct the `msgid` instead, but that is not the way it has been done in some cases. There are a number of consequences of this situation:

* The translation work (by the translators) is naturally done against the `msgid` in English because that is what there is in the `.po` files. This means that the translation is done against the message without corrections or improvements.
* If translation does not work for some reason, the `msgid`strings are printed, without correction and improvement.
* The `en.po` file must be maintained even though there is no use for it. Or it could be generated where all `msgstr` strings are identical to the `msgid` string it belong to.

Solution:
* Capture all differences between `msgid` and `msgstr` in `en.po`. 
* Update the `msgid` in the Perl module when that is relevant.
* Update all other `.po` files so that new `msgid` strings match the correct `msgstr`.
* Remove `en.po` or generate it from the Perl modules.
* Change all processes so that `en.po` is never manually updated.

This is an expanded version of #635.

mattias-p:
This change is implemented in four commits to make it easier/possible to review for correctness.

A notable change auxiliary to #704 is the change of the order of messages in the PO files is changed from by msgid to by source location. This makes it easier to review changes where several msgids are updated.

Fixes #704.


mattias-p:
I rebased this to make the git history less of a mess. Could you review again? Sorry for the inconvenience.

PR #749, opened 2020-06-16
Make locale not get stuck on FreeBSD 11

### LINKED ISSUE #748: Translation language gets stuck on FreeBSD 11 (Engine)
Selecting language does not work correctly on FreeBSD 11. After you construct the Zonemaster::Engine::Translator instance and call $translator->locale() with a locale argument, the result is correctly translated to that language. However, if you keep making those calls the translation language argument is no longer respected. Instead locale get stuck on the value from a previous call. 

Fixing this should also fix zonemaster/zonemaster-backend#530.

mattias-p:
This makes Z::E::Translator properly communicate locale to gettext.

Fixes #748 and zonemaster/zonemaster-backend#530.

This PR is best reviewed commit by commit.

### DIFF
--- a/lib/Zonemaster/Engine/Translator.pm
+++ b/lib/Zonemaster/Engine/Translator.pm
@@ -16,9 +16,10 @@ use Readonly;
 
 use Moose;
 
-has 'locale' => ( is => 'rw', isa => 'Str' );
-has 'data' => ( is => 'ro', isa => 'HashRef', lazy => 1, builder => '_load_data' );
+has 'locale'               => ( is => 'rw', isa => 'Str' );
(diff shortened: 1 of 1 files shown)


lib/Zonemaster/Engine/Translator.pm
```diff
@@ -118,6 +119,11 @@ sub BUILD {
     my ( $self ) = @_;
 
     my $locale = $self->{locale} // _get_locale();
+
+    # Make sure LC_MESSAGES can be effectively set down the line.
+    delete $ENV{LANGUAGE};
```
mattias-p, line 124: I'm having second thoughts on clearing LANGUAGE here. It makes it impossible to create applications that respect LANGUAGE in a clean way. I'll make an update to fix that.

PR #755, opened 2020-06-30
Make Z::E::Translator into a proper singleton

mattias-p:
This makes misuse of Z::E::Translator->new() difficult (impossible?), rather than merely discouraged.


matsduf:
I will merge on the request of @mattias-p.

Issue #721, opened 2020-04-29
Missing msgid


matsduf:
The message tags `DNSSEC:NO_RESPONSE_DS` and `DNSSEC:KEY_SIZE_OK`, respectively, do not return any message id.

The test was run where IPv6 was not available, but IPv6 was not disabled.

```
$ zonemaster-cli ed25519.nl --show_module --level INFO --locale en_US.UTF-8 | egrep "[A-Z]:[A-Z]" 
  12.85 WARNING   DNSSEC       DNSSEC:NO_RESPONSE_DS address=2001:678:2c:0:194:0:28:53; ns=ns1.dns.nl; rrtype=DS; zone=ed25519.nl
  12.87 WARNING   DNSSEC       DNSSEC:NO_RESPONSE_DS address=2001:67c:1010:10::53; ns=ns2.dns.nl; rrtype=DS; zone=ed25519.nl
  12.91 WARNING   DNSSEC       DNSSEC:NO_RESPONSE_DS address=2001:678:20::24; ns=ns3.dns.nl; rrtype=DS; zone=ed25519.nl
  14.15 INFO      DNSSEC       DNSSEC:KEY_SIZE_OK 
```

PR #758, opened 2020-07-23
DNSSEC module MsgID missing + MsgID Typo Fixed

vlevigneron:
fixes zonemaster/zonemaster-engine#721

### DIFF
--- a/lib/Zonemaster/Engine/Test/DNSSEC.pm
+++ b/lib/Zonemaster/Engine/Test/DNSSEC.pm
@@ -5,7 +5,7 @@ use 5.014002;
 use strict;
 use warnings;
 
-use version; our $VERSION = version->declare( "v1.1.17" );
+use version; our $VERSION = version->declare( "v1.1.18" );
 
 ###
 ### This test module implements DNSSEC tests.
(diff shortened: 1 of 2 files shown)


lib/Zonemaster/Engine/Test/DNSSEC.pm
```diff
@@ -669,6 +669,10 @@ Readonly my %TAG_DESCRIPTIONS => (
         __x    # DNSSEC:KEY_DETAILS
           'Key with keytag {keytag} details : Size = {keysize}, Flags ({sep}, {rfc5011}).', @_;
     },
+    KEY_SIZE_OK => sub {
+          __x    # DNSSEC:KEY_SIZE_OK
+            'All keys from apex DNSKEY RRset have a correct size', @_;
```
matsduf, line 674: "All keys from the DNSKEY RRset have the correct size."

lib/Zonemaster/Engine/Test/DNSSEC.pm
```diff
@@ -669,6 +669,10 @@ Readonly my %TAG_DESCRIPTIONS => (
         __x    # DNSSEC:KEY_DETAILS
           'Key with keytag {keytag} details : Size = {keysize}, Flags ({sep}, {rfc5011}).', @_;
     },
+    KEY_SIZE_OK => sub {
+          __x    # DNSSEC:KEY_SIZE_OK
+            'All keys from the DNSKEY RRset have the correct size', @_;
```
matsduf, line 674: It is a very small detail, but I think we should be consistent with the final period '.' in the messages. I am sorry I was not more explicit in my previous comment.

Issue #578, opened 2019-07-31
Extend messages with test case


matsduf:
### Background

A message outputted from a test case is local to the module, e.g. NO_RECURSOR belongs to module NAMESERVER and not to test case Nameserver01. This has some consequences:

1. A message cannot be traced back to the test case and its specification.
2. The same message tag (e.g. NO_RECURSOR) within the same module will have the same level, e.g. ERROR or INFO, but when the same message tag is used in different modules they are independent.

### Proposal

The test case is added as parameter to the message.

```
  {
    "args": {
      "address": "2001:67c:124c:100a::45",
      "ns": "ns.nic.se"
    },
    "level": "INFO",
    "module": "NAMESERVER",
    "tag": "NO_RECURSOR",
    "timestamp": 8.46649193763733
  },
```

Will be fixed by #762.

matsduf:
> I chose 'test_case' for the moment but it could be 'method' which is also used elsewhere in documentation.

I think that "test case" is the correct reference since that is what we call the specifications in Zonemaster/Zonemaster. Where do you see "method"?

I suggest "testcase" instead of "test_case" to make it easier to read and write, just as "time stamp" is written as "timestamp".

vlevigneron:
> > I chose 'test_case' for the moment but it could be 'method' which is also used elsewhere in documentation.
> 
> I think that "test case" is the correct reference since that is what we call the specifications in Zonemaster/Zonemaster. Where do you see "method"?
> 

In CLI documentation, it is written : 
`'Specify test to run. Should be either the name of a module, or the name of a module and the name of a method in that module separated by a "/" character (Example: "Basic/basic1"). The method specified must be one that takes a zone object as its single argument. This switch can be repeated.'
`
> I suggest "testcase" instead of "test_case" to make it easier to read and write, just as "time stamp" is written as "timestamp".

OK, if we decide it is not  "method" I will use "testcase" with no space.

matsduf:
@vlevigneron, good catch of the inconsistency of terminology. We should update to be consistent. What document so that I can create an issue, or you do it?

vlevigneron:
@matsduf https://github.com/zonemaster/zonemaster-cli/blob/master/lib/Zonemaster/CLI.pm#L163

matsduf:
Issue for the terminology inconsistency is created as 
https://github.com/zonemaster/zonemaster-cli/issues/123.

matsduf:
Proposal accepted by F2F.

[...]

PR #762, opened 2020-07-23
- Add method name (test case) in Logger.

vlevigneron:
- Add method name (test case) in Logger.
- Fixes zonemaster/zonemaster-engine#578

DO NOT MERGE NOW, do a test first to check if it is what you are waiting for

### DIFF
--- a/lib/Zonemaster/Engine/Logger.pm
+++ b/lib/Zonemaster/Engine/Logger.pm
@@ -1,10 +1,15 @@
 package Zonemaster::Engine::Logger;
 
-use version; our $VERSION = version->declare("v1.0.7");
-
 use 5.014002;
+
+use strict;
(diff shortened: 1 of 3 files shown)


mattias-p (approved):
The code looks good to me.

matsduf:
@sandoche2k,  can you ask @pnax, to verify if this will be a problem for backend or not?

vlevigneron:
Just to chek before I merge. Can you confirm that we decided to merge this PR even if we are not sure of the impacts in CLI and Backend ??

matsduf:
> Just to chek before I merge. Can you confirm that we decided to merge this PR even if we are not sure of the impacts in CLI and Backend ??

@vlevigneron, please merge. We will have to test anyway.

Issue #665, opened 2019-11-01
Include start and end message for each test case


matsduf:
When debugging with the help of CLI it would be valuable to be able to see the start and end of the execution of a test case. To be able to see that I suggest that the following code is added to each test module perl module:

```
TEST_CASE_START => sub {
    __x    # XXX:TEST_CASE_START
   'TEST_CASE_START {testcase}', @_;
},
```
And then as the very start of every test case method:

```
my $testcase = "EXAMPLE01";
push @results,
        info(
                TEST_CASE_START  => {
                      testcase  => $testcase,
                     }
                );
```

If it can be done in a reasonably way, a "TEST_CASE_END" with the same format should also be added so that it always comes as the last thing executed for the test case method.

The default level of these messages should be DEBUG.

matsduf:
Agreed at F2F.

Issue #732, opened 2020-05-13
Remove SYSTEM:MODULE_CALL


matsduf:
Update to this issue:

These MODULE_CALL is too limited and will instead be replaced with what is being implemented in #764 to resolve #665. Then MODULE_CALL has no meaning and should be removed. This is resolved in #767.

* 

There is a message pair, SYSTEM:MODULE_CALL/SYSTEM:MODULE_CALL_END that har sent out with level DEBUG, which show start and end of test case. There is, for some reason, a limitation so that they are only outputted if only one test case is tested.

~~They should always be outputted (with level debug).~~

No SYSTEM:MODULE_CALL/SYSTEM:MODULE_CALL_END are outputted:
```sh
zonemaster-cli ip.se --no-ipv6 --test DNSSEC --show_module --level debug --raw
zonemaster-cli ip.se --no-ipv6  --show_module --level debug --raw
```

SYSTEM:MODULE_CALL/SYSTEM:MODULE_CALL_END **are** outputted:
```sh
zonemaster-cli ip.se --no-ipv6 --test DNSSEC/dnssec13 --show_module --level debug --raw > dnssec13-raw
grep SYSTEM:MODULE_CALL dnssec13-raw 
   0.01 DEBUG     SYSTEM       SYSTEM:MODULE_CALL method=dnssec13; module=DNSSEC; version=v1.1.16
   3.80 DEBUG     SYSTEM       SYSTEM:MODULE_CALL_END method=dnssec13; module=DNSSEC
```

This issue relates to or duplicates #665

vlevigneron:
In PR zonemaster/zonemaster-engine#764 discussion, it has been decided to remove theses MODULE_CALL messages.

matsduf:
Resolved by #767

PR #764, opened 2020-07-23
Add START/END messages in case of debug of test cases 

vlevigneron:
Fixes zonemaster/zonemaster-engine#665

Implementation not yet finished but can you confirm @matsduf that it is what you are waiting for (Only 3 modules for the moment Basic/Address/Connectivity). If it's OK, I'll finish with the other modules.

### DIFF
--- a/lib/Zonemaster/Engine/Test/Address.pm
+++ b/lib/Zonemaster/Engine/Test/Address.pm
@@ -1,11 +1,11 @@
 package Zonemaster::Engine::Test::Address;
 
-use version; our $VERSION = version->declare("v1.0.5");
+use 5.014002;
 
 use strict;
 use warnings;
 
-use 5.014002;
+use version; our $VERSION = version->declare("v1.0.6");
 
 use Zonemaster::Engine;
 
@@ -13,6 +13,7 @@ use Carp;
 use List::MoreUtils qw[none any];
 use Locale::TextDomain qw[Zonemaster-Engine];
 use Readonly;
+use Zonemaster::Engine::Recursor;
 use Zonemaster::Engine::Constants qw[:addresses :ip];
 use Zonemaster::Engine::TestMethods;
 use Zonemaster::Engine::Util;
@@ -47,6 +48,8 @@ sub metadata {
             qw(
               NAMESERVER_IP_PRIVATE_NETWORK
               NO_IP_PRIVATE_NETWORK
+              TEST_CASE_END
+              TEST_CASE_START
               )
         ],
         address02 => [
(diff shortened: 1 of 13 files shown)


vlevigneron:
@matsduf Yes I looked at zonemaster/zonemaster-engine#732 but the fact is that the messages MODULE_CALL are written outside of test case codes that's why they are there only when we run a unique test case. It can not be used if we want a START/END message for all test cases.

If it's OK for you, I'll add the messages for all test cases.

matsduf:
> @matsduf Yes I looked at #732 but the fact is that the messages MODULE_CALL are written outside of test case codes that's why they are there only when we run a unique test case. It can not be used if we want a START/END message for all test cases.

I understand, and then I find it more or less worthless. I suggest that the code for the message MODULE_CALL is removed unless you see a reason to keep it (does not have to be done now). I wanted to make sure that you had not missed that and #732.

> If it's OK for you, I'll add the messages for all test cases.

Yes, it looks good. It will make it much easier to troubleshooting.

vlevigneron:
@matsduf Finished. I'll remove MODULE_CALL in another PR related to zonemaster/zonemaster-engine#732

vlevigneron:
@matsduf if you had to choose, what do you prefer, a full qualified method name (this is how it is implemented now) or the short version for testcase ?

>    0.01 DEBUG     SYSTEM       SYSTEM:MODULE_VERSION module=Zonemaster::Engine::Test::DNSSEC; version=v1.1.21
>   0.01 DEBUG     DNSSEC       DNSSEC:TEST_CASE_START testcase=Zonemaster::Engine::Test::DNSSEC::dnssec13

vs

>    0.01 DEBUG     SYSTEM       SYSTEM:MODULE_VERSION module=Zonemaster::Engine::Test::DNSSEC; version=v1.1.21
>   0.01 DEBUG     DNSSEC       DNSSEC:TEST_CASE_START testcase=dnssec13

matsduf:
> @matsduf if you had to choose, what do you prefer, a full qualified method name (this is how it is implemented now) or the short version for testcase ?

The short version.

> > 0.01 DEBUG     SYSTEM       SYSTEM:MODULE_VERSION module=Zonemaster::Engine::Test::DNSSEC; version=v1.1.21

Here I guess that the long, explicit format makes sense.

> > 0.01 DEBUG     DNSSEC       DNSSEC:TEST_CASE_START testcase=dnssec13

Here it is better with short name. It will be easier to read. It is possible to derive the full name from the testcase.

matsduf:
The changes look good so far. When completed, I will test.

Issue #769, opened 2020-08-06
Reduce the number of duplicate messages in dnssec14 test method


repperille:
When testing a zone for DNSSEC I get a lot of equal messages. In my case 16 messages in two variants

norid.no has got four nameservers (ns.uninett.no. nac.no. server.nordu.net. nn.uninett.no.) using two protocols (IPv4 and IPv6). Each query returns two DNSKEYs.

Example:
`zonemaster-cli --test DNSSEC norid.no`

Result:
Seconds Level     Message
======= ========= =======
   6.23 WARNING   DNSKEY with tag 53884 and using algorithm 8 (RSA/SHA-256) has a size (1280) smaller than the recommended one (2048).
   6.23 WARNING   DNSKEY with tag 44233 and using algorithm 8 (RSA/SHA-256) has a size (1280) smaller than the recommended one (2048).
   6.23 WARNING   DNSKEY with tag 44233 and using algorithm 8 (RSA/SHA-256) has a size (1280) smaller than the recommended one (2048).
   6.23 WARNING   DNSKEY with tag 53884 and using algorithm 8 (RSA/SHA-256) has a size (1280) smaller than the recommended one (2048).
   6.24 WARNING   DNSKEY with tag 53884 and using algorithm 8 (RSA/SHA-256) has a size (1280) smaller than the recommended one (2048).
   6.24 WARNING   DNSKEY with tag 44233 and using algorithm 8 (RSA/SHA-256) has a size (1280) smaller than the recommended one (2048).
   6.24 WARNING   DNSKEY with tag 44233 and using algorithm 8 (RSA/SHA-256) has a size (1280) smaller than the recommended one (2048).
   6.24 WARNING   DNSKEY with tag 53884 and using algorithm 8 (RSA/SHA-256) has a size (1280) smaller than the recommended one (2048).
   6.24 WARNING   DNSKEY with tag 53884 and using algorithm 8 (RSA/SHA-256) has a size (1280) smaller than the recommended one (2048).
   6.24 WARNING   DNSKEY with tag 44233 and using algorithm 8 (RSA/SHA-256) has a size (1280) smaller than the recommended one (2048).
   6.24 WARNING   DNSKEY with tag 53884 and using algorithm 8 (RSA/SHA-256) has a size (1280) smaller than the recommended one (2048).
   6.24 WARNING   DNSKEY with tag 44233 and using algorithm 8 (RSA/SHA-256) has a size (1280) smaller than the recommended one (2048).
[...]

PR #770, opened 2020-08-06
Reduce duplicate messages about key size

repperille:
When the same key from all name servers are tested a lot of duplicate messages are generated.

This fix will reduce the number of duplicate messages, i.e messages having the same keytag, keysize and algorithm.

Normally the same key is propagated to all authoritative name servers so testing each key from each name server will give the same result.

This is a fix for issue #769


repperille:
I replaced the tabs with spaces...

vlevigneron (approved):
I approve this message

PR #772, opened 2020-08-07
Updated instructions for translators

matsduf:
This PR does two things:
* Updates the instructions for translators
* Creates a separate document for translators

The POD format is fine for code documents when you want to create "man" pages but for presentation on Github the markdown format is much better. A separate document in markdown has therefore been created for the translators. That was the largest part of the original document.

Updates:
* More details for the steps.
* Norwegian has been added.
* Information on how to create a new language (a new PO file) has been added.

This PR expands the solution added by PR #701, it partly resolves #594, it gives a better resolution to #591 and it resloves #756 

The new "docs/Translation.pod" will link to "docs/Translation-translators.md" in the *develop* branch due to lack of relative links in POD files, which means that the link will not work until this PR has been merged.

Issue #713, opened 2020-04-17
Standardized and defined arguments in msgid


matsduf:
The test cases produces `msgid` with arguments, e.g. `{nss}`. It is not always obvious what an argument will be replaced by. IP address? Name? Name+ip address? Or list of elements?

When translating the `msgid` to `msgstr` it is important to know what it is to know how the argument could be handled.

There is a document, "[Types of information in Zonemaster log objects]", that seems to try to create standardized and defined arguments. Or at least started to do that.

We should revive that work and create good argument names to be used both for implementation and translation. The first step (the milestone) is to do that for the most simple test module to get a proof of concept.

[Types of information in Zonemaster log objects]: https://github.com/zonemaster/zonemaster-engine/blob/develop/docs/logentry_args.md

mattias-p:
See also #60.

matsduf:
Partially resolved by #790, #792, #799,  #853 and #854. The solutions are not yet complete.

Issue #60, opened 2015-03-31
Coherent naming of args in the log


pawal:
Sometimes nsset is called ns. Sometimes ns is called name. Sometimes an "ns" is only a name of a name server, and sometimes it is attached to an IP-adress. Sometimes it is split into an "ns" and an "address".

Sometimes the args is also empty, when there there is likely to actually contain data.

We really need to have coherent naming of the args in the log, with specific types attached to the name.

This is assigned to the whole team, and should be a team effort. This should also be documented.

pawal:
I have prepared this document as a starting point:
https://github.com/dotse/zonemaster/blob/master/docs/benchmark/logconsistency.md

pawal:
@vlevigneron Is this something you can spend some time on?

cdybedahl:
I'm currently going through the code and writing a list of all the types of information logged, so we can start with making a list of names for all the different kinds. When we have that, we can start changing the code to use them.

vlevigneron:
Yes of course, once I'll be back from vacation.

2015-04-27 11:18 GMT+02:00 Patrik Wallström <notifications@github.com>:

cdybedahl:
https://github.com/dotse/zonemaster-engine/blob/master/doc/logentry_args.md has a table where we can fill in the names to be used for the various kinds of data used. There are 71 different kinds, and that does not include and SYSTEM messages. Do we need to cover those as well?

pawal:
I don't think there is a need to look at the SYSTEM messages at this moment. At least I have never used them for anything useful - yet.

sandoche2k:
@pawal - can we move this to 2015.4?

pawal:
I strongly think that we should have a higher priority on this than that...

sandoche2k:
Ok. I think we will discuss with Vincent to decide on the release timings.

pawal:
If it is ok, I will start fixing a number of those things soon.

I will create a new issue for each "name".

sandoche2k:
Hi Patrik,

  In case if you do not need Vincent, i think you can start looking at it.

Sandoche.

sandoche2k:
@pawal move to next release

sandoche2k:
@sandoche2k - Put https://github.com/dotse/zonemaster/blob/master/docs/benchmark/logconsistency.md into ZM/docs
@pawal - Keep the issue open

sandoche2k:
@pawal - Can we move it to 2015.4?

[...]

PR #776, opened 2020-08-28
Validate use of arguments in translated messages

mattias-p:
This adds a util (check-msg-args) that verifies that the args used in the msgstr are the same ones that are used in the msgid for perl-brace-format messages.
The utility is hooked up to the `make check-msg-args` target in `share/Makefile`. ~That target is in turn hooked up in the unit tests.~ This check should not be run as a unit test because it would be impossible to update msgids without also updating the translations for all languages.

In order for the unit test not to immediately fail on these new checks, I had to fix the PO files. In doing so I found a couple of msgids where arguments were available but unused, and one where the word "Nameserver" hadn't been added before {ns}/{address}, so I fixed those as well. In the Danish translation I found a msgstr that didn't correspond to its msgid - instead of trying to fix that I just removed it. In the French translation I found a couple of messages that used arguments that weren't available. In those cases I just removed the arguments hoping that I didn't butcher the grammar or the meaning of the sentences. Please look extra closely at those when reviewing.

The translation documentation should be updated with instructions to use this utility, but I don't want to touch it at the moment because of #772.

~A new test-time dependency is introduced: Locale::PO.~ The translation documentation should be amended with installation of Local::PO, which is a new translation-time dependency. Locale::PO is packaged for all supported platforms except CentOS.

These checks will be very helpful when resolving #60 and #713.

### DIFF
--- a/share/Makefile
+++ b/share/Makefile
@@ -35,5 +35,8 @@ $(POFILES): $(POTFILE)
 show-fuzzy:
 	@for f in $(POFILES) ; do msgattrib --only-fuzzy $$f ; done
 
+check-msg-args:
+	@for f in $(POFILES) ; do ../util/check-msg-args $$f ; done
+
 modules.txt: $(TESTMODULEFILES)
 	@echo $(TESTMODULEFILES) | xargs basename -s .pm -a | grep -vE '^Basic$$' | sort > modules.txt
--- a/util/check-msg-args
+++ b/util/check-msg-args
@@ -1,0 +1,110 @@
+#!/usr/bin/env perl
+use utf8;
+use strict;
+use warnings;
+use open qw(:std :utf8);
+use feature 'say';
+
+use Locale::PO;
+use Try::Tiny;
+use Getopt::Long qw( GetOptions );
+use Pod::Usage qw( pod2usage );
+
+sub extract_perl_braces {
+    my $string = shift;
+
+    # Strip text before the first and after the last brace
+    $string =~ s/^[^{]*{?|}?[^}]*$//g;
+
+    # Extract the fields
+    my @fields = split /}[^{]*{/, $string;
+
+    return @fields;
+}
+
+sub diff_perl_braces {
+    my $got_string      = shift // die 'undefined value for $got_string';
+    my $expected_string = shift // die 'undefined value for $expected_string';
+
+    my %got_fields      = map { $_ => 1 } extract_perl_braces( $got_string );
+    my %expected_fields = map { $_ => 1 } extract_perl_braces( $expected_string );
+
+    my %missing = %expected_fields;
+    delete @missing{ keys %got_fields };
+
+    my %extra = %got_fields;
+    delete @extra{ keys %expected_fields };
+
+    return [ sort keys %missing ], [ sort keys %extra ];
+}
+
+sub check_perl_braces {
+    my $po = shift;
+    my $filename = shift;
+
+    my $msgid  = $po->dequote( $po->msgid );
+    my $msgstr = $po->dequote( $po->msgstr );
+    my ( $missing, $extra ) = diff_perl_braces( $msgstr, $msgid );
+
+    if ( @{$missing} or @{$extra} ) {
+        say "";
+        say '# ' . $filename . ' line ' . $po->loaded_line_number;
+        print map { "#. $_\n" } split /\n/, $po->automatic // '';
+        say "msgid ",  $po->msgid;
+        say "msgstr ", $po->msgstr;
+        for my $field ( @{$missing} ) {
+            say "  Only in msgid: {$field}";
+        }
+        for my $field ( @{$extra} ) {
+            say "  Only in msgstr: {$field}";
+        }
+    }
+
+    return;
+}
+
+my $opt_help = 0;
+my $opt_man  = 0;
+GetOptions(
+    "help|h" => \$opt_help,
+    "man"    => \$opt_man,
+) or pod2usage( 2 );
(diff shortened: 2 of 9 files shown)


share/Makefile
```diff
@@ -35,5 +36,8 @@ $(POFILES): $(POTFILE)
 show-fuzzy:
 	@for f in $(POFILES) ; do msgattrib --only-fuzzy $$f ; done
 
+check-msg-args:
+	@for f in $(POFILES) ; do $(ROOTDIR)/util/check-msg-args $$f ; done
+
```
matsduf, line 41: This does not seem to work, at least not in FreeBSD:

```
[~/git/zonemaster-engine/share]$ make check-msg-args
/bin/sh: /../util/check-msg-args: not found
*** Error code 127

Stop.
make: stopped in /usr/home/matsd/git/zonemaster-engine/share
[~/git/zonemaster-engine/share]$ ls -l ../util/check-msg-args 
-rwxr-xr-x  1 matsd  matsd  2098 sep.  8 13:56 ../util/check-msg-args
```

util/check-msg-args
```diff
@@ -0,0 +1,77 @@
...
+            if ( $po->has_flag( 'perl-brace-format' ) ) {
+                check_perl_braces( $po, $filename );
+            }
+        }
+        catch {
+            say STDERR "error processing $filename at line " . $po->loaded_line_number . ": " . $_;
+            exit 1;
+        };
+    }
+}
```
matsduf, line 77: There is no documentation what so ever in this script. Even though this script is meant to be run from Makefile there should be a paragraph explaining the purpose of it and that it is hooked from share/Makefile. Also, there could be situations when this is run outside Makefile, so at least explain what parameters it expects would be good.

matsduf (changes requested):

matsduf:
> This check should not be run as a unit test because it would be impossible to update msgids without also updating the translations for all languages.

Is that really true? If the message IDs are updated in the Perl modules it will not spread to the PO files until translation work is done.

>  The translation documentation should be amended with installation of Local::PO, which is a new translation-time dependency. Locale::PO is packaged for all supported platforms except CentOS.

If this tool is only run during translation work, then there is no need to install it for normal installation.

docs/Installation.md
```diff
@@ -51,7 +51,7 @@ This instruction covers the following operating systems:
 5) Install packages from CPAN:
 
    ```sh
-   sudo cpanm Email::Valid Locale::Msgfmt Module::Install Module::Install::XSUtil Test::More
+   sudo cpanm Email::Valid Locale::Msgfmt Locale::PO Module::Install Module::Install::XSUtil Test::More
```
matsduf, line 54: Is the addition needed if the tool is only used during translation work? Even if it is run by Travis it should not be run during normal installation because the PO files are not available at that time (from CPAN files).

matsduf (changes requested):

share/Makefile
mattias-p: Good catch. I'll take another look.

util/check-msg-args
mattias-p: I'll add some basic usage documentation.

mattias-p:
Github is being funny. I can't reply to the top-level review comment so I'm quoting it here instead.

>>    This check should not be run as a unit test because it would be impossible to update msgids without also updating the translations for all languages.

> Is that really true? If the message IDs are updated in the Perl modules it will not spread to the PO files until translation work is done. It is only meaningful to run the script when the PO files are available, and they are not when installation is done from CPAN..

Maybe you're right. I'll hook up the check as a unit test again.

>>    The translation documentation should be amended with installation of Local::PO, which is a new translation-time dependency. Locale::PO is packaged for all supported platforms except CentOS.

> If this tool is only run during translation work, then there is no need to install it for normal installation.

That's correct. There's no need at installation-time to install tools and dependencies that are only used at translation-time. But since I'm re-including this tool in the unit tests, and since we're running our tests as part of the normal installation, these dependencies will need to be installed at installation-time anyway.

docs/Installation.md
mattias-p: You make a good point that we don't need Locale::PO at installation-time if check-msg-args is only executed at translation-time. At least in theory.

However if we execute check-msg-args as part of a unit test, we should have a `test_requires 'Locale::PO'` entry in Makefile.PL. And then `cpanm Zonemaster::Engine` would install it even if check-msg-args is never invoked by the unit tests because there are no PO files.

I guess we could add something to the Travis configuration to explicitly run the check-msg-args checks, or we could add `--notest` to the `cpanm Zonemaster::Engine` command, but I feel both of those are out of scope for this PR.

I feel the right thing for this PR is to remove Locale::PO from the `cpanm`, `apt` and `pkg` commands.

share/Makefile
mattias-p: Fixed.

util/check-msg-args
mattias-p: Fixed.

docs/Installation.md
mattias-p: Fixed.

mattias-p:
I rebased to fix a merge conflict. I also took the opportunity to squash some of the commits.

@matsduf, @sandoche2k, @vlevigneron Please (re)review..

sandoche2k (approved):

share/Makefile
matsduf: Now it works in FreeBSD

matsduf (approved):

PR #777, opened 2020-09-03
Fix NO_SOA_IN_RESPONSE message

andreaso:
Turns

    Response from nameserver {ns}/{address} on SOA queries does not contain SOA record.

into

    Response from nameserver ns.example.com/198.51.100.53 on SOA queries does not contain SOA record.


matsduf (approved):

matsduf:
@vlevigneron, please review.

vlevigneron (approved):
Fine for me.

PR #778, opened 2020-09-10
Add type constraint to Z::E::Nameserver::Cache->data

### LINKED ISSUE #324: Crash in Zonemaster::Engine::Nameserver
Zonemaster::Engine::Nameserver is tripped up by the Backend test suite in dotse/zonemaster-backend#299.

```
$ make test TEST_FILES='t/test01.t'
...
t/test01.t .. 1/? Can't call method "string" on unblessed reference at /home/mattiasp/local/lib/perl5/Zonemaster/Engine/Nameserver.pm line 190.
...
```

Update:

The problem is that the internal state of Zonemaster::Engine can be poisoned, causing Zonemaster::Engine->add_fake_delegation() to crash when called with valid input.

The poisoning is caused in this case by Zonemaster::Engine::Nameserver->preload_cache() at the interface level, and by Zonemaster::Engine::Nameserver::Cache->new() at a deep level. These procedures are called with invalid input, and instead of catching and reporting the error, they poison the internal state.

The invalid input should be caught before the internal state is poisoned so that error symptoms surface close to their source.

Here's a minimized test case. Zonemaster::Engine::Nameserver::Cache->new() returns normally even though it's arguments are invalid, and Zonemaster::Engine->add_fake_delegation() crashes even though it's arguments are valid.

```perl
use Zonemaster::Engine;

eval {
    print "In one corner of the code base:\n";
    print "Buggy code injecting junk into the cache.\n";
    my $cache = Zonemaster::Engine::Nameserver::Cache->new(
        {
            data => { "nv7L7BL/+Xv5gj+KmxkC/w" => "This is a string instead of a Zonemaster::Engine::Packet" },
            address => Zonemaster::Engine::Net::IP->new('2001:dc3::35')
        }
    );
    print "Successfully injected junk into cache.\n";
};
if ( $@ ) {
    print "Failure occurred close to the bug in the code. This is easy to debug.";
    print "The error message:\n";
    print $@;
}

print "In another corner of the code base:\n";
print "Perform important but unrelated calculations.\n";
my $count;
for ( 1..1000000 ) { $count++ }

eval {
    print "In a third corner of the code base:\n";
    print "Correct code that relies on the cache being consistent.\n";
    Zonemaster::Engine->add_fake_delegation(
        'example.com' => {
            'ns1.example.com' => ['192.0.2.1'],
        }
    );
    print "No problem. Happy times.\n";
};
if ( $@ ) {
    print "Failure occurred far away from the bug in the code. This is very difficult to debug.\n";
    print "The error message:\n";
    print $@;
}
```

Update 2020-09-10:

#498 updated the cache implementation, breaking the test code given here. So I've updated the test code to demonstrate that the original issue is still present in the code.

mattias-p:
Fixes #324.

### DIFF
--- a/lib/Zonemaster/Engine/Nameserver/Cache.pm
+++ b/lib/Zonemaster/Engine/Nameserver/Cache.pm
@@ -10,7 +10,7 @@ use Zonemaster::Engine;
 
 our %object_cache;
 
-has 'data' => ( is => 'ro', isa => 'HashRef', default => sub { {} } );
+has 'data' => ( is => 'ro', isa => 'HashRef[Maybe[Zonemaster::Engine::Packet]]', default => sub { {} } );
 has 'address' => ( is => 'ro', isa => 'Zonemaster::Engine::Net::IP', required => 1 );
 
 around 'new' => sub {


matsduf (approved):

mattias-p:
@sandoche2k, @vlevigneron Please review.

sandoche2k (approved):

vlevigneron (approved):

PR #782, opened 2020-09-16
"Fix NO_SOA_IN_RESPONSE message" to correct branch 

matsduf:
Turns

    Response from nameserver {ns}/{address} on SOA queries does not contain SOA record.

into

    Response from nameserver ns.example.com/198.51.100.53 on SOA queries does not contain SOA record.


matsduf:
This #777 but against correct branch.

mattias-p (approved):

PR #784, opened 2020-09-21
Silence PACKET_BIG for the time being

mattias-p:
#731 suggests that this message should be demoted to DEBUG until it is properly documented and implemented. This PR contains the demotion part.


matsduf:
Even as DEBUG message, it is an incorrect message. From #731: "If this test is kept as-is, it should be down graded to DEBUG **_and the message should be changed so that nothing is indicated that the size over some limit_**, just indicate that it is a large message."

Can you also change the message too, e.g.:

"Big packet size ({size}) (try with \"{command}\")."

mattias-p:
Sure. I updated the message per your suggestion.

matsduf (approved):

vlevigneron (approved):

PR #787, opened 2020-09-22
Add missing dots at the end of several msgid

matsduf:
The msgid should be terminated by a dot, and usually it is. This PR fixes some cases where the dot was missing.

### DIFF
--- a/lib/Zonemaster/Engine/Test/Connectivity.pm
+++ b/lib/Zonemaster/Engine/Test/Connectivity.pm
@@ -164,15 +164,15 @@ Readonly my %TAG_DESCRIPTIONS => (
     },
     ASN_INFOS_RAW => sub {
         __x    # CONNECTIVITY:ASN_INFOS_RAW
-          '[ASN:RAW] {address};{data}', @_;
+          '[ASN:RAW] {address};{data}.', @_;
     },
     ASN_INFOS_ANNOUNCE_BY => sub {
         __x    # CONNECTIVITY:ASN_INFOS_ANNOUNCE_BY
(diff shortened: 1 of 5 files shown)


lib/Zonemaster/Engine/Test/Connectivity.pm
```diff
@@ -164,15 +164,15 @@ Readonly my %TAG_DESCRIPTIONS => (
     },
     ASN_INFOS_RAW => sub {
         __x    # CONNECTIVITY:ASN_INFOS_RAW
-          '[ASN:RAW] {address};{data}', @_;
+          '[ASN:RAW] {address};{data}.', @_;
```
mattias-p, line 167: Does it really make sense to add trailing dots to these three messages? They aren't really English sentences that need to be terminated with a full stop. Maybe we should give them English texts?

mattias-p (commented):
Great initiative! I just have one question regarding the non-English messages.

lib/Zonemaster/Engine/Test/Connectivity.pm
matsduf: I think it makes sense that we consistently end all msgid's with a dot. If so, we do not have to construct some rule on what kind of phrase that should have dot, and what should not. The translator does not have to be unsure. The msgid might just be a none phrase where as the msgstr is set to be a sentence with a finite verb.

Secondly, the example you gave is not the only non-language like string. Other examples are 'TEST_CASE_END {testcase}.' and 'TEST_CASE_START {testcase}.' from Zone.pm, and they are terminated by a dot.

Keep it simple -- always terminate with a dot.

mattias-p (approved):

PR #790, opened 2020-09-24
Update some message args for consistency

mattias-p:
This PR concerns itself with a very small number of argument names, but it's by far the most common ones so it's still a very big diff. I believe it is a good approach to establish a good pattern for the most common argument names first, and look into less common argument names afterwards.

The most common argument names in the current develop branch are by far "ns" and "address". I looked at the usage of these and came up with the following convention:
* {nsname} - a name server name
* {nsip} - a name server IP address
* {ns} - a name server identified by both name and IP, separated with a slash
* {nsnames} - a semicolon separated list of {nsname}
* {nsips} - a semicolon separated list of {nsip}
* {nss} - a semicolon separated list of {ns}

A reasonable alternative I could think of is to use "list" for pluralization instead of just using "s". I.e. the last three would become {nsnamelist}, {nsiplist} and {nslist}.
Another idea is to use snake_case or kebab-case for improved readability. E.g. {ns_name}, {ns_ip_list} or {ns-name}, {ns-ip-list}. IMO kebab-case is better looking, but snake_case names are accepted as Perl barewords which plays nicer in the source code.

Note: There are no examples of {nsips} or {nss} in this PR.

Note: There are examples of messages whose arguments get more confusing as a whole because of this rename. E.g. because {nsname} is combined with {name}. That'll be resolved in a future PR where we come up with a better name for the {name} argument, etc.

Note: Even more arguments should probably bare these names, but they weren't included here because they're currently named something other than the names listed here. We'll rename those in future PRs.

I'll make a separate PR for the SYSTEM messages. They aren't defined and used in the same file, so I figured reviewing them would be more difficult if they were included in this PR.

If you think this PR is too large to review properly, I can split off some modules into one or more separate PRs. The changes to the Nameserver and DNSSEC test modules are particularly voluminous.

This is a step towards resolving #713.

### DIFF
--- a/docs/logentry_args.md
+++ b/docs/logentry_args.md
@@ -2,14 +2,14 @@
 
 ## Aggregate list
 
-| Argument | Type of value         | Description and formatting                                    |
-|----------|-----------------------|---------------------------------------------------------------|
-| nsname   | Domain name           | The domain name of a name server.                             |
-| nsip     | IP address            | The IP address of a name server.                              |
-| ns       | Domain name and IP address pair | The name and IP address of a name server, separated by "/".|
-| nsnames  | List of domain names  | The names of a list of nameservers, separated by ";".         |
-| nsips    | List of IP addresses  | The addresses of a list of nameservers, separated by ";".     |
-| nss      | List of domain name and IP address pairs | The names and addresses of a list of nameservers, the name and address of a nameserver are separated by "/", nameservers are separated by ";".|
+| Argument    | Type of value        | Description and formatting                                  |
+|-------------|----------------------|-------------------------------------------------------------|
+| nsname      | Domain name          | The domain name of a name server.                           |
+| ns_ip       | IP address           | The IP address of a name server.                            |
+| ns          | Domain name and IP address pair | The name and IP address of a name server, separated by "/". |
+| nsname_list | List of domain names | A list of name servers, as specified by "nsname", separated by ";". |
+| ns_ip_list  | List of IP addresses | A list of name servers, as specified by "nsip", separated by ";". |
+| ns_list     | List of domain name and IP address pairs | A list of name servers, as specified by "ns", separated by ";". |
 || AS number| An Autonomous Space number for an IP address.|
 || Address record type (A or AAAA)| Used to tell the difference between IPv4 and IPv6.|
 || Count of different SOA RNAMEs.| Total number of different SOA RNAME fields seen.|
--- a/docs/logentry_args.md
+++ b/docs/logentry_args.md
@@ -2,8 +2,14 @@
 
 ## Aggregate list
 
-|Argument name|Type of value | Description |
-|-------------|----------|----------------------------------------------|
+| Argument | Type of value         | Description and formatting                                    |
+|----------|-----------------------|---------------------------------------------------------------|
+| nsname   | Domain name           | The domain name of a name server.                             |
+| nsip     | IP address            | The IP address of a name server.                              |
+| ns       | Domain name and IP address pair | The name and IP address of a name server, separated by "/".|
+| nsnames  | List of domain names  | The names of a list of nameservers, separated by ";".         |
+| nsips    | List of IP addresses  | The addresses of a list of nameservers, separated by ";".     |
+| nss      | List of domain name and IP address pairs | The names and addresses of a list of nameservers, the name and address of a nameserver are separated by "/", nameservers are separated by ";".|
 || AS number| An Autonomous Space number for an IP address.|
 || Address record type (A or AAAA)| Used to tell the difference between IPv4 and IPv6.|
 || Count of different SOA RNAMEs.| Total number of different SOA RNAME fields seen.|
@@ -19,12 +25,12 @@
 || DNSSEC delegation verification failure reason| A somewhat human-readable reason why the delegation step between the tested zone and its parent is not secure.|
 || DS digest type| The digest type used in a DS record.|
 || DS/DNSKEY/RRSIG keytag| A keytag for a DS, DNSKEY or RRSIG record.|
-| dname | Domain name| A domain name.|
-| dlabel | Domain name label| A single label from a domain name.|
-| dlength | Domain name label length| The length of a domain name label.|
+| dname (?) | Domain name| A domain name.|
+| dlabel (?) | Domain name label| A single label from a domain name.|
+| dlength (?) | Domain name label length| The length of a domain name label.|
 || Duration in seconds| An integer number of seconds.|
-| fqdn | FQDN| A fully qualified domain name (with terminating dot).|
-| fqdnlength | FQDN length| The length of an FQDN.|
+| fqdn (?) | FQDN| A fully qualified domain name (with terminating dot).|
+| fqdnlength (?) | FQDN length| The length of an FQDN.|
 || IP address| An IPv4 or IPv6 address.|
 || IP address or nothing| An IPv4 or IPv6 address, or no value.|
 || IP range| An IP range.|
@@ -40,23 +46,18 @@
 || List of SOA RNAMEs| A list of RNAME values from SOA records.|
 || List of SOA serial numbers| A list of serial number values from SOA records.|
 || List of domain names| A list of domain names.|
-|| List of nameserver name/IP pairs.| A list of nameservers, specified as name/address pairs.|
-| nsnlist | List of nameserver names| A list of nameserver names.|
 || NS names from child| A list of nameserver names taken from a zone's child servers.|
 || NS names from parent| A list of nameserver names taken from a zone's parent servers.|
 || NSEC3 iteration count| An iteration count from an NSEC3PARAM record.|
-| nsip | Nameserver IP| The IP address of a name server.|
-| nsname | Nameserver name| The domain name of a name server.|
-| ns | Nameserver name/IP pair| The name and IP address of a name server, separated by a "/" character.|
 || Number of DNSKEY RRs in packet| The number of DNSKEY records found in a packet.|
 || Number of RRSIG RRs in packet| The number of RRSIG records found in a packet.|
 || Number of SOA RRs in packet| The number of SOA records found in a packet.|
 || PTR query name| The domain name generated from an IP address for a reverse name lookup.|
-| pname | Parent zone name| The name of a tested zone's parent zone.|
+| pname (?) | Parent zone name| The name of a tested zone's parent zone.|
 || Protocol (UDP or TCP)| The protocol used for a query.|
-| rcode | RCODE| An RCODE from a DNS packet.|
+| rcode (?) | RCODE| An RCODE from a DNS packet.|
 || RFC reference| A reference to an RFC.|
-| rrtype | RR type| The type of RR the message pertains to.|
+| rrtype (?) | RR type| The type of RR the message pertains to.|
 || RRSIG Expiration date| The time when a signature expires.|
 || RRSIG validation error message| The human-readable reason why the cryptographic validation of a signature failed.|
 || SOA MNAME| The MNAME value from a SOA record.|
@@ -74,7 +75,11 @@
 || Smallest SOA serial number seen| The smallest value seen in a SOA serial field in the tested zone.|
 || TLD| The name of a top-level domain.|
 || `time_t` value when RRSIG validation was attempted| The time when an RRSIG validation was attempted, in Unix `time_t` format.|
-| zname | Zone name| The domain name of the zone being tested.|
+| zname (?) | Zone name| The domain name of the zone being tested.|
+
+Message names maked with a question mark should not be considered stable.
+
+
 ## List by test module
 
 ### Basic
--- a/docs/logentry_args.md
+++ b/docs/logentry_args.md
@@ -8,7 +8,7 @@
 | ns_ip       | IP address           | The IP address of a name server.                            |
 | ns          | Domain name and IP address pair | The name and IP address of a name server, separated by "/". |
 | nsname_list | List of domain names | A list of name servers, as specified by "nsname", separated by ";". |
-| ns_ip_list  | List of IP addresses | A list of name servers, as specified by "nsip", separated by ";". |
+| ns_ip_list  | List of IP addresses | A list of name servers, as specified by "ns_ip", separated by ";". |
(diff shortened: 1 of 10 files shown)


matsduf:
@mattias-p, have you looked at [Types of information in Zonemaster log objects]? It relates to this PR. I think it should either be updated or replaced by a new document as part of this PR. We need some documentation even it is just a commented list.

Should all Perl modules with msgid's link to the documentation?

[Types of information in Zonemaster log objects]: https://github.com/zonemaster/zonemaster-engine/blob/master/docs/logentry_args.md

matsduf:
"{nsname}" and "{nsnames}" work well even though an explicit {nsname_list} is clearer. "{nsname}" is clear enough so it does not need any snake casing.

"{nsip}" works, but "{nsips}" is less clear. I think both should be snake cased and the second made explicit, i.e. "{ns_ip}" and "{ns_ip_list}", respectively.

"{ns}" is a too short name for a complex structure. At least it requires that the documentation is good and accessible. "{ns_and_ip}" or "{name_and_ip}" would be clearer.

"{nss}" does not work well at all. If "{ns}" is kept, then "{ns_list}" is the preferred option. If "{ns_and_ip}" is taken, then it should be "{ns_and_ip_list}".

Could it be the case that the IP address is not known when "{ns}" ({ns_and_ip}) is used? If so, will the IP address be an empty string? 


In all list cases it is stated that is is "a semicolon separated list". That use of semicolon does not work well in natural language, which these lists are inserted in. Why not comma separated list instead? Do we ever have a case were the list element can have comma in the data?

matsduf (changes requested):
This is a good work, but it needs some discussions and adjustments. We definitely need an updated documentation.

No, the PR is not too large to review.

docs/logentry_args.md
```diff
@@ -2,8 +2,14 @@
...
 ## Aggregate list
 
-|Argument name|Type of value | Description |
-|-------------|----------|----------------------------------------------|
+| Argument | Type of value         | Description and formatting                                    |
+|----------|-----------------------|---------------------------------------------------------------|
+| nsname   | Domain name           | The domain name of a name server.                             |
+| nsip     | IP address            | The IP address of a name server.                              |
+| ns       | Domain name and IP address pair | The name and IP address of a name server, separated by "/".|
+| nsnames  | List of domain names  | The names of a list of nameservers, separated by ";".         |
```
matsduf, line 10: I propose that
```
The names of a list of nameservers, separated by ";". 
```
is changed to
```
A list of name servers, as specified by "nsname", separated by ";".
```

docs/logentry_args.md
```diff
@@ -2,8 +2,14 @@
...
+| nsips    | List of IP addresses  | The addresses of a list of nameservers, separated by ";".     |
```
matsduf, line 11: I propose that
```
The addresses of a list of nameservers, separated by ";". 
```
is changed to
```
A list of name servers, as specified by "nsip", separated by ";".
```

docs/logentry_args.md
```diff
@@ -2,8 +2,14 @@
...
+| nss      | List of domain name and IP address pairs | The names and addresses of a list of nameservers, the name and address of a nameserver are separated by "/", nameservers are separated by ";".|
```
matsduf, line 12: I propose that
```
List of domain name and IP address pairs | The names and addresses of a list of nameservers, the name and address of a nameserver are separated by "/", nameservers are separated by ";".
```
is changed to
```
A list of name servers, as specified by "ns", separated by ";".
```

docs/logentry_args.md
```diff
@@ -19,12 +25,12 @@
 || DNSSEC delegation verification failure reason| A somewhat human-readable reason why the delegation step between the tested zone and its parent is not secure.|
 || DS digest type| The digest type used in a DS record.|
 || DS/DNSKEY/RRSIG keytag| A keytag for a DS, DNSKEY or RRSIG record.|
-| dname | Domain name| A domain name.|
-| dlabel | Domain name label| A single label from a domain name.|
-| dlength | Domain name label length| The length of a domain name label.|
+| dname (?) | Domain name| A domain name.|
```
matsduf, line 28: I really dislike "dname" since there is a record type DNAME.

matsduf (changes requested):

mattias-p:
> @mattias-p, have you looked at Types of information in Zonemaster log objects? It relates to this PR. I think it should either be updated or replaced by a new document as part of this PR. We need some documentation even it is just a commented list.

I included an update to the documentation that reflects the current state of this PR.

> Should all Perl modules with msgid's link to the documentation?

I don't think it would be very relevant there. To me it would make more sense to include it in a code review checklist.

> "{nsname}" and "{nsnames}" work well even though an explicit {nsname_list} is clearer. "{nsname}" is clear enough so it does not need any snake casing.

The casing should be consistent across all argument names. If we start introducing exceptions we'll just keep forgetting which ones those exceptions are.

I take it that you like the "-list" suffix better than the "-s" suffix. I tend to agree.

I take it that you prefer snake_case over kebab-case and flatcase.

> "{ns}" is a too short name for a complex structure. At least it requires that the documentation is good and accessible. "{ns_and_ip}" or "{name_and_ip}" would be clearer.

To me having a short name for a complex structure is no problem. Short argument names is a good thing. They shouldn't grab your attention away from the intention of the message and should generally be low key. Having a really short name for the most common argument is a great feature, IMO.

> Could it be the case that the IP address is not known when "{ns}" ({ns_and_ip}) is used? If so, will the IP address be an empty string?

Interesting question, but I'd say no. In the test specifications you always know whether you have the IP address or not.

> In all list cases it is stated that is is "a semicolon separated list". That use of semicolon does not work well in natural language, which these lists are inserted in. Why not comma separated list instead? Do we ever have a case were the list element can have comma in the data?

Well, that's how it's currently implemented. In this PR I'd like to focus on the few argument names I listed. The footprint of this change is large enough as it is. I'd very much prefer if we could deal with argument value formatting later and separately from #713.

docs/logentry_args.md
mattias-p: Yeah, we'll get to that.

matsduf:
> > Should all Perl modules with msgid's link to the documentation?
> 
> I don't think it would be very relevant there. To me it would make more sense to include it in a code review checklist.

That is a good suggestion, but do we have such a checklist? It is also important that the coder remembers that the parameters are defined somewhere, and that you cannot just invest one.


> > "{nsname}" and "{nsnames}" work well even though an explicit {nsname_list} is clearer. "{nsname}" is clear enough so it does not need any snake casing.
> 
> The casing should be consistent across all argument names. If we start introducing exceptions we'll just keep forgetting which ones those exceptions are.

I do not get that. The implementer will have to check the documentation for which parameter name to use anyway. I do not understand what you mean by "exceptions"?

> I take it that you prefer snake_case over kebab-case and flatcase.

What is "flatcase"? You gave a good argument for why snake case was better than kebab case and I accepted that.

> > "{ns}" is a too short name for a complex structure. At least it requires that the documentation is good and accessible. "{ns_and_ip}" or "{name_and_ip}" would be clearer.
> 
> To me having a short name for a complex structure is no problem. Short argument names is a good thing. They shouldn't grab your attention away from the intention of the message and should generally be low key. Having a really short name for the most common argument is a great feature, IMO.

There is no problem when there is a natural connection between name and the structure. Here "ns" feels familiar, but the data type is special. There is high risk that you think that you understand that "ns" means name server name. I think it would be better to have a name that says that this is not what you think unless you know what it is.

> > In all list cases it is stated that is is "a semicolon separated list". That use of semicolon does not work well in natural language, which these lists are inserted in. Why not comma separated list instead? Do we ever have a case were the list element can have comma in the data?
> 
> Well, that's how it's currently implemented. In this PR I'd like to focus on the few argument names I listed. The footprint of this change is large enough as it is. I'd very much prefer if we could deal with argument value formatting later and separately from #713.

I agree with you. That should be a separate question not handled by this PR.

sandoche2k:
@mattias-p I feel that better let @matsduf review this issue.

mattias-p:
>>> Should all Perl modules with msgid's link to the documentation?

>> I don't think it would be very relevant there. To me it would make more sense to include it in a code review checklist.

> That is a good suggestion, but do we have such a checklist? It is also important that the coder remembers that the parameters are defined somewhere, and that you cannot just [invent] one.

I created #791 regarding the checklist. If a coder is unaware of the naming convention, the reviewer will point it out.

> What is "flatcase"?

It's when you just remove the spaces between lower-case words. (Source: https://en.wikipedia.org/wiki/Naming_convention_(programming)#Examples_of_multiple-word_identifier_formats).

>>> "{ns}" is a too short name for a complex structure. At least it requires that the documentation is good and accessible. "{ns_and_ip}" or "{name_and_ip}" would be clearer.

>> To me having a short name for a complex structure is no problem. Short argument names is a good thing. They shouldn't grab your attention away from the intention of the message and should generally be low key. Having a really short name for the most common argument [name] is a great feature, IMO.

> There is no problem when there is a natural connection between name and the structure. Here "ns" feels familiar, but the data type is special. There is high risk that you think that you understand that "ns" means name server name. I think it would be better to have a name that says that this is not what you think unless you know what it is.

Could you elaborate on the probability and severity of this risk? From my understanding they're both on the low end of the scale.

docs/logentry_args.md
```diff
@@ -2,14 +2,14 @@
...
-|----------|-----------------------|---------------------------------------------------------------|
-| nsname   | Domain name           | The domain name of a name server.                             |
-| nsip     | IP address            | The IP address of a name server.                              |
-| ns       | Domain name and IP address pair | The name and IP address of a name server, separated by "/".|
-| nsnames  | List of domain names  | The names of a list of nameservers, separated by ";".         |
-| nsips    | List of IP addresses  | The addresses of a list of nameservers, separated by ";".     |
-| nss      | List of domain name and IP address pairs | The names and addresses of a list of nameservers, the name and address of a nameserver are separated by "/", nameservers are separated by ";".|
+| Argument    | Type of value        | Description and formatting                                  |
+|-------------|----------------------|-------------------------------------------------------------|
+| nsname      | Domain name          | The domain name of a name server.                           |
```
matsduf, line 7: It will be more consistent and clearer if "domain" is dropped, "the name of a name server". Compare with "ns" below.

docs/logentry_args.md
```diff
@@ -2,14 +2,14 @@
...
+| ns_ip       | IP address           | The IP address of a name server.                            |
+| ns          | Domain name and IP address pair | The name and IP address of a name server, separated by "/". |
+| nsname_list | List of domain names | A list of name servers, as specified by "nsname", separated by ";". |
+| ns_ip_list  | List of IP addresses | A list of name servers, as specified by "nsip", separated by ";". |
```
matsduf, line 11: "nsip" --> "ns_ip".

mattias-p:
@matsduf Please re-review.

@matsduf, @vlevigneron It would be great if you guys could pay some extra attention to the refactorings to make sure I didn't make any silly mistakes. It's a rather tedious thing to review so maybe you want to split that up between yourselves.

matsduf (changes requested):

lib/Zonemaster/Engine/Test/DNSSEC.pm
```diff
@@ -987,7 +985,7 @@ sub dnssec02 {
 
             my $ds_p = $ns->query( $zone->name, q{DS}, { usevc => 0, dnssec => 1 } );
             if ( not $ds_p ) {
-                push @results, info( NO_RESPONSE => $ns_args );
+                push @results, info( NO_RESPONSE => { ns => $ns->string } );
```
matsduf, line 988: Why is this code changed when the equivalent for e.g. IPV4_DISABLED on 6 lines up is not changed?

matsduf (commented):
It should be "nsname", shouldn't it?

```
    NAMESERVER_DISCOURAGED_DOUBLE_DASH => sub {
        __x    # SYNTAX:NAMESERVER_DISCOURAGED_DOUBLE_DASH
          'Nameserver ({name}) has a label ({label}) with a double hyphen (\'--\') '
          . 'in position 3 and 4 (with a prefix which is not \'xn--\').',
          @_;
    },
```

matsduf:
> @matsduf, @vlevigneron It would be great if you guys could pay some extra attention to the refactorings to make sure I didn't make any silly mistakes. It's a rather tedious thing to review so maybe you want to split that up between yourselves.

I have read through all the changes. I only found what I have commented. It would have helped if the refactoring was in a separate commit.

docs/logentry_args.md
mattias-p: We'll give this table a bigger makeover later on. I see no point in polishing it to some local maximum right now.

lib/Zonemaster/Engine/Test/DNSSEC.pm
mattias-p: When there's only one argument in the hash there's no point in using $ns_args.

mattias-p:
> It should be "nsname", shouldn't it?
> ```
>    NAMESERVER_DISCOURAGED_DOUBLE_DASH => sub {
>        __x    # SYNTAX:NAMESERVER_DISCOURAGED_DOUBLE_DASH
>          'Nameserver ({name}) has a label ({label}) with a double hyphen (\'--\') '
>          . 'in position 3 and 4 (with a prefix which is not \'xn--\').',
>          @_;
>    },
> ```

I suspect there are lots of these. We'll get back to it when we take care of the {name} arguments.

> I have read through all the changes. I only found what I have commented.

Thanks a lot for reviewing!

> It would have helped if the refactoring was in a separate commit.

Yeah, I agree your point on separating out the refactorings. They weren't even technically necessary, but they were a byproduct of a somewhat messy process of preparing the PR. I guess I could have spent some more time removing them to make the review easier.

docs/logentry_args.md
mattias-p: Fixed.

lib/Zonemaster/Engine/Test/DNSSEC.pm
matsduf: I would leave it as-is in this PR.

matsduf (approved):
More work to be done, but this is an improvement.

mattias-p:
> More work to be done

Yeah, I expect this will be ongoing for a while. I'm sure the following PRs can be made much smaller though.

PR #792, opened 2020-09-29
Update ns args in SYSTEM messages

mattias-p:
This is a follow-up to #790 with the corresponding updates to the SYSTEM messages.

It is also a step towards solving #713.

### DIFF
--- a/lib/Zonemaster/Engine/Translator.pm
+++ b/lib/Zonemaster/Engine/Translator.pm
@@ -77,11 +77,11 @@ Readonly my %TAG_DESCRIPTIONS => (
     },
     SKIP_IPV4_DISABLED => sub {
         __x    # SYSTEM:SKIP_IPV4_DISABLED
-          "IPv4 is disabled, not sending query to {ns}.", @_;
+          "IPv4 is disabled, not sending query to {ns_list}.", @_;
     },
     SKIP_IPV6_DISABLED => sub {
         __x    # SYSTEM:SKIP_IPV6_DISABLED
-          "IPv6 is disabled, not sending query to {ns}.", @_;
+          "IPv6 is disabled, not sending query to {ns_list}.", @_;
     },
     FAKE_DELEGATION => sub {
         __x    # SYSTEM:FAKE_DELEGATION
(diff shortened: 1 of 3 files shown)


lib/Zonemaster/Engine/Translator.pm
```diff
@@ -97,13 +97,13 @@ Readonly my %TAG_DESCRIPTIONS => (
     },
```
matsduf, line 97: ADDED_FAKE_DELEGATION and FAKE_DELEGATION_TO_SELF, a few lines up, should also be updated.

lib/Zonemaster/Engine/Translator.pm
```diff
@@ -77,11 +77,11 @@ Readonly my %TAG_DESCRIPTIONS => (
...
         __x    # SYSTEM:SKIP_IPV4_DISABLED
-          "IPv4 is disabled, not sending query to {ns}.", @_;
+          "IPv4 is disabled, not sending query to {ns_list}.", @_;
     },
     SKIP_IPV6_DISABLED => sub {
         __x    # SYSTEM:SKIP_IPV6_DISABLED
-          "IPv6 is disabled, not sending query to {ns}.", @_;
+          "IPv6 is disabled, not sending query to {ns_list}.", @_;
     },
     FAKE_DELEGATION => sub {
```
matsduf, line 86: Defined without arguments, but used with. (Just an observation, not for this PR.)

matsduf (changes requested):

lib/Zonemaster/Engine/Translator.pm
mattias-p: In what way should they be updated?

matsduf: I assumed that "ns" was a name server name, as it has usually been used for, not a name server name / IP address pair. Where ADDED_FAKE_DELEGATION and FAKE_DELEGATION_TO_SELF are used it says `ns => "$self"` and I am not sure what `$self` will return there.

mattias-p: It evaluates to a name / IP address string.

matsduf (approved):

PR #798, opened 2020-10-02
Initialize Z::E::Translator::_last_language

mattias-p:
Initialize Z::E::Translator::_last_language. If it's not initialized we get warnings.


matsduf (approved):
This is a small code change that looks unproblematic.

PR #804, opened 2020-10-14
Fix installation from dist tarball

mattias-p:
This fixes installation from dist tarball. The problem is that `*.po` isn't expanded to the empty string if there are no PO files. One thing leads to another and all of a sudden `xgettext` is required at installation time, which it's not supposed to be.

The PR also includes a small simplification of the installation instruction.

### DIFF
--- a/share/Makefile
+++ b/share/Makefile
@@ -2,7 +2,7 @@
 .SUFFIXES: .po .mo
 .PHONY: all check-msg-args dist extract-pot tidy-po show-fuzzy touch-po update-po
 
-POFILES != echo *.po
+POFILES != find . -type f -name '*.po'
 MOFILES := $(POFILES:%.po=%.mo)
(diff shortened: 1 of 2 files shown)


share/Makefile
```diff
@@ -2,7 +2,7 @@
 .SUFFIXES: .po .mo
 .PHONY: all check-msg-args dist extract-pot tidy-po show-fuzzy touch-po update-po
 
-POFILES != echo *.po
+POFILES != find . -type f -name '*.po'
```
matsduf, line 5: This will find files further down. "-depth0", maybe?

matsduf (approved):

share/Makefile
mattias-p: I doubt there will be any. But sure, I'll add that as a drive-by fix in a future PR.

PR #775, opened 2020-08-21
Implement singleton pattern idiomatically

mattias-p:
Apparently Moose doesn't like that you wrap the `new()` constructor unless you're a Moose extension. This is an update implement the singleton pattern using a Moose extension instead of doing it in an ad-hoc manner.

Fixes #766.

PR #805, opened 2020-10-16
Fix locale initialization

mattias-p:
It seems #775 broke the initialization of the locale attribute. MooseX::Singleton doesn't call the BUILD method. This PR fixes that. It also fixes a problem on Debian 9 where setlocale() returns undef when the locale environment settings are invalid.

This also includes a drive-by improvement to share/Makefile.

### DIFF
--- a/lib/Zonemaster/Engine/Translator.pm
+++ b/lib/Zonemaster/Engine/Translator.pm
@@ -11,7 +11,7 @@ use Zonemaster::Engine;
 use Carp;
 use Locale::Messages qw[textdomain];
 use Locale::TextDomain qw[Zonemaster-Engine];
-use POSIX qw[setlocale LC_ALL LC_MESSAGES];
+use POSIX qw[setlocale LC_MESSAGES];
 use Readonly;
 
 use Moose;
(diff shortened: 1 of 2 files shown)


lib/Zonemaster/Engine/Translator.pm
```diff
@@ -117,25 +117,43 @@ Readonly my %TAG_DESCRIPTIONS => (
...
+sub _init_locale {
+    my $locale = setlocale( LC_MESSAGES, "" );
+
+    if ( !defined $locale ) {
+        my $language = $ENV{LANGUAGE} // "";
+        for my $value ( split /:/, $language ) {
+            if ( $value ne "" && $value !~ /[.]/ ) {
+                $value .= ".UTF-8";
+            }
+            $locale = setlocale( LC_ALL, $value );
```
matsduf, line 146: "Updates the program's underlying LC_MESSAGES to the returned value" -- but on line 146 it looks like LC_ALL is set to the value. Do I misunderstand it?

mattias-p: No, that's a good point. I'll make an update.

matsduf (approved):

matsduf:
@vlevigneron, can you review this PR?

sandoche2k (approved):

vlevigneron (approved):

PR #807, opened 2020-10-19
Remove path from find output i share/Makefile

matsduf:
FreeBSD make does not consider "./sv.po" (as returned by `find`) to be the same thing as "sv.po" as written as target in `make sv.po`. This PR fixes that by making `find` removing the path "./" before printing the file names.


pnax (approved):
This looks fine

mattias-p (approved):

PR #815, opened 2020-10-21
Use gmake for FreeBSD

matsduf:
* Updated Makefile.PL so that gmake is used for share/Makefile
* Updated installation instruction to include devel/gmake
* Updated MANIFEST to include yet another file under "inc"

### DIFF
--- a/MANIFEST
+++ b/MANIFEST
@@ -9,6 +9,7 @@ docs/Translation-translators.md
 inc/Module/Install.pm
 inc/Module/Install/Base.pm
 inc/Module/Install/Can.pm
+inc/Module/Install/External.pm
 inc/Module/Install/Fetch.pm
 inc/Module/Install/Makefile.pm
 inc/Module/Install/Metadata.pm
--- a/Makefile.PL
+++ b/Makefile.PL
@@ -38,14 +38,25 @@ test_requires 'Test::Exception'   => 0;
 test_requires 'Test::Fatal'       => 0;
 test_requires 'Test::Pod'         => 1.22;
 
+if ($^O eq "freebsd") {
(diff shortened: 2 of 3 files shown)


matsduf:
Please test this on Linux too.

MANIFEST
```diff
@@ -9,6 +9,7 @@ docs/Translation-translators.md
 inc/Module/Install.pm
 inc/Module/Install/Base.pm
 inc/Module/Install/Can.pm
+inc/Module/Install/External.pm
```
pnax, line 12: This file is not present on Ubuntu : 

```
$ make distcheck
"/usr/bin/perl" "-Iinc" "-MExtUtils::Manifest=fullcheck" -e fullcheck
No such file: inc/Module/Install/External.pm
```

pnax (commented):
Where is the `inc/Module/Install/External.pm` from ? I doesn't seem to be present when building on Ubuntu.

MANIFEST
matsduf: I suspected that. It appeared when building on FreeBSD after adding
```
if ($^O eq "freebsd") {
    requires_external_bin 'gmake';
};
```
to Makefile.PL. I will investigate.

matsduf: Now inc/Module/Install/External.pm is included by the second commit. Please review again.

mattias-p (approved):
I haven't tested it but it makes sense to me in principle.

matsduf:
@pnax, can you test again?

pnax (approved):
This works on Ubuntu.

@mattias-p sorry I misclicked for the review (and I don't know how to reset it)

Issue #816, opened 2020-10-22
t/po-files.t depends on Gnu Make and does not work on FreeBSD


matsduf:
This is run on develop branch on commit 81fc8abe4cbf2aa3a12d30f6d266a32c2c08ed07 and on FreeBSD. `t/po-files.t` failed because BSD Make is not compatible with the options given.
```
$ PERL_DL_NONLAZY=1 "/usr/local/bin/perl" "-MExtUtils::Command::MM" "-MTest::Harness" "-e" "undef *Test::Harness::Switches; test_harness(0, 'inc', 'blib/lib', 'blib/arch')" t/po-files.t 
t/po-files.t .. usage: make [-BeikNnqrstWwX] 
            [-C directory] [-D variable] [-d flags] [-f makefile]
            [-I directory] [-J private] [-j max_jobs] [-m directory] [-T file]
            [-V variable] [-v variable] [variable=value] [target ...]

    #   Failed test 'make show-fuzzy exits with value 0'
    #   at t/po-files.t line 31.
    #          got: '2'
    #     expected: '0'
    # Looks like you failed 1 test of 2.
t/po-files.t .. 1/1 
#   Failed test 'no fuzzy marks'
#   at t/po-files.t line 33.
# Looks like you failed 1 test of 1.
t/po-files.t .. Dubious, test returned 1 (wstat 256, 0x100)
Failed 1/1 subtests 

Test Summary Report
-------------------
t/po-files.t (Wstat: 256 Tests: 1 Failed: 1)
  Failed test:  1
  Non-zero exit status: 1
Files=1, Tests=1,  0 wallclock secs ( 0.04 usr  0.00 sys +  0.09 cusr  0.05 csys =  0.18 CPU)
Result: FAIL
Failed 1/1 test programs. 1/1 subtests failed.
```
If `make` is replaced by `gmake` in t/po-files.t then the test will run as it should in FreeBSD.

```
$ PERL_DL_NONLAZY=1 "/usr/local/bin/perl" "-MExtUtils::Command::MM" "-MTest::Harness" "-e" "undef *Test::Harness::Switches; test_harness(0, 'inc', 'blib/lib', 'blib/arch')" t/po-files.t 
t/po-files.t .. ok   
All tests successful.
Files=1, Tests=1,  1 wallclock secs ( 0.02 usr  0.00 sys +  0.10 cusr  0.04 csys =  0.16 CPU)
Result: PASS
```

matsduf:
[...]

PR #819, opened 2020-10-23
Use gmake with Freebsd in po-files.t

matsduf:
This PR resolves issue #816 by using gmake, instead of make, for FreeBSD.

### DIFF
--- a/t/po-files.t
+++ b/t/po-files.t
@@ -16,7 +16,7 @@ if ($^O eq "freebsd") {
     $makebin = 'gmake';
 };
 
-sub makecommand {
+sub make {
     my @make_args = @_;
 
     my $command = join( ' ', $makebin, '--silent', '--no-print-directory', @make_args );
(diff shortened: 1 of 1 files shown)


sandoche2k (dismissed):

t/po-files.t
```diff
@@ -10,10 +10,16 @@ use File::Basename qw( dirname );
...
 chdir 'share' or BAIL_OUT( "chdir: $!" );
 
-sub make {
+my $makebin = 'make';
+if ($^O eq "freebsd") {
+    # This unit test requires GNU Make
+    $makebin = 'gmake';
+};
+
+sub makecommand {
```
mattias-p, line 19: What's the idea behind renaming this procedure? At face value I'd say "make" is a better name.

matsduf: I felt there were too many "make" that makes it confusing. What is the problem om renaming?

matsduf: To avoid discussion on something that really does not matter I reverted the name change.

mattias-p (approved):

Issue #809, opened 2020-10-19
Update Swedish translation


matsduf:
New message IDs (msgid) have been created and some have changed. Updated translation of the message strings (msgstr) is needed. The file is found as `share/sv.po`. See [Instructions for translators].

[Instructions for translators]: https://github.com/zonemaster/zonemaster-engine/blob/develop/docs/Translation-translators.md

matsduf:
resolved by #823

PR #823, opened 2020-10-25
Translate msgid into Swedish

matsduf:
This is an updated Swedish PO files with translation of untranslated `msgid` and also updates of the parameters. This resolves issue #809.

### DIFF
--- a/share/sv.po
+++ b/share/sv.po
@@ -2,8 +2,8 @@ msgid ""
 msgstr ""
 "Project-Id-Version: 1.0.0\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2020-09-20 06:54+0200\n"
-"PO-Revision-Date: 2020-04-21 10:01+0200\n"
+"POT-Creation-Date: 2020-10-23 15:48+0200\n"
+"PO-Revision-Date: 2020-10-25 11:42+0100\n"
 "Last-Translator: mats.dufberg@iis.se\n"
 "Language-Team: Zonemaster Team\n"
 "Language: sv\n"
@@ -52,6 +52,32 @@ msgstr "Varje bakåtuppslagning (PTR) stämmer med motsvarande namnservernamn."
 #, perl-brace-format
 msgid "No response from nameserver(s) on PTR query ({reverse})."
 msgstr "Inget svar från namnservern på PTR-fråga ({reverse})."
+
+#. ADDRESS:TEST_CASE_END
+#. BASIC:TEST_CASE_END
+#. CONNECTIVITY:TEST_CASE_END
+#. CONSISTENCY:TEST_CASE_END
+#. DELEGATION:TEST_CASE_END
+#. DNSSEC:TEST_CASE_END
+#. NAMESERVER:TEST_CASE_END
+#. SYNTAX:TEST_CASE_END
+#. ZONE:TEST_CASE_END
+#, perl-brace-format
+msgid "TEST_CASE_END {testcase}."
+msgstr "TEST_CASE_END {testcase}."
+
+#. ADDRESS:TEST_CASE_START
+#. BASIC:TEST_CASE_START
+#. CONNECTIVITY:TEST_CASE_START
+#. CONSISTENCY:TEST_CASE_START
+#. DELEGATION:TEST_CASE_START
+#. DNSSEC:TEST_CASE_START
+#. NAMESERVER:TEST_CASE_START
+#. SYNTAX:TEST_CASE_START
+#. ZONE:TEST_CASE_START
+#, perl-brace-format
+msgid "TEST_CASE_START {testcase}."
+msgstr "TEST_CASE_START {testcase}."
 
 #. BASIC:DOMAIN_NAME_LABEL_TOO_LONG
 #, perl-brace-format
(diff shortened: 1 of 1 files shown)


share/sv.po
```diff
@@ -1996,23 +1980,20 @@ msgid ""
...
 #. SYSTEM:PACKET_BIG
 #, perl-brace-format
-msgid ""
-"Packet size ({size}) exceeds common maximum size of {maxsize} bytes (try "
-"with \"{command}\")."
-msgstr ""
-"Paketstorleken ({size}) byte överskrider den normalt maximala storleken om "
-"{maxsize} byte. Försök med kommandot '{command}'."
+msgid "Big packet size ({size}) (try with \"{command}\")."
+msgstr "Stort paket (storlek: {size} byte). Försök med kommandot '{command}'."
```
mattias-p, line 2074: IMO `({size} byte)` would be perfectly lagom.

mattias-p (commented):
I do appreciate the tidy commit, but this would have been much easier to review it the tidy commit would have come first. Removing of obsolete messages would also have been a nice preparation. With a clean an tidy start like that having the final tidy included with the substantive changes would have been gravy.

matsduf:
> I do appreciate the tidy commit, but this would have been much easier to review it the tidy commit would have come first.

The tidy was needed after my update to make the lines wrap correctly.

> Removing of obsolete messages would also have been a nice preparation. With a clean an tidy start like that having the final tidy included with the substantive changes would have been gravy.

Then I suggest that the make command is updated to make the PO file, i.e. "make sv.po" should tidy too.

share/sv.po
```diff
@@ -154,54 +180,55 @@ msgstr "IPv6 är aktiverat. Skickar ingen fråga om '{rrtype}' till {ns}."
...
 
-#. CONNECTIVITY:NAMESERVERS_IPV4_WITH_UNIQ_AS
+#. CONNECTIVITY:ERROR_ASN_DATABASE
 #, perl-brace-format
-msgid ""
-"All nameservers in the delegation have IPv4 addresses in the same AS ({asn})."
-msgstr ""
-"Alla IPv4-adresser för namnservrarna i delegeringen finns i samma AS ({asn})."
+msgid "ASN Database error. No data to analyze for {ns_ip}."
+msgstr "Fel på AS-databas. Det finns ingen data att analysera för {ns_ip}."
```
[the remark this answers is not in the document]
matsduf: Sometimes it says AS, sometimes ASN. What do you mean by "database friendly"?

share/sv.po
```diff
@@ -542,6 +592,26 @@ msgstr ""
...
 
+#. DNSSEC:DNSKEY_KSK_NOT_SEP
+#, perl-brace-format
+msgid ""
+"Flags field of DNSKEY record with tag {keytag} returned by nameserver {ns} "
+"has not SEP bit set although DS with same tag is present in parent."
+msgstr ""
+"Namnserver {ns} skickade en DNSKEY-post med \"keytag\" {keytag}, i vilken "
+"SEP-flaggan inte är satt trots att det finns en DS-post som refererar till "
+"samma \"keytag\" i moderzonen.\n"
```
mattias-p, line 603: IMO the comma doesn't really serve a purpose here. The flow of the sentence gets staggered without good reason.

The same goes for the next message.

share/sv.po
```diff
@@ -845,6 +875,26 @@ msgstr ""
...
+#. DNSSEC:NO_MATCHING_RRSIG
+#, perl-brace-format
+msgid ""
+"Nameserver {ns} returned no signature on the DNSKEY RRset that corresponds "
+"to the DNSKEY with tag {keytag} even though there is a DS record in the "
+"parent zone for that DNSKEY record."
+msgstr ""
+"Namnserver {ns} inkluderade ingen kryptografisk signatur på DNSKEY-"
+"uppsättningen (\"RRset\") från DNSKEY-posten med \"keytag\" {keytag} trots "
+"att det finns en DS-post i moderzonen som motsvarar den DNSKEY-posten.\n"
```
mattias-p, line 896: Does the `("RRset")` clarification really help anyone?

share/sv.po
```diff
@@ -1010,16 +1068,18 @@ msgstr ""
...
 msgid ""
-"Nameserver {ns}/{address} responded with an unexpected rcode ({rcode}) on a "
-"DS query for zone {zone}."
+"Nameserver {ns} responded with an unexpected rcode ({rcode}) on a DS query "
+"for zone {zone}."
 msgstr ""
-"Namnserver {ns}/{address} besvarade en DS-fråga för zonen {zone} med den "
-"icke-väntade svarskoden '{rcode}'."
+"Namnserver {ns} besvarade en DS-fråga för zonen {zone} med den icke-väntade "
+"svarskoden \"{rcode}\"."
```
mattias-p, line 1075: I still don't get how `icke-väntad` is better than `oväntad`. To me it's just weird, and I'm both Swedish and familiar with the theory of logic.

share/sv.po
```diff
@@ -154,54 +180,65 @@ msgstr "IPv6 är aktiverat. Skickar ingen fråga om '{rrtype}' till {ns}."
...
+msgid ""
+"All authoritative nameservers have the IPv6 addresses in the same AS ({asn})."
 msgstr ""
-"IPv4-adresserna för de auktoritativa namnservrarna för domännamnet finns i "
-"mer än ett AS ({asn})."
-
-#. CONNECTIVITY:NAMESERVERS_IPV6_NO_AS
-msgid "No IPv6 nameserver address is in an AS."
-msgstr "Ingen namnserver har någon IPv6-adress i något AS."
+"IPv6-adresser för alla auktoritativa namnservrar tillhör samma AS ({asn})."
```
mattias-p, line 232: There's an edge case where the Swedish and English wordings disagree. That is when all name servers have an IPv6 address in the same AS, but some of the name servers have additional IPv6 addresses in different AS.

mattias-p (commented):
I noticed that the tidy commit updated the lines where I made most of my review comments, so I went back and copied them to the right places.

share/sv.po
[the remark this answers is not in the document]
matsduf: In the msgid, sometimes it is AS and other times ASN. More often AS than ASN. When it comes to database there is also
```
msgid "AS database returned no informations for IP address {ns_ip}."
```

share/sv.po
```diff
@@ -675,45 +712,10 @@ msgstr ""
...
-#. DNSSEC:DS_RFC4509_NOT_VALID
-msgid ""
-"Existing DS with digest type 2, while they do not match DNSKEY records, "
-"prevent use of DS with digest type 1 (RFC4509, section 3)."
-msgstr ""
-"Befintlig DS med 'digest' typ 2 stämmer inte med DNSKEY-posten. Det finns en "
-"DS med 'digest' typ 1, men RFC 4509 (stycke 3) tillåter inte att den används "
-"när det finns en DS med 'digest' typ 2."
+"The DS records in the parent zone match DNSKEY records in the child zone."
+msgstr "DS-posterna i moderzonen stämmer med DNSKEY-poster i dotterzonen."
```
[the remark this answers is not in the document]
matsduf: I find 7 instances, of which one is "stämmer överens", one is "överensstämmande" and the rest is "stämmer".

The msgids are not consistent, and think we should then:
1. Decide if consistency on language level is important or not. And how important it is.
2. If it is important enough, we should make the msgids consistent.

share/sv.po
matsduf: Agree. Being there, I reworded the second half too.

share/sv.po
```diff
@@ -510,10 +543,27 @@ msgstr ""
...
 #, perl-brace-format
-msgid "There are both DS and DNSKEY records with key tags {keytags}."
-msgstr "Det finns både DS- and DNSKEY-poster med 'keytag' {keytags}."
+msgid ""
+"DNSKEY record with tag {keytag} returned by nameserver {ns} does not match "
+"the algorithm and hash values in a DS record with same tag in parent zone."
+msgstr ""
+"DNSKEY-posten med \"keytag\" {keytag} som kom från namnservern {ns} har inte "
+"överensstämmande algoritm och hashvärde som DS-posten från moderzonen med "
+"samma \"keytag\"."
```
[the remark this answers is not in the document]
matsduf: Yes. Fix.

share/sv.po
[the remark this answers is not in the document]
mattias-p: If we are to let msgids decide the terminology for the msgstrings, then clearly this one should be ASN.

I'm trying to find out what the most useful terminology would be. I noticed the apparent inconsistency in the msgids so I tried my hand at the search engines. ASN database turns up reasonable answers whereas AS database does not.

Do we have good reason for having AS database more often in the msgids? Otherwise I think we should update our msgids to use the ASN database term. Of course not in this PR, but we could create an issue for it.

share/sv.po
matsduf: OK. Removed those.

share/sv.po
matsduf: Yes, I think it helps those with good DNS knowledge. "DNSKEY-uppsättning" is not established among those.

share/sv.po
matsduf: That is out of scope of this PR. The msgstr was only updated to adjusted the change parameter name change.

share/sv.po
matsduf: Yes, it is an edge case. In most cases a nameserver only has one IPv6 address for DNS. Fixed.

matsduf:
@mattias-p, please rereview.

share/sv.po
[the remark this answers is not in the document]
matsduf: I think AS is better than ASN.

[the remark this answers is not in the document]
matsduf: > If we are to let msgids decide the terminology for the msgstrings, (...)

That was not that I meant. What I meant was that if we should have the language consistent, then we should start with msgid.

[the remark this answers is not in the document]
matsduf: > I'm trying to find out what the most useful terminology would be. I noticed the apparent inconsistency in the msgids so I tried my hand at the search engines. ASN database turns up reasonable answers whereas AS database does not.

If we us "ASN database" then we should also use "ASN" in other messages to make the users being able to connect one with the other.

mattias-p (approved):

PR #827, opened 2020-10-28
Make line endings consistent so msgfmt doesn't crash

mattias-p:
This makes msgfmt happy. That's required to `make all` in `share/` when PO files are present.


pnax (approved):

sandoche2k (approved):

matsduf (approved):
Could we make "tidy" remove any trailing line breaks?

PR #849, opened 2020-11-20
Root hint data updated

matsduf:
The root hint data was not up to date. It is updated by this PR.

* Hint data sorted
  * Sorted by name server name
  * Name listed before address
  * IPv4 listed before IPv6
* Hint data updated
  * Old address replaced by new
  * Added IPv6 address where missing
  * IPv6 address always presented in normal (short) format
* Data files updated for unit tests effect of updated hint data


mattias-p (approved):
I applaud that you sorted the entries! But the changes would have been easier to review if the sorting would have been done in a separate commit.

matsduf:
@mattias-p, please re-review.

mattias-p (approved):

Issue #859, opened 2020-12-29
Zonemaster::Engine v4.0.3 fails installation due to unit test error


matsduf:
Installation of Zonemaster::Engine v4.0.3 (from [v2020.1.2 release]) failes with errors as below. The problem does not sit in the Zonemaster code, but in the unit test that is only executed during installation. The problem does not even sit in the unit test code, but in the recorded data used by the unit test.

The problem has come after the creation of the release and must be due to changes "on the wire" and some leaking of data "from the wire" to the unit test.

**Work-around**
If the only errors that appear are the ones below, then Zonemaster::Engine can safely be installed ignoring the unit tests. This will have no negative impact on the installation.

Installting ignoring the tests
```
cpanm --notest Zonemaster::Engine
```

**Solution**
Solution will be included in next release. Also see https://github.com/zonemaster/zonemaster-engine/pull/803/commits/f7a87d4d7dbd8532a206f2b0d0336937651bae88 in pull-request #803.

**The problem**
Failing installation:

```
# cpanm Zonemaster::Engine
--> Working on Zonemaster::Engine
Fetching http://www.cpan.org/authors/id/Z/ZN/ZNMSTR/Zonemaster-Engine-v4.0.3.tar.gz ... OK
Configuring Zonemaster-Engine-4.000003 ... OK
Building and testing Zonemaster-Engine-4.000003 ... FAIL
! Installing Zonemaster::Engine failed. See
/root/.cpanm/work/1609249816.47539/build.log for details. Retry with --force to force install it.
```

From the log file:

```
(...)
t/Test-dnssec-more.t ...... ok

#   Failed test 'afnic.fr gives ALL_ALGO_SIGNED'
#   at t/Test-dnssec.t line 25.

#   Failed test 'afnic.fr does not give RRSIG_BROKEN'
#   at t/Test-dnssec.t line 36.
# Looks like you failed 2 tests of 270.
t/Test-dnssec.t ...........
Dubious, test returned 2 (wstat 512, 0x200)
Failed 2/270 subtests
        (less 2 skipped subtests: 266 okay)
t/Test-dnssec05-A.t ....... ok
(...)
```

[v2020.1.2 release]: https://github.com/zonemaster/zonemaster/releases/tag/v2020.1.2

matsduf:
Resolved by #864

PR #860, opened 2020-12-29
Add new feature in Profiles to customize some variables values

vlevigneron:
Fixes issue zonemaster/zonemaster-engine#200
Fixes issue zonemaster/zonemaster-engine#859. For this fix, we have found out what is wrong (signature lifetime registered in the file  `t/Test-dnssec.data` is no longer valid if the test is executed too much time after `.data` file creation).


lib/Zonemaster/Engine/Profile.pm
```diff
@@ -751,6 +770,41 @@ matter if they're excluded from this property.
...
+
+At the top level of this data structure are two levels of nested hashrefs.
+The keys of the top level hash are names of test cases.
+The keys of the second level hashes are variable names that should be used in 
+test case in place of test case encoded values.
+The values of the second level hashes are values of those variables.
+
+The following test cases variables are implemented :
+
+=head3 dnssec04.DURATION_LONG_LIMIT
```
matsduf, line 787: I think it makes more sense to have the second level identical to the message tag. I am not sure if you want another level below that or not.
* dnssec04.DURATION_LONG.posint

lib/Zonemaster/Engine/Profile.pm
```diff
@@ -751,6 +770,41 @@ matter if they're excluded from this property.
...
+
+Positive integer value.
+
+Returns DURATION_LONG message tag in case signature lifetime is less
+than DURATION_LONG_LIMIT.
```
matsduf, line 792: Shouldn't it be *more than*?

lib/Zonemaster/Engine/Profile.pm
```diff
@@ -751,6 +770,41 @@ matter if they're excluded from this property.
...
+
+=head3 dnssec04.REMAINING_LONG_LIMIT
```
matsduf, line 794: * dnssec04.REMAINING_LONG.posint

lib/Zonemaster/Engine/Profile.pm
```diff
@@ -751,6 +770,41 @@ matter if they're excluded from this property.
...
+
+Positive integer value.
+
+Returns REMAINING_LONG message tag in case signature remaining time is
+more than REMAINING_LONG_LIMIT (in seconds).
+
+=head3 dnssec04.REMAINING_SHORT_LIMIT
```
matsduf, line 801: * dnssec04.REMAINING_SHORT.posint

lib/Zonemaster/Engine/Profile.pm
```diff
@@ -127,6 +127,25 @@ my %profile_properties_details = (
...
+                    die "Property test_cases_vars support only one value : dnssec04";
+                }
+                foreach my $var_name ( keys %{ ${$_[0]}{$test_name} } ) {
+                    if ( uc($var_name) ne q{DURATION_LONG_LIMIT} and uc($var_name) ne q{REMAINING_LONG_LIMIT} and uc($var_name) ne q{REMAINING_SHORT_LIMIT} ) {
+                        die "Property test_cases_vars.dnssec04 keys have 3 possible values : DURATION_LONG_LIMIT, REMAINING_LONG_LIMIT or REMAINING_SHORT_LIMIT (case insensitive)";
+                    }
+                    if ( ${$_[0]}{$test_name}{$var_name} !~ /^[1-9][0-9]*$/ ) {
+                        die "Property test_cases_vars.dnssec04.$var_name is not a positive integer";
+                    }
+                    ${$_[0]}{$test_name}{uc($var_name)} = delete ${$_[0]}{$test_name}{$var_name};
```
matsduf, line 145: This puts a heavy burden on this module. Wouldn't be better if the code just verifies that the first level matches a test case ID in this profile, the second a message tag in the profile and the third level be valid data type specification. And the data matches that data. Now we only define one datatype, positive integer:

* testcase00.MESSAGE_TAG.posint : 1234

If someone adds a message tag which does not use the feature, then nothing will break. The support can be added by just updating the test case module.

matsduf (changes requested):
Issue #200 also asks for more standardized date expression. If that is not to be included here, it should be noted.

We also should update the test case since the limits are configurable.

lib/Zonemaster/Engine/Profile.pm
vlevigneron: Ok, I'll try to do something more generic.

vlevigneron:
The last commit should address @matsduf remarks.
BUT it has a drawback... It does not pass "make test" with success, and zonemaster-cli does not work anymore. 
That's why I sent it, I do not get  the problem... If one of you could have a look and help me to find what is wrong.

matsduf:
The documentation that you have added to `Profile.pm` should be split up between that and `DNSSEC.pm`.

General documentation about the feature should reside in `Profile.pm`:

> =head2 test_cases_vars
> A complex data structure.
> Specify some variable values used in test cases.
> At the top level of this data structure are two levels of nested hashrefs.
> The keys of the top level hash are names of test cases.
> The keys of the second level hashes are variable names (and their associated 
> type) that should be used in test case in place of test case encoded values.
> The values of the second level hashes are values of those variables.

The specific variables that have been implemented should reside in the same module as has it implemented, in this case `DNSSEC.pm`:

> =head3 dnssec04.DURATION_LONG.postint
> Positive integer value.
> Returns DURATION_LONG message tag in case signature lifetime is less
> than DURATION_LONG.posint.
>
> =head3 dnssec04.REMAINING_LONG.posint
> Positive integer value.
> Returns REMAINING_LONG message tag in case signature remaining time is
> more than REMAINING_LONG.posint (in seconds).
>
> =head3 dnssec04.REMAINING_SHORT.posint
> Positive integer value.
> Returns REMAINING_SHORT message tag in case signature remaining time is
> less than DURATION_LONG.posint (in seconds).

lib/Zonemaster/Engine/Test/DNSSEC.pm
```diff
@@ -2271,4 +2285,25 @@ Check for valid RSA DNSKEY key size
...
 =back
 
+=head1 CONFIGURABLE VARIABLES IN PROFILE
+
+=over
+
+=item test_cases_vars.dnssec04.REMAINING_SHORT (Positive integer value)
+
+Returns REMAINING_SHORT message tag in case signature remaining validity time is
+less than the value provided (in seconds).
```
mattias-p, line 2295: I had to read this sentence several times to understand its meaning. Here's my take on rewording it to make it clearer:

> If the remaining validity time of the signature is less than `test_cases_vars.dnssec04.REMAINING_SHORT` (in seconds) this test case returns the REMAINING_SHORT message tag.

mattias-p (changes requested):

lib/Zonemaster/Engine/Profile.pm
```diff
@@ -230,6 +233,7 @@ sub default {
             $new->set( $property_name, $profile_properties_details{$property_name}{default} );
         }
     }
+    use Data::Dumper;print Data::Dumper::Dumper($new);
```
mattias-p, line 236: This should be removed, right?

lib/Zonemaster/Engine/Profile.pm
```diff
@@ -751,6 +767,21 @@ matter if they're excluded from this property.
 This is because part of their function is to verify that the given domain name
 can be tested at all.
 
+=head2 test_cases_vars.dnssec04.REMAINING_SHORT (Positive integer value)
```
mattias-p, line 770: The text on =head2 lines are anchors you can link to. When the data type is on the =head2 line and you want to link to the heading, you have to include the data type in the link. So, could you put "Positive integer value" on its own line?

lib/Zonemaster/Engine/Profile.pm
```diff
@@ -751,6 +767,21 @@ matter if they're excluded from this property.
...
+
+If the remaining validity time of the signature is less than test_cases_vars.dnssec04.REMAINING_SHORT (in seconds) 
+this test case returns the REMAINING_SHORT message tag.
```
mattias-p, line 773: Could you also spell out the default value here so people updating their profile configuration don't have to figure out where it is in the code?

mattias-p (approved):

matsduf (commented):
I do not think this is a good solution. Values specific to DNSSEC04 should stay in the DNSSEC.pm module, not in the general Profile.pm. If these three values are the only ones, it could be OK, but in the general case it is not good.

mattias-p:
@matsduf As problems arise when we add more test case specific properties we should revisit the design of the Profile module. I can't see any problems at this point, so I claim YAGNI.

matsduf:
> (...) YAGNI.

Yagni? What language is that?

matsduf:
@vlevigneron, this PR also fixes  `t/Test-dnssec.data` and  `t/Test-dnssec.t`. Those fixes are too important to be hidden in this PR. The changes are not even in a commit by themselves with some description. Can you please update the description of this PR with the problem and the fixes? We have been haunted by this type of errors several times. Does your correct solve it for good? Or do we have a problem that signatures expires? Should these unit tests run with a fixed clock, i.e. reset to the time when the data was collected?

mattias-p:
@matsduf said:

> Should these unit tests run with a fixed clock, i.e. reset to the time when the data was collected?

That would be great! I'll create an issue for that.

vlevigneron:
@matsduf For the moment it is more a workaround than a fix. I have disabled the dnnsec unit test that triggers this issue. So with that  `t/Test-dnssec.data` and  `t/Test-dnssec.t` files, the problem will no longer occur but I have to configure a new testing zone that will have a very long signature lifetime if we want to enable again this unit test.

mattias-p:
[YAGNI](https://en.wikipedia.org/wiki/YAGNI) is English.

matsduf:
> @matsduf For the moment it is more a workaround than a fix. I have disabled the dnnsec unit test that triggers this issue. So with that `t/Test-dnssec.data` and `t/Test-dnssec.t` files, the problem will no longer occur but I have to configure a new testing zone that will have a very long signature lifetime if we want to enable again this unit test.

Please write some sentences in the description of this PR that is has resolved the issue and what you have described here.

vlevigneron:
Good idea.

le 11 janv., Mattias Päivärinta a ?crit :
> @matsduf said:
> 
> > Should these unit tests run with a fixed clock, i.e. reset to the time when the data was collected?
> 
> That would be great! I'll create an issue for that.
> 
> -- 
> https://github.com/zonemaster/zonemaster-engine/pull/860#issuecomment-757888318

-- 
	Vincent Levigneron  A.F.N.I.C.  Vincent.Levigneron@afnic.fr

matsduf (approved):
This is good enough as a "pilot" and POC.

matsduf:
@vlevigneron, please merge so that we can get the solution on the unit tests for other PRs.

PR #867, opened 2021-01-20
Rename profile_example_logfilter.json into profile_example.json

pnax:
This PR :
* ~creates a new `share/profile_example.json` file based on `share/profile_example_logfilter.json` with specific content in each file~
* rename the `profile_example_logfilter.json` into `profile_example.json`
* fixes `edns_size` type
* removes obsolete messages
* update messages order and severity level associated to match default and ones in `share/profile.json` to get a digest output when using `diff` on the files (as modified in #866)
* ~points the Profile documentation to `share/profile_example.json`~
* update Profile.md documentation

_edited_


matsduf (commented):
With this PR we have three profile files:

* profile.json -- The profile that is always loaded by a default installation
* profile_example_logfilter.json  -- An example where log filter setting has been added
* profile_example.json -- An example where test case variable setting has been added

I suggest that the second example file is renamed to "profile_example_test_case_vars.json" or something to point out what it is example of.

In the documentation "Creating profile" I think we should list these three profile files to make them more visible, and there describe them short (maybe the text above).

pnax:
I will make my own critic on this PR.

What I had in mind when doing this PR, was to provide a superset of the `profile.json` file where we can find all default values and that can be easy to diff.

I realize, this is not what this PR offers.
Instead, we get 3 very similar and very long files. This looks like a burden to maintain in the near and far future.

Looking at commit https://github.com/zonemaster/zonemaster-engine/commit/ac217820785f77a651a940eda036f4c56fbccb9c , it looks like the new stanza was added to the `profile_example_logfilter.json` file to provide an example whose place was not in the `profile.json` file.

I think it would be easier to keep 2 files here (especially in regard to the length of one profile file) :
* `profile.json` : loaded by default by Zonemsaster
* `profile_example.json` : a complete example of all available stanzas for the profile file (including the `logfilter` and `test_cases_vars` stanzas)

Regarding this PR, going this way means renaming the `profile_example_logfilter.json` file into `profile_example.json` with some updates in the order of the JSON keys.

@matsduf we have discussed the possibility to split the examples into several files. However, while this is discussed in more details, I think it will be easier to keep 2 files here. And I ask myself if it won't ease a future splitting work of the examples as well.

What do you think ?

matsduf:
I think your suggestion works. -- I then suggest that the documentation mentions the example file and points out that it includes section for log filter and test case variables, respectively, not included in the default file. It that way it makes more sense to why we have the example profile.

pnax:
Force pushed :
* rebased on develop
* renamed the `profile_example_logfilter.json` into `profile_example.json`
* applied all the changes from the PR.
* updated documentation.

Please review.

matsduf (approved):
Looks fine.

PR #878, opened 2021-02-11
Encapsulate the fake addresses cache in Z::E::Recursor

### LINKED ISSUE #877: Failure to recognize upper case fake delegation nameservers
Zonemaster fails to recognize fake delegation nameservers when their names are given in upper (or mixed) case. If fake delegation name server are specified and all the names are in upper or mixed case, BASIC02 emits NO_GLUE_PREVENTS_NAMESERVER_TESTS.

IPV4_ENABLED is emitted for each fake nameserver given in lower case, but upper case fake nameservers are lost:

```sh
$ zonemaster-cli --json_stream --level DEBUG --test basic/basic02 --ns ns.nic.se/91.226.36.45 --ns NS3.NIC.SE/91.226.37.45 iis.se | grep IPV4_ENABLED
{"args":{"ns":"ns.nic.se/91.226.36.45","rrtype":"NS"},"level":"INFO","module":"BASIC","tag":"IPV4_ENABLED","testcase":"UNSPECIFIED","timestamp":0.340873003005981}
```

Fake nameservers should be accounted for no matter what casing they're specified in.

N.B. The example commands use Zonemaster-CLI for the sake of simplicity, but the bug is present in Zonemaster-Engine.

mattias-p:
The idea is to make sure that addresses are inserted and queried in the fake addresses cache using the same normalization scheme.

The fake addresses cache is marked as private. Instead of accessing the cache directly, other modules should query it through the new class methods Z::E::Recursor->has_fake_addresses($domain) and Z::E::Recursor->get_fake_addresses($domain, $nsname).

Fixes #877.

### DIFF
--- a/lib/Zonemaster/Engine/Recursor.pm
+++ b/lib/Zonemaster/Engine/Recursor.pm
@@ -5,6 +5,7 @@ use version; our $VERSION = version->declare("v1.0.10");
 use 5.014002;
 use warnings;
 
+use Carp;
 use Moose;
 use JSON::PP;
 use Zonemaster::Engine::Util;
@@ -48,8 +49,10 @@ sub has_fake_addresses {
 
 sub get_fake_addresses {
     my ( undef, $domain, $nsname ) = @_;
+    ( defined $domain ) or croak 'Argument must be defined: $domain';
+
     $domain = lc $domain;
-    $nsname = lc $nsname;
+    $nsname = ( defined $nsname ) ? lc $nsname : q{};
 
     if ( exists $_fake_addresses_cache{$domain}{$nsname} ) {
         return @{ $_fake_addresses_cache{$domain}{$nsname} };
--- a/lib/Zonemaster/Engine/Recursor.pm
+++ b/lib/Zonemaster/Engine/Recursor.pm
@@ -14,7 +14,7 @@ use Zonemaster::Engine;
 my $seed_data;
 
 our %recurse_cache;
-our %fake_addresses_cache;
+our %_fake_addresses_cache;
 
 {
     local $/;
@@ -24,15 +24,39 @@ our %fake_addresses_cache;
 
 sub add_fake_addresses {
     my ( $self, $domain, $href ) = @_;
+    $domain = lc $domain;
 
     foreach my $name ( keys %{$href} ) {
-        push @{ $fake_addresses_cache{$domain}{$name} }, ();
-        foreach my $ip (@{ $href->{$name} }) {
-            push @{ $fake_addresses_cache{$domain}{$name} }, $ip;
+        my @ips = @{ $href->{$name} };
+        $name = lc $name;
+
+        push @{ $_fake_addresses_cache{$domain}{$name} }, ();
+        foreach my $ip ( @ips ) {
+            push @{ $_fake_addresses_cache{$domain}{$name} }, $ip;
         }
     }
 
     return;
+}
+
+sub has_fake_addresses {
+    my ( undef, $domain ) = @_;
+    $domain = lc $domain;
+
+    return !!$_fake_addresses_cache{$domain};
+}
+
+sub get_fake_addresses {
+    my ( undef, $domain, $nsname ) = @_;
+    $domain = lc $domain;
+    $nsname = lc $nsname;
+
+    if ( exists $_fake_addresses_cache{$domain}{$nsname} ) {
+        return @{ $_fake_addresses_cache{$domain}{$nsname} };
+    }
+    else {
+        return ();
+    }
 }
 
 sub recurse {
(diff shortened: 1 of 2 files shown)


lib/Zonemaster/Engine/Recursor.pm
```diff
@@ -24,17 +24,41 @@ our %fake_addresses_cache;
 
 sub add_fake_addresses {
     my ( $self, $domain, $href ) = @_;
+    $domain = lc $domain;
```
matsduf, line 27: Could this be an IDN name with at least one U-label?

mattias-p: It's not explicitly forbidden. But I'm pretty sure such entries have always been ignored.

matsduf: Unless there is a mechanism here to convert to A-label, and that is true for the NS names in "fake delegation", then there will be problem for that test anyway.

Otherwise, down casing should probably not be done on U-labels, or at least not without a warning, for at tool like Zonemaster. Down casing is not part of IDN 2008 handling. It is a preprocessing like converting "www.sunet.se" to "http://www.sunet.se/".

mattias-p: Maybe you could create an issue for that? This PR is about making ASCII domain names case insensitive.

matsduf: My comment above should not be considered to a blocker or anything. We should rather take a decision where the conversion of U-labels should be done, in the core or at the client.

mattias-p: Right.

In general I think Engine should only work with ASCII-labels at the interface level. I.e. ASCII-label format should always be required in inputs and always returned in outputs. In case IDN operations are required to implement the business logic, the IDN conversions (to and from U-labels, etc) should be encapsulated as implementation details.

That way we push IDN conversion towards presentation layer where we want it, we keep it out of the Engine where we want as little of it as possible, and the Engine interface is as simple as possible.

matsduf: The question is if Engine then should refuse to accept any binary labels in the API, but maybe handle such when they come from the wire.

mattias-p: Since this is just a discussion thread, could you approve the PR?

I did not consider binary labels when I wrote the above. What I was arguing is that Engine should be free from IDN to the extent possible, and I extrapolated a bit on what that would look like.

If Engine can pick up binary labels pick up from the wire, it should be able to return them and accept them as well.

mattias-p: @matsduf Can you approve?

mattias-p:
@pnax @vlevigneron @sandoche2k Can you review?

pnax (approved):
Yes, sorry for the delay, I had some vacations :)
I tested it and this looks good.

lib/Zonemaster/Engine/Recursor.pm
```diff
@@ -24,17 +24,41 @@ our %fake_addresses_cache;
...
+    my ( undef, $domain ) = @_;
+    $domain = lc $domain;
+
+    return !!$_fake_addresses_cache{$domain};
+}
+
+sub get_fake_addresses {
+    my ( undef, $domain, $nsname ) = @_;
+    $domain = lc $domain;
+    $nsname = lc $nsname;
```
vlevigneron, line 52: I do not know if we can have a case where $domain and/or $nsname is not defined but it it can happend we should have something like :
    $domain = $domain ? lc $domain : q{};
    $nsname = $nsname ? lc $nsname : q{};

vlevigneron (dismissed):
Not sure that the requested change is needed.

matsduf (approved):
I have no objections.

lib/Zonemaster/Engine/Recursor.pm
mattias-p: $domain always comes from the name attribute of a Zone object. This attribute is required so it can never be undef.
$ns always comes from the name attribute of a Nameserver object. This attribute is not required (I thought it was!) so it can indeed be undef.

I'll make an update to check for undef in the $domain argument and to handle undef in the $ns argument.

mattias-p: Does anyone know why we allow Nameserver to not have names? I tried to look for places where Nameserver->new is called with undef in the name argument, but it's a fair amount of work. Many of the call sites simply propagate the name argument from their own call sites, an so on transitively.

Reviews like this would be much easier if all functions began by validating that all arguments that should be defined really are defined. IMO think we should add such validations to all new code in Engine and make at least a some effort to update old code when we modify it for other reasons. Do you guys agree with this?

mattias-p:
I added another commit to address @vlevigneron's comment. Please check again.

matsduf (approved):

lib/Zonemaster/Engine/Recursor.pm
matsduf: I see no reason to allow name servers without name, and I agree that it is better to have a specific place where that is verified so that other code can assume that.

mattias-p:
## Release testing for v2021.1

I've verified that #877 is fixed on CentOS 7 and CentOS 8.

PR #890, opened 2021-03-24
Use Github Actions instead of Travis

mattias-p:
This PR migrates our CI from Travis to Github Actions. The reason for the change is that the latency of Travis has grown unreasonably. Sometimes we have to wait for hours on end just to get at test to run.

This PR also changes some behaviors of our CI. Here is a list of all changes I'm aware of and as I understand them:
* The new CI is only triggered by `push` and `pull_request` events. This matches the examples for CI in the Github Actions documentation. Travis is configured for those two, but also "Branch or tag creation" as well as some other events that seem a bit nonsensical in this context.
* The new CI is only triggered by the `develop`, `master` and `release/**` branches. This matches the examples for CI in the Github Actions (actually the examples use `main` instead of `master` and `develop`). Travis is configured to be triggered by all branches. If we need CI in more branches than those we can add extend it later.
* The new CI specifies that we use `ubuntu-20.04`. I believe the operating used under Travis is unspecified.
* The new CI installs (almost) all Perl dependencies from CPAN. Travis is configured to install most dependencies from binary packages. When I tried to use custom Perl versions together with binary dependencies on Github Actions I ran into problems. Maybe those problems could be solved, I'm not sure.
* The new CI installs Zonemaster::LDNS is installed with support for elliptic curves. Travis installs it without.
* The new CI tests against both the latest version of Zonemaster::LDNS (from CPAN) and the current develop branch. Travis is configured to install it from a branch in the zonemaster-ldns repo, specifically the branch with the same name as the target branch that triggered Travis.
* The new CI tests against three Perl versions - the newest and oldest supported versions, and one in the middle. Travis is configured to test against all supported Perl versions. The limitation in the new CI was made to counteract the "combinatorial explosion" caused by testing against both latest and develop of Zonemaster::LDNS.
* The new CI cancels all jobs when one job fails. Travis finishes all jobs even if there are failures.


.github/workflows/ci.yml
```diff
@@ -0,0 +1,71 @@
...
+    branches: [ develop, master ]
+  pull_request:
+    branches: [ develop, master ]
+
+jobs:
+  run-tests:
+    strategy:
+      matrix:
+        os:
+          - ubuntu-latest
```
matsduf, line 14: I do not know what we can choose from, but I would think that the latest Ubuntu that is used for Zonemaster would be a better choice, i.e. Ubunut 20.04. Right now Ubuntu 20.10 is the latest, I guess. I think it is better with a fixed value, if possible. It is better if it is changed when we decide to change it.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,71 @@
+name: CI
+
+on:
+  push:
+    branches: [ develop, master ]
+  pull_request:
+    branches: [ develop, master ]
```
matsduf, line 7: Do we have to specify? Else it would be good to have it on all branches. When releasing v2020.2.1 we created a new branch for that. Of course we could update ci.yml for that case.

.github/workflows/ci.yml
mattias-p: My thought was to use `ubuntu-20.04` and `ubuntu-18.04`, since those are the ones we support at the moment. But then all listed Perl versions would be tested on both OS versions. IMO that's excessive. At this point I should probably only be using a single Perl version as well to avoid wasting resources.

.github/workflows/ci.yml
mattias-p: Not sure. I'll look it up.

.github/workflows/ci.yml
mattias-p: I'm thinking the best solution is to test all Perl versions on the latest supported Ubuntu version. Additionally for each other supported Ubuntu version we test the one Perl version that happens to come with that Ubuntu version. Is there something even better we could aim for? (Ubuntu, Windows and MacOs are the only supported OS flavors.)

matsduf: I think latest Ubuntu that we support and all versions of Perl that we support would be a simple configuration to manage.

.github/workflows/ci.yml
mattias-p: The documentation suggested adding "releases/**", so I did that.

matsduf:
If find a number of errors, even though it says "all checks have passed", e.g.
```
not ok 19 - IS_NOT_AUTHORITATIVE # TODO Need to find domain name with that error
#   Failed (TODO) test 'IS_NOT_AUTHORITATIVE'
Error: #   at t/Test-delegation.t line 55.
not ok 20 - NS_IS_CNAME # TODO Need to find domain name with that error
#   Failed (TODO) test 'NS_IS_CNAME'
Error: #   at t/Test-delegation.t line 57.
not ok 21 - REFERRAL_SIZE_LARGE # TODO Need to find domain name with that error
#   Failed (TODO) test 'REFERRAL_SIZE_LARGE'
Error: #   at t/Test-delegation.t line 59.
```

mattias-p:
Yes, those are TODO tests. They aren't expected to succeed. See http://testanything.org/tap-version-13-specification.html.

matsduf:
The firest bullets are fine.

>  * The new CI installs Zonemaster::LDNS from CPAN. Travis is configured to install it from a branch in the zonemaster-ldns repo, specifically the branch with the same name as the target branch that triggered Travis. IMO it's better if we can keep the new version of Engine compatible with the old version of Zonemaster LDNS. If we really need the new Engine to break compatibility with the old Zonemaster LDNS that could be solved, but IMO we should make reasonable efforts to avoid such measures.

If we add a feature to Zonemaster::Engine that depends on a change in Zonemaster::LDNS we will have to wait one cycle before we can introduce that change to Zonemster::Engine. The same thing happens between ::Engine and ::CLI and ::Engine and ::Backend, respectively.

That situation has happend several times, and that is the background to the configuration for Travis that Zonemaster::Engine develop branch uses Zonemaster::LDNS develop branch.

Can you see any other work-around besides doing things in two cycles? Or three if there are two steps of dependency.

Else, I see this as problematic.

>  * The new CI installs Zonemaster::LDNS is installed with support for elliptic curves. Travis installs it without.

That is a good feature.

mattias-p:
> If we add a feature to Zonemaster::Engine that depends on a change in Zonemaster::LDNS we will have to wait one cycle before we can introduce that change to Zonemster::Engine. The same thing happens between ::Engine and ::CLI and ::Engine and ::Backend, respective
>
> That situation has happend several times, and that is the background to the configuration for Travis that Zonemaster::Engine develop branch uses Zonemaster::LDNS develop branch.
>
> Can you see any other work-around besides doing things in two cycles? Or three if there are two steps of dependency.ly.

I agree that we need to be able to release a new version of Zonemaster::Engine that depends on a new version of Zonemaster::LDNS. But I think we should only use that as an escape hatch. Testing this way helps us detect such incompatibilities at an early stage.

As the Zonemaster software matures (and the team too) I expect we will need to release fewer and fewer breaking changes without warning.

That said, there are no technical obstacles to testing Zonemaster::Engine against Zonemaster::LDNS develop branch. In fact it should be easy. When we do need it we can surely implement it.

matsduf:
> That said, there are no technical obstacles to testing Zonemaster::Engine against Zonemaster::LDNS develop branch. In fact it should be easy. When we do need it we can surely implement it.

Do you say that it is possible to let Github Actions test develop branch of Zonemaster::Engine using Zonemaster::LDNS from develop branch? Would that require updating ci.yml for that PR?

What problem do you see of always using Zonemaster::LDNS from develop branch instead of last release from CPAN?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,74 @@
...
+
+      - name: Install Zonemaster::LDNS (latest)
+        if: ${{ matrix.compatibility == 'latest' }}
+        run: |
+          cpanm --notest Module::Install Zonemaster::LDNS
+
+      - name: Install Zonemaster::LDNS (develop)
+        if: ${{ matrix.compatibility == 'develop' }}
+        run: |
+          cpanm --notest Devel::CheckLib Module::Install Module::Install::XSUtil
```
matsduf, line 59: For "latest" only Module::Install is installed from CPAN. For develop Devel::CheckLib Module::Install::XSUtil are added. Why this difference.

mattias-p: I added them because I got error messages without them. I didn't get those error messages for "latest". It's possible that I've changed something else that makes

mattias-p: I went through lots of trial and error in getting this to work at all. I got errors without them for "develop" but not for "latest". I'm not sure why. I can try to remove them and see what happens.

matsduf: Could it be that they get automatically installed from CPAN when Zonemaster::LDNS is installed from CPAN?

mattias-p: No, we need to have them there. If you figure out why please let me know. In the meanwhile I'm going down a few other rabbit holes.

matsduf:
Can you have comments in the yml file? If so, I think it could be valuable to have documentation included. Some time in the future the yml file will be updated or inspected because there is some issue, and then comments can make it easier.

.github/workflows/ci.yml
matsduf: Could it be that they get automatically installed from CPAN when Zonemaster::LDNS is installed from CPAN?

mattias-p:
I added some notes that I believe will be helpful in the future. In general I think this script is pretty straight forward.

.github/workflows/ci.yml
matsduf: Now I cannot see how it looked like when they were not included for "develop", but for "latest" those dependencies are pulled in when Zonemaster-LDNS is installed:
```
--> Working on Zonemaster::LDNS
Fetching http://www.cpan.org/authors/id/Z/ZN/ZNMSTR/Zonemaster-LDNS-2.1.0.tar.gz ... OK
==> Found dependencies: Devel::CheckLib, Module::Install::XSUtil
--> Working on Devel::CheckLib
Fetching http://www.cpan.org/authors/id/M/MA/MATTN/Devel-CheckLib-1.14.tar.gz ... OK
Configuring Devel-CheckLib-1.14 ... OK
Building Devel-CheckLib-1.14 ... OK
Successfully installed Devel-CheckLib-1.14
--> Working on Module::Install::XSUtil
Fetching http://www.cpan.org/authors/id/G/GF/GFUJI/Module-Install-XSUtil-0.45.tar.gz ... OK
Configuring Module-Install-XSUtil-0.45 ... OK
(...)
Building Module-Install-XSUtil-0.45 ... OK
Successfully installed Module-Install-XSUtil-0.45
```

matsduf:
> I added some notes that I believe will be helpful in the future. In general I think this script is pretty straight forward.

Could you please add a note about the modules that are explicitly stated for develop but not latest and why? Or maybe they should be explicitly stated for both because they are direct dependencies of Zonemaster::LDNS?

mattias-p:
> Could you please add a note about the modules that are explicitly stated for develop but not latest and why? Or maybe they should be explicitly stated for both because they are direct dependencies of Zonemaster::LDNS?

I don't know what to say about them that would be helpful. I guess I could make a note that I added them as a result of trial and error but I don't see how that would be helpful to our future selves. To me it's better to leave out "I don't know" comments.

matsduf:
> If you figure out why they're needed in one place but not the other, feel free to add such a comment.

I can see from the output that they are automatically drawn in by cpanm when Zonemaster::LDNS is installed from CPAN. They are direct dependencies and if we use that as a criteria (as with the installation instructions) they should be listed anyway.

When you are done I could create a PR where they are removed from develop to see what it is being said then in the Github action log file.

pnax (commented):
It looks like you specifically run the tests after the installation process. Is it because of the new CI or just a good practice (or anything else) ?

matsduf (approved):

PR #896, opened 2021-05-07
DNSSEC15 test case implementation

vlevigneron:
- DO NOT MERGE at this point, the code is incomplete (messages not completed yet).
- Unit tests have been added. DS15_MISMATCH_CDS_CDNSKEY is missing for the moment.

Updated 2021-05-17:
## How to test this PR

Read the test case specification for DNSSEC15 and verify that the implementation correctly reports issues and non-issues on selected zones to cover as many message tags as possible.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,7 +18,7 @@ jobs:
       matrix:
         compatibility:
           - develop
-          - latest
+          #- latest
         perl:
           - '5.32'
           - '5.26'


lib/Zonemaster/Engine/Test/DNSSEC.pm
```diff
@@ -267,6 +269,10 @@ sub all {
             push @results, $class->dnssec14( $zone );
         }
 
+        if ( Zonemaster::Engine::Util::should_run_test( q{dnssec15} ) ) {
+            push @results, $class->dnssec14( $zone );
```
matsduf, line 273:
```
push @results, $class->dnssec15( $zone );
```

lib/Zonemaster/Engine/Test/DNSSEC.pm
```diff
@@ -2181,6 +2226,179 @@ sub dnssec14 {
...
+        if ( scalar @{ $cdnskey_rrsets{ $ns_ip } } ) {
+            $no_cds_cdnskey = 0;
+        }
+    }
+
+    if ( $no_cds_cdnskey ) {
+        push @results, info( DS15_NO_CDS_CDNSKEY => {} );
+    }
+    else {
+        for my $nss_key ( sort keys %nss ) {
```
matsduf, line 2314: This is a loop over all name server addresses, also those addresses that are ignored above because they do not reply, are not authoritative or return unexpecte RCODE. They should be ignored below too.

lib/Zonemaster/Engine/Test/DNSSEC.pm
vlevigneron: cut/past syndrome

lib/Zonemaster/Engine/Test/DNSSEC.pm
vlevigneron: Just to be sure.
Step 5 of specs states " Repeat the following steps for each name server IP address in NS IP:" Step 7 states the same thing. It is not clear that it should be treated differently. "NS IP" is not supposed to be modified between Setp 5 and Step7.

matsduf: *NS IP* does not change after its creation in big step 4.

In big step 7 we have three possibilities for a name server IP from *NS IP*:
1. It has a key in *CDS RRsets* and the CDS RRset is non-empty.
2. It has a key in *CDS RRsets* and the CDS RRset is empty.
3. It has no key in the *CDS RRsets*

I think the big step 7 should be changed to

```
7.  Repeat the following steps for each name server IP address in *CDS RRsets*:
```

vlevigneron:
@mattias-p @matsduf Any idea why checks fail here ? I have a success when executing a "make test" on my server.

matsduf:
> @mattias-p @matsduf Any idea why checks fail here ? I have a success when executing a "make test" on my server.

Has it failed several times? Always the same (latest, 526)?

lib/Zonemaster/Engine/Test/DNSSEC.pm
vlevigneron: Modified as sppecified in new revision of specs.

vlevigneron:
It should be related to that. I have just merged zonemaster/zonemaster-ldns#113. We'll see if it changes something.

vlevigneron:
Implementation finished with all messages updated according to the specs.
The only thing missing is a broken configuration for unit tests for DS15_MISMATCH_CDS_CDNSKEY message tag.
'make test' works in my environment but not here with CI.
@matsduf please review

vlevigneron:
@matsduf If you are looking for broken zones, look at t/Test-dnssec.t file, I have created a dozen for the unit tests.

matsduf:
> @mattias-p @matsduf Any idea why checks fail here ? I have a success when executing a "make test" on my server.

If you disable "latest" in .github/workflows/ci.yml I think it will work. Please create an issue to restore too.

matsduf (commented):
I have tested by using the test zones. I have compared the Zonemaster output with manual inspection.  Everything looks fine.

The specification asks for name server IP addresses, but there are none.

matsduf:
## Release testing for v2021.1
Works as expected.

PR #926, opened 2021-05-28
Update CentOS installation instruction

mattias-p:
## Purpose

This PR provides some incremental improvements to the installation experience for CentOS. It also performs some clean-up.

## Changes

* On CentOS we're installing a pretty big group of dependencies named Development Tools. The only package we need from this group is actually gcc. This PR replaces Development Tools with gcc.
* The perl-MailTools package is removed from the installation instruction. It's not a direct dependency of ours so it doesn't belong there.
* All yum commands are made non-interactive to make the installation experience less staggered.
* Formatting is cleaned up in the CentOS section of the installation instruction.

## How to test this PR

This PR is implicitly tested as part of installation testing.

### DIFF
--- a/docs/Installation.md
+++ b/docs/Installation.md
@@ -33,28 +33,22 @@ This instruction covers the following operating systems:
 2) Install the [EPEL] repository:
 
    ```sh
-   sudo yum --enablerepo=extras install epel-release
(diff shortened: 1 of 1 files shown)


matsduf (commented):
Looks fine.

matsduf (approved):

Issue #902, opened 2021-05-17
Check MANIFEST when PR is created


matsduf:
Whenever a file is added or removed MANIFEST or MANIFEST.SKIP must be updated. This is something that is often missed by all developers.

Add a test so that Travis/Github action complains if MANIFEST* is not correct, comparable to `make distcheck`.

PR #929, opened 2021-06-01
Replace "make xx.po" with a script

### LINKED ISSUE #930: make xx.po does not always work
If the `share` directory is empty then it works well to run `make sv.po` (as an example). When then `sv.po` it updated I want to run `make sv.po` again to have all lines normalized and make sure that it is valid, but it does not work unless `Zonemaster-Engine.pot` is removed or has its timestamp updated with e.g. `touch`.

mattias-p:
## Purpose

This PR makes the updating of msgids in PO files more robust.

## Context

This PR fixes #930.
It also provides some ground work for fixing #902.

## Changes

This PR changes how translators update their PO files with fresh msgids.

This PR makes the PO files not be make targets anymore. The `share/update-po` script is provided as a replacement for updating PO files with fresh msgids.

The touch-po hack is cleaned up since it's no longer needed after the PO files are demoted from being make targets.

## How to test this PR

Verify that all translator use cases are still supported by verifying all steps in the updated Translation-translators.md.

### DIFF
--- a/Makefile.PL
+++ b/Makefile.PL
@@ -56,10 +56,10 @@ sub MY::postamble {
         # Make FreeBSD use gmake for share/Makefile
         $text = 'GMAKE ?= "gmake"' . "\n"
             . 'pure_all :: share/Makefile' . "\n"
-            . "\t" . 'cd share && $(GMAKE) touch-po all' . "\n";
+            . "\t" . 'cd share && $(GMAKE) all' . "\n";
     } else {
         $text = 'pure_all :: share/Makefile' . "\n"
-            . "\t" . 'cd share && $(MAKE) touch-po all' . "\n";
+            . "\t" . 'cd share && $(MAKE) all' . "\n";
     };
     return $text;
 };
--- a/docs/Translation-translators.md
+++ b/docs/Translation-translators.md
@@ -79,8 +79,8 @@ ids based on fuzzy matching of similar strings. This is not always desirable
 and you can disable fuzzy matching by executing one of the following
 commands instead:
 ```
-make update-po MSGMERGE_OPTS=--no-fuzzy-mathing POFILES=xx.po
-make update-po MSGMERGE_OPTS=--no-fuzzy-mathing
+make update-po MSGMERGE_OPTS=--no-fuzzy-matching POFILES=xx.po
+make update-po MSGMERGE_OPTS=--no-fuzzy-matching
 ```
 
 ## Github preparation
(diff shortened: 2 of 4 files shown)


matsduf (commented):
I am not sure what the gain of this change is, but I see no problem with it so far. I have not yet tested it since this PR is still a draft.

Please review issue #930 which is relevant in this scope.

Also, if we can have the `update-po` script in the share directory, then it would be helpful if the `check-msg-args` script could be moved there too.

I assume that you plan to make zonemaster-cli work the same way in respect to PO updates and MO creation.

mattias-p:
> I am not sure what the gain of this change is

I guess the Purpose section I'd written wasn't very good, so I updated it.

> Please review issue #930 which is relevant in this scope.

That's an example of a problem that is fixed by this PR.

> it would be helpful if the check-msg-args script could be moved [to the share directory] too.

I'm not super happy about littering the share directory with lots of maintenance scripts. Maybe we should move the PO files and the translation tooling into a separate directory? The Gettext documentation suggests having a `po` directory. I'd prefer to do that in a separate PR though.

> I assume that you plan to make zonemaster-cli work the same way

Thanks for pointing that out! I didn't think of that.

matsduf:
> I'm not super happy about littering the share directory with lots of maintenance scripts. Maybe we should move the PO files and the translation tooling into a separate directory? The Gettext documentation suggests having a `po` directory. I'd prefer to do that in a separate PR though.

I think that would be a good idea. PO-files and scripts needed for PO-handling in one directory! That would make translating work easier.

matsduf:
I will try before further comments/approval.

matsduf (approved):
I have tested and I see issues. #930 is resolved by this PR.

matsduf:
@pnax, can you review this PR?

pnax (commented):
With this new script, we are able to update a single PO file (or all PO files with `make update-po`). Do you think it is worth to make it possible to update more than one PO file at the same time (as with the previous Makefile) ?

mattias-p:
`make update-po` still works. It's even documented in the instruction.

matsduf:
@pnax, can you approve?

pnax:
@mattias-p what I meant is that before you could run `make fr.po sv.po` and both files were updated. Now you can only update one PO file (`make update-po` updates all the PO files). I found that if you want to update only 2 files you need to run something like `make update-po POFILES="fr.po sv.po"`. Which is fine, but I thought using the new `update-po` script would be even nicer (with something like `./update-po fr.po sv.po`). But maybe this is just being picky for a small detail.

pnax (approved):
> can you approve ?

Yes totally :)

mattias-p:
@pnax Okay, I see. Maybe you could create an issue for that? Or even a PR? I'd like to have this merged.

matsduf:
## Testing v2021.2
LGTM.

I have used the new solution for translation work several times.

PR #995, opened 2021-10-30
Implementation of DNSSEC10 with new specifications

vlevigneron:
## Purpose

This PR is the implementation of zonemaster/zonemaster#875

## Context

fixes issue zonemaster/zonemaster-engine#992
fixes issue zonemaster/zonemaster-engine#722 (no longer relevant)

## Changes

method Zonemaster::Engine::Test::DNSSEC::dnssec10 complete rewrite
15 new messages tag implemented (no translation yet)
15 old messages removed
Unit tests updated but many missing

## How to test this PR

Use one of the following domains to do the tests:

dnssec10-non-existent-domain-name-exists-01.zft-root.rd.nic.fr
dnssec10-non-existent-domain-name-exists-02.zft-root.rd.nic.fr
dnssec10-non-existent-domain-name-exists-03.zft-root.rd.nic.fr
dnssec10-non-existent-domain-name-exists-04.zft-root.rd.nic.fr
dnssec10-non-existent-domain-name-exists-05.zft-root.rd.nic.fr
dnssec11-inconsistent-dnskey.zft-root.rd.nic.fr
dnssec10-inconsistent-nsec-nsec3.zft-root.rd.nic.fr
nic.se
zonemaster.fr

PR #997, opened 2021-11-05
Updates logentry_args.md

matsduf:
## Purpose

`logentry_args.md` defines argument names to be used in messages, but there is a limitation in the message model. An argument name can only be used once in the same message. This PR specifies a method to extend the name for such cases.

This PR also:

* Updates the headline
* Adds an introduction and some specification
* Splits the table into two tables, defined and not yet defined names
* Removes old list not relevant

## How to test this PR

This is document only, but the new method is already used in #995.

### DIFF
--- a/docs/logentry_args.md
+++ b/docs/logentry_args.md
@@ -1,45 +1,48 @@
-# Argument names for test case messages
+# Arguments for test case messages
 
 
 ## Introduction
 
-This document defines arguments names, and their type of value, for the
-Zonemaster messages. The messages are defined in the Perl modules for the test
-cases, e.g. [Basic.pm], and translated in the PO files, e.g. [fr.po] and [sv.po].
-The arguments are used in the messages, and are there put within curly brackets,
-e.g. as `{ns}`.
+This document defines *arguments*. An *argument* is defined with its name, its
+type of value and its usage and formatting. The *arguments* are used in a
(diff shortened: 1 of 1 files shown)


docs/logentry_args.md
```diff
@@ -1,25 +1,74 @@
-# Types of information in Zonemaster log objects
+# Argument names for test case messages
```
mattias-p, line 1: `Argument names` -> `Arguments`?

This document also concerns itself with the argument values.

docs/logentry_args.md
```diff
@@ -1,25 +1,74 @@
...
+cases, e.g. [Basic.pm], and translated in the PO files, e.g. [fr.po] and [sv.po].
+The arguments are used in the messages, and are there put within curly brackets,
+e.g. as `{ns}`.
+
+When a message is created or updated only argument names defined in this document
+should be used. If there is defined argument name that can be used for the
+message then a new argument name must be defined and this document is to be
+updated.
+
+## Multiple argument of the same type
```
mattias-p, line 17: argument -> arguments

matsduf (commented):
When working on the text, I realized that "message" and "argument" could be used consistently. I have done further updates.

matsduf:
@mattias-p and @vlevigneron, can you (re-)review this?

mattias-p (approved):

matsduf:
@vlevigneron, does this looks good to you?

PR #1010, opened 2021-11-22
Fixes typo

matsduf:
## Purpose

This PR corrects typos.

## Changes

Updates DNSSEC.pm.

## How to test this PR

Verify that messages with message tags DS17_CDNSKEY_IS_NON_SEP or DS17_CDNSKEY_IS_NON_ZONE do not state "Fetched rom" but "Fetched from".

### DIFF
--- a/lib/Zonemaster/Engine/Test/DNSSEC.pm
+++ b/lib/Zonemaster/Engine/Test/DNSSEC.pm
@@ -999,13 +999,13 @@ Readonly my %TAG_DESCRIPTIONS => (
     DS17_CDNSKEY_IS_NON_SEP => sub {
         __x    # DNSSEC:DS17_CDNSKEY_IS_NON_SEP
           'The CDNSKEY record with tag {keytag} has the SEP bit (bit 15) unset. Fetched '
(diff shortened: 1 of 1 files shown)


vlevigneron (dismissed):

mattias-p (approved):
LGTM.

pnax (approved):

mattias-p (approved):

mvw-afnic:
Okay, I was still unaware of that; my bad. I thought that a pull request like this should also update the translations to make sure the same messages still get translated, but having a separate process makes sense as well. In that case, we'll just need to remind ourselves to update the translations before the next release.

matsduf:
> Okay, I was still unaware of that; my bad. I thought that a pull request like this should also update the translations to make sure the same messages still get translated, but having a separate process makes sense as well.

Since we have several languages we must involve several translators. We have lately done the final translation in a fix release just after the release. That works well because translation is done in separate files.

>  In that case, we'll just need to remind ourselves to update the translations before the next release.

Yes. It is already in my mind.

PR #1018, opened 2021-11-30
Clean up obsolete documentation

mattias-p:
## Purpose

This removes a piece of obsolete documentation.

## Context

The separated config and policy system was replaced with the combined profile system in #446.

While it's late in the release cycle I believe it could be included in the release without problems. On the other hand it isn't urgent either, so it could also be postponed to the next release without problems.

## Changes

This PR touches documentation only. It doesn't add anything new but it removes a source of confusion. I had a look and I couldn't find anything in the removed file that isn't already documented elsewhere.

## How to test this PR

N/A


matsduf (approved):

matsduf (changes requested):
You have to update MANIFEST too.

mattias-p:
> You have to update MANIFEST too.

Yeah, the CI tests noticed that too. I made an update.

matsduf (approved):

pnax (approved):

PR #1023, opened 2021-12-03
 Preparation for release v2021.2 (Engine)

matsduf:
## Purpose

* Bumps version in perl module.
* Updates Changes file

## Changes

Updated files:
* Changes
* lib/Zonemaster/Engine.pm


mattias-p (approved):
There's an inconsistency in phrasing among the items. There's one noun phrase while all the others are verb phrases. I don't think that's a huge problem. Personally like it better when the items are noun phrases.

PR #1026, opened 2021-12-08
Updated the Norwegian language translation

### LINKED ISSUE #1006: Update translation for Norwegian (Engine)
There are new messages that need translation. Follow the [Instructions for translators](https://github.com/zonemaster/zonemaster/blob/master/docs/internal-documentation/maintenance/Instructions-for-translators.md).

repperille:
## Purpose

This PR includes updated translations

## Context

Fixes #1006

## Changes

...

## How to test this PR

...

### DIFF
--- a/share/nb.po
+++ b/share/nb.po
@@ -3,7 +3,7 @@ msgstr ""
 "Project-Id-Version: 1.0.0\n"
 "Report-Msgid-Bugs-To: \n"
 "POT-Creation-Date: 2021-11-16 12:16+0100\n"
-"PO-Revision-Date: 2021-12-08 13:13+0100\n"
+"PO-Revision-Date: 2021-12-09 08:49+0100\n"
 "Last-Translator: richard.persson@norid.no\n"
 "Language-Team: Zonemaster Team\n"
 "Language: nb\n"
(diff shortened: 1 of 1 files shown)


share/nb.po
```diff
@@ -582,324 +585,709 @@ msgstr ""
...
-msgstr "RRSIG for DNSKEY-posten opprettet av \"key tag\" {keytag} returnert "
-"av navneserveren {ns} kunne ikke bekreftes med feilen \"{error}\" (en "
-"DS-post med samme tag er tilstede i foreldresonen)."
+"Flags field of DNSKEY record with tag {keytag} has not ZONE bit set although "
+"DS with same tag is present in parent. Fetched from the nameservers with IP "
+"addresses \"{ns_ip_list}\"."
+msgstr ""
+"Flagg-feltet for denne DNSKEY-posten med \"key tag\" {keytag}, "
+"indikerer at det er en ZSK men DS-post med samme \"key tag\" finnes i "
+"foreldre-zonen. Hentet fra navnetjenere med IP-adressene ({ns_ip_list})."
```
matsduf, line 652: The ZONE bit is not ZSK or KSK. It is whether the key is for zone signing at all, i.e. for DNSSEC. If the ZONE bit is not set the key is for something else. 

"Bit 7 of the Flags field is the Zone Key flag.  If bit 7 has value 1, then the DNSKEY record holds a DNS zone key, and the DNSKEY RR's owner name MUST be the name of a zone.  If bit 7 has value 0, then the DNSKEY record holds some other type of DNS public key and MUST NOT be used to verify RRSIGs that cover RRsets."

https://datatracker.ietf.org/doc/html/rfc4034#section-2.1.1

share/nb.po
```diff
@@ -582,324 +585,709 @@ msgstr ""
...
+msgid "Parent name servers are inconsistent on the existence of DS."
+msgstr "Navnetjenerne svarer inkonsekvent på spørring om DS-poster."
+
+#. DNSSEC:DS11_INCONSISTENT_SIGNED_ZONE
+msgid ""
+"Name servers for the child zone are inconsistent on whether the zone is "
+"signed or not."
+msgstr ""
+"Navnetjenerne for sonen svarer inkonsekvent på om sonen er signert eller "
+"ikke.<"
```
matsduf, line 934: An extra "<" at the end.

matsduf (changes requested):
I found two items where update is needed.

matsduf (approved):
Looks good. I approve.

PR #1027, opened 2021-12-09
"Fix" dnssec07 profile.json list order

andreaso:
Technically this presumably doesn't matter, but it did manage to
through me for a loop while reading _profile.json_.


matsduf:
Thanks for the proposal. We will check if there is a reason for the disorder.

pnax (commented):
From what I see, this order has been introduced with commit b9c4c94d1c06ac7e9189d1905139d25258d5cb97. And it seems this has been introduced for a good reason, [DNSSEC06](https://github.com/zonemaster/zonemaster/blob/master/docs/specifications/tests/DNSSEC-TP/dnssec06.md#intercase-dependencies) says:

> This test should only run if [DNSSEC07](https://github.com/zonemaster/zonemaster/blob/master/docs/specifications/tests/DNSSEC-TP/dnssec07.md) has been successful in finding a DNSKEY for the domain.

The question that remains: is the order relevant in the profile.json file?

pnax (approved):
> The question that remains: is the order relevant in the profile.json file?

From what I see, the order is not relevant. So we can use a natural order here.

matsduf (approved):

Issue #945, opened 2021-07-21
Move SOA values from Constants to profile


matsduf:
In `Constants.pm` the following SOA values are found:
```
Readonly our $SOA_DEFAULT_TTL_MAXIMUM_VALUE => 86_400;     # 1 day
Readonly our $SOA_DEFAULT_TTL_MINIMUM_VALUE => 300;        # 5 minutes
Readonly our $SOA_EXPIRE_MINIMUM_VALUE      => 604_800;    # 1 week
Readonly our $SOA_REFRESH_MINIMUM_VALUE     => 14_400;     # 4 hours
Readonly our $SOA_RETRY_MINIMUM_VALUE       => 3_600;      # 1 hour
```

Being there prevents the user from doing a simple policy change. It would be better to keep them in the policy.

The suggestion came from https://github.com/zonemaster/zonemaster/issues/972.

Ressy66:
ok after playing merry-go-rounds with all these issue numbers constantly redirecting all over the place, i'll ask here one more time, is this implemented, if so what is the filename/location that is edited for custom values, it really shouldnt be this hard.
or do we just keep editing constants which seems simplest for those who have found it

matsduf:
No, this is not yet implemented. There is not even a firm decision to have it implemented. I'm sorry, it will take some time.

As long as this issue is kept open it is not implemented. Any implementation will be done through a pull request that will be linked from this issue.

matsduf:
A proposal is found in #1032.

pnax:
Implemented with #1032

PR #1032, opened 2021-12-23
Config SOA values

pnax:
## Purpose

Make default SOA values configurable.

## Context

Addresses #945

## Changes

Remove hard coded values and replace them with Profile values.
Update Zone.pm file to use the profile values.

## How to test this PR

This PR comes with its own tests using `zone02.zut-root.rd.nic.fr`.
To manually test, update or create a Profile with one of the `test_cases_vars.zone` keys and run zonemaster-cli.

```
    "test_cases_vars": {
        "zone" : {
            "SOA_DEFAULT_TTL_MAXIMUM_VALUE" : 3600,
            "SOA_DEFAULT_TTL_MINIMUM_VALUE" : 7200,
            "SOA_EXPIRE_MINIMUM_VALUE"      : 86400,
            "SOA_REFRESH_MINIMUM_VALUE"     : 86400,
            "SOA_RETRY_MINIMUM_VALUE"       : 7200
        }
    }
```
```
zonemaster-cli --show-module --show-test --test zone --raw zone02.zut-root.rd.nic.fr
```


matsduf:
I think this looks good.

lib/Zonemaster/Engine/Profile.pm
```diff
@@ -812,6 +812,43 @@ Returns DURATION_LONG message tag in case signature lifetime is more
...
+seconds).
+Default C<300> (5 minutes in seconds).
+
+=head2 test_cases_vars.zone.SOA_DEFAULT_TTL_MAXIMUM_VALUE
+
+A positive integer value.
+Returns SOA_DEFAULT_TTL_MAXIMUM_VALUE_HIGHER message tag if the SOA minimum
+value is greater than test_cases_vars.zone.SOA_DEFAULT_TTL_MAXIMUM_VALUE (in
+seconds).
+Default C<86400> (1 day in seconds).
```
matsduf, line 850: I think these could be shorter by only having
```
=head2 test_cases_vars.zone.SOA_DEFAULT_TTL_MAXIMUM_VALUE

A positive integer value.
Default C<86400> (1 day in seconds).
```

I think it would easier to read. The remaining information belongs to the test case method.

pnax: To be honest I followed how it was done for `test_cases_vars.dnssec04` items.
Do you suggest to shorten the documentation for these items too?

pnax: see https://metacpan.org/pod/Zonemaster::Engine::Profile#test_cases_vars.dnssec04.REMAINING_SHORT

matsduf: I understand that, and when there were just a few it works, but with many I feel that it should be shorter, and I feel that the extra information does not really belong here. What do you think?

I you agree with me, please shorten all to make it more readable, maybe even on one line.

matsduf: I guess one should use `perldoc` to inspect the POD before merging.

pnax: Thinking about it, I'm realizing that if we remove these sentences, we might be losing some information on how these parameters interact with the tests. Since I can't see anywhere else where we have specified this interaction, I think it might be good to keep it here.

matsduf: Could a short section with the headling "test_cases_vars" be put before the list of "vars" that shortly describes that these are related to messages with the same name as after the dot?

t/Test-zone.t
```diff
@@ -71,6 +71,53 @@ ok( $res{RETRY_MINIMUM_VALUE_OK},     q{SOA 'retry' value is more than the minim
...
+        %res = map { $_->tag => 1 } Zonemaster::Engine->test_method( q{Zone}, q{zone06}, $zone );
+        ok( $res{SOA_DEFAULT_TTL_MAXIMUM_VALUE_HIGHER}, q{SOA 'minimum' value is too high} );
+    };
+
+    # reset the profile
+    Zonemaster::Engine::Profile->effective->set( q{test_cases_vars.zone.SOA_RETRY_MINIMUM_VALUE}, $old_retry );
+    Zonemaster::Engine::Profile->effective->set( q{test_cases_vars.zone.SOA_REFRESH_MINIMUM_VALUE}, $old_refresh );
+    Zonemaster::Engine::Profile->effective->set( q{test_cases_vars.zone.SOA_EXPIRE_MINIMUM_VALUE}, $old_expire );
+    Zonemaster::Engine::Profile->effective->set( q{test_cases_vars.zone.SOA_DEFAULT_TTL_MINIMUM_VALUE}, $old_ttl_min );
+    Zonemaster::Engine::Profile->effective->set( q{test_cases_vars.zone.SOA_DEFAULT_TTL_MAXIMUM_VALUE}, $old_ttl_max );
```
mattias-p, line 118: Maybe use `Zonemaster::Engine::Profile->effective->merge( Zonemaster::Engine::Profile->default )` instead of restoring individual properties?

mattias-p (approved):
LGTM. I have a small suggestion.

lib/Zonemaster/Engine/Profile.pm
```diff
@@ -142,6 +142,31 @@ my %profile_properties_details = (
         type    => q{Num},
         min     => 1,
         default => $DURATION_180_DAYS_IN_SECONDS
+    },
+    q{test_cases_vars.zone.SOA_DEFAULT_TTL_MAXIMUM_VALUE} => {
```
matsduf, line 146: The already existing variable have variable names with the test case identifier where it is used. I think the is a good principle, i.e. for this variable:

test_cases_vars.zone.SOA_DEFAULT_TTL_MAXIMUM_VALUE => test_cases_vars.zone06.SOA_DEFAULT_TTL_MAXIMUM_VALUE

lib/Zonemaster/Engine/Profile.pm
matsduf: My error. They are values used in specific test case, but they are not directly settings of messages.

The structure of the name, so far, is "test_cases_vars." + %Test-Case-ID% + "." + %VAR-NAME%

matsduf (changes requested):

t/Test-zone.t
pnax: Thanks for the tip. I made the change.

lib/Zonemaster/Engine/Profile.pm
pnax: Updated to follow the suggested format.

pnax:
@matsduf can you re-review?

lib/Zonemaster/Engine/Constants.pm
```diff
@@ -70,8 +70,13 @@ Readonly our $ALGO_STATUS_NOT_RECOMMENDED => 9;
...
 
-Readonly our $DURATION_12_HOURS_IN_SECONDS => 12 * 60 * 60;
-Readonly our $DURATION_180_DAYS_IN_SECONDS => 180 * 24 * 60 * 60;
+Readonly our $DURATION_5_MINUTES_IN_SECONDS  =>             5 * 60;
+Readonly our $DURATION_1_HOUR_IN_SECONDS     =>            60 * 60;
+Readonly our $DURATION_4_HOURS_IN_SECONDS    =>        4 * 60 * 60;
+Readonly our $DURATION_12_HOURS_IN_SECONDS   =>       12 * 60 * 60;
+Readonly our $DURATION_1_DAY_IN_SECONDS      =>       24 * 60 * 60;
+Readonly our $DURATION_1_WEEK_IN_SECONDS     =>   7 * 24 * 60 * 60;
+Readonly our $DURATION_180_DAYS_IN_SECONDS   => 180 * 24 * 60 * 60;
```
matsduf, line 79: Why not provide them as 300, 3600 etc?

lib/Zonemaster/Engine/Profile.pm
```diff
@@ -787,6 +812,43 @@ Returns DURATION_LONG message tag in case signature lifetime is more
 than test_cases_vars.dnssec04.DURATION_LONG (in seconds).
 Default C<15552000> (180 days in seconds).
 
+=head2 test_cases_vars.zone02.SOA_REFRESH_MINIMUM_VALUE
+
+A positive integer value.
+Returns REFRESH_MINIMUM_VALUE_LOWER message tag in case the SOA refresh value
+is less than test_cases_vars.zone02.SOA_REFRESH_MINIMUM_VALUE (in seconds).
+Default C<14400> (4 hours in seconds).
```
matsduf, line 820: I do not think the exact message tag should or need to be documented here. When the tag is updated (the current tag does not match the current template) then it has to be updated here too. I think it should be enough to refer to the test case specification and test case implementation. E.g.:

```
A positive integer value.
For its use, see the implementation in Zone.pm (in this repository) and zone02.md
(in the zonemaster/zonemaster repository).
Default C<14400> (4 hours in seconds).
```

This comment also applies to the definitions below.

matsduf (approved):
This looks good, but I have two comments that are not crucial. I approve.

lib/Zonemaster/Engine/Profile.pm
mattias-p: Since this is a configuration parameter rather than a method, I don't think it's appropriate to say that it "returns" something. I suggest simply stating what it represents.

```
A positive integer value.
The lower bound recommended for SOA refresh values.
Default C<14400> (4 hours in seconds).
```

lib/Zonemaster/Engine/Constants.pm
pnax: I kept to way it was done before. While implementing it I also felt it would make it easier to review the values.

lib/Zonemaster/Engine/Profile.pm
pnax: I updated the documentation, trying to take from both suggestions.

matsduf: You still refer to the exact message tag, which I find unnecessary and will increase the risk that it is not updated when the tag is replaced. It will not break anything, but the more configuration items there are the more important it is to keep them short.

matsduf (approved):
I have left a comment, but I approve anyway.

lib/Zonemaster/Engine/Profile.pm
pnax: There is a tight coupling between the tag and the configuration item. I think it helps to have the documentation highlights this coupling.

Issue #920, opened 2021-05-25
Nameserver14 is not implemented


matsduf:
Test case [Nameserver14] is defined but it is not implemented in [Nameserver.pm].

[Nameserver.pm]: https://github.com/zonemaster/zonemaster-engine/blob/develop/lib/Zonemaster/Engine/Test/Nameserver.pm

[Nameserver14]: https://github.com/zonemaster/zonemaster/blob/develop/docs/specifications/tests/Nameserver-TP/nameserver14.md

mattias-p:
Maybe we should remove the specification from v2021.1 and reinstate it once we have an implementation? That way it's clearer what is implemented and what's not.

vlevigneron:
I'll have a look. If it's easy to implement, I'll add the code this week.

matsduf:
@vlevigneron, we can wait until next release.

matsduf:
For v2022.1.

marc-vanderwal:
All links to Nameserver14 specifications are now broken. Is this issue still relevant?

matsduf:
It is stated in the specification

> This test case actually combines the test options in test cases [NAMESERVER10](https://github.com/zonemaster/zonemaster/blob/master/docs/public/specifications/tests/Nameserver-TP/nameserver10.md) and [NAMESERVER11](https://github.com/zonemaster/zonemaster/blob/master/docs/public/specifications/tests/Nameserver-TP/nameserver11.md).

I am not convinced that it is really meaningful. In theory a name server can behave different in the combination, but I think it is less likely, an I think we should other specification and coding.

For me it can be removed.

PR #1033, opened 2022-01-03
Implementation of Nameserver14

tgreenx:
## Purpose

This PR proposes an implementation of test procedure "Nameserver14" as described [here](https://github.com/zonemaster/zonemaster/blob/develop/docs/specifications/tests/Nameserver-TP/nameserver14.md).

## Context

Addresses #920


matsduf:
I can see that the specification needs an update to avoid multiple messages for the same error. I would probably be good to move it into the new format specified by the template before implementation.

PR #1034, opened 2022-01-03
New implementation of Nameserver11 - #1110

tgreenx:
## Purpose

This PR proposes a new implementation for Nameserver11, following the update of the specification in https://github.com/zonemaster/zonemaster/pull/1111.

## Context

Addresses https://github.com/zonemaster/zonemaster/issues/1110 (previously https://github.com/zonemaster/zonemaster/issues/993)

## Testing

```
zonemaster-cli --test=nameserver/nameserver11 bemacom.se --level=INFO --show-testcase --raw
   3.83 WARNING   NAMESERVER11   N11_UNEXPECTED_RCODE   ns_ip_list=194.17.14.66;92.33.14.98; rcode=FORMERR
```


matsduf:
The process we have followed so far is to have the specification completed before implementation starts. In that we lower that risk of have work to be redone.

vlevigneron:
I agree with @matsduf. We should follow the same process we had before. Specifications first, then implementation.
BTW @tgreenx, some links in your PR are wrong, for instance you refer to zonemaster/zonemaster-engine#993 instead of zonemaster/zonemaster#993 in many places

lib/Zonemaster/Engine/Test/Nameserver.pm
```diff
@@ -1077,44 +1078,30 @@ sub nameserver11 {
...
-                NO_RESPONSE => {
-                    ns     => $ns->string,
-                    domain => $zone->name,
-                }
-              );
-        }
+    }
 
+    if ( scalar @unknown_opt_code ){
+        push @results, info( NS11_UNKNOWN_OPTION_CODE => { ns => join( q{;}, uniq sort @unknown_opt_code ) } );
```
matsduf, line 1104: The specification and the message defined on lines 363-365 says "ns_ip_list", here it is "ns".

The example in the PR description indicates that it is outputted as `ns_list` rather than `ns_ip_list`.
```
1.07 DEBUG Nameserver ns3.securedns.ch/77.109.136.195 responds with an unknown EDNS OPTION-CODE.
```

It is not wrong to have `ns_list` but then the specification must be adjusted first.

matsduf (changes requested):

lib/Zonemaster/Engine/Test/Nameserver.pm
tgreenx: Indeed, I missed that, it is fixed now. The example in the description is also outdated (it will not output the same thing anymore), I will remove it.

Note that we also have no unit test for this test case, could you maybe create test zones for it? And I'll then make the unit tests.

tgreenx:
Commit 39bb1c7 addresses PR [zonemaster#1093](https://github.com/zonemaster/zonemaster/pull/1093).

lib/Zonemaster/Engine/Test/Nameserver.pm
matsduf: > Note that we also have no unit test for this test case, could you maybe create test zones for it? And I'll then make the unit tests.

I will do that, but it might take a while since it cannot be done by a normal name server like Bind.

matsduf: We may have found a tool that permits the creation of response that we need for the unit test.

matsduf:
@tgreenx, can https://github.com/zonemaster/zonemaster/pull/1111 be merged and incorporated in this update?

tgreenx:
> @tgreenx, can [zonemaster/zonemaster#1111](https://github.com/zonemaster/zonemaster/pull/1111) be merged and incorporated in this update?

Yes

matsduf (approved):
This looks fine, and I see no obvious error. There are, however, not enough unit tests or test zones to verify this implementation. I think we can merge, and then let the upcoming tests zones that are defined in https://github.com/zonemaster/zonemaster/pull/1113 verify it. In that PR there is a test-data specification for the test case.

lib/Zonemaster/Engine/Test/Nameserver.pm
```diff
@@ -1077,44 +1103,89 @@ sub nameserver11 {
...
+                push @unexpected_answer, $ns->address->short;
             }
-            else {
-                push @results, info( NS_ERROR => { ns => $ns->string, } );
+            
+            elsif ( not $p->aa ){
+                push @unset_aa, $ns->address->short;
+            }
+            
+            elsif ( defined $p->edns_data ) {
```
mattias-p, line 1151: This check is more general than what is prescribed in the specification. It doesn't seem like a huge deal, but I wanted to point it out.

mattias-p (approved):
LGTM. I found one small thing that could be improved but from my understanding it's not big enough to block merging.

lib/Zonemaster/Engine/Test/Nameserver.pm
matsduf: I looked at that, and since the query has no flags and no other options in EDNS, besides the undefined option, which the server is expected to ignore and not include in the response, I think this should work.

Issue #1036, opened 2022-01-14
Fatal error in DNSSEC: Key list is empty at /usr/local/share/perl/5.26.1/Zonemaster/Engine/Test/DNSSEC.pm line 2627.


matsduf:
The second error is some kind of bug. (The first error is test case specification issue.)

```
$ zonemaster-cli atiger.se --show-testcase --level error --test dnssec --raw
   9.34 ERROR     DNSSEC09       DS09_NO_MATCHING_DNSKEY   keytag=53612; ns_ip_list=185.42.137.98;194.58.192.46;194.58.198.32;2a01:3f0:400::32;2a01:3f1:3032::53;2a01:3f1:46::53
   9.36 CRITICAL  UNSPECIFIED    MODULE_ERROR   module=DNSSEC; msg=Key list is empty at /usr/local/share/perl/5.26.1/Zonemaster/Engine/Test/DNSSEC.pm line 2627.
```
Saved data: [atiger.se-issue-1035-2022-01-14.save.gz](https://github.com/zonemaster/zonemaster/files/7869581/atiger.se-issue-1035-2022-01-14.save.gz)

Also see https://github.com/zonemaster/zonemaster/issues/1035

mattias-p:
There are actually three bugs in play here. The on in the business logic is addressed in #1037. But there are also two error handling bugs.
1. We're leaking internal details in the messages. The msg value should be something generic like "Internal error" without references into the code. The exception itself should be sent to Log::Any so that it can be handled appropriately by CLI or Backend or whatever application.
2. We're failing to identify where the error originated, labeling the test case as "UNSPECIFIED".

mattias-p:
> When running CLI it should be OK to reveal the code. In many cases I think we want to be able to get that information with CLI.

Yes, for CLI a log adapter that prints to STDERR would be quite appropriate.

PR #1037, opened 2022-01-14
- Fix dnssec10 case when there are DNSKEY empty lists

vlevigneron:
## Purpose

This PR fixes dnssec10 case when there are no DNSKEY availables but RRSIG exist. Which is possible during a "conservative" Key rollover (Knot does that, RRSIG are published before the keys)

## Context

Fixes zonemaster/zonemaster-engine#1036

## Changes

Changes in dnssec10 to prevent to do some tests when no DNSKEY is found.

## How to test this PR

 zonemaster-cli atiger.se --show-testcase --level error --test dnssec --raw

### DIFF
--- a/lib/Zonemaster/Engine/Test/DNSSEC.pm
+++ b/lib/Zonemaster/Engine/Test/DNSSEC.pm
@@ -5,7 +5,7 @@ use 5.014002;
 use strict;
 use warnings;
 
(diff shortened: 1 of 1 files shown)


vlevigneron:
@matsduf I did not changed levels of messages yet. This fix just prevents the engine to have a fatal error.

tgreenx (approved):

PR #1040, opened 2022-02-21
add new basic00 implementation

blacksponge:
## Purpose

Implement new Basic00. I implemented the updated "test case" as sanitization methods, it is not used anywhere yet, I am waiting for design validation before using it.

## Context

https://github.com/zonemaster/zonemaster/pull/942

## Changes

* Add `Zonemaster::Engine::Normalization::normalize_name` that follow the procedure described in [RequirementsAndNormalizationOfDomainNames](https://github.com/zonemaster/zonemaster/blob/2419f32fe36c3118393e188022c7648d59c60232/docs/specifications/tests/RequirementsAndNormalizationOfDomainNames.md).
  * The method return a list of error and the normalized domain name if it is valid.
  * If the domain is invalid, the domain is undef, if it is valid the list of error is empty.
  * Errors are object of type `Zonemaster::Engine::Normalization::Errors`. Those have a tag and a localized error message (translated when called to allow changes of the locale between calls)
 * Documentation (waiting for design validation)

## How to test this PR

* A new test file "t/normalization.t" is testing the method.

### DIFF
--- a/lib/Zonemaster/Engine/Normalization.pm
+++ b/lib/Zonemaster/Engine/Normalization.pm
@@ -65,14 +65,14 @@ sub normalize_label {
     if ( $label =~ $VALID_ASCII ) {
         $alabel = lc $label;
     } elsif ( $label =~ $ASCII ) {
-        push @messages, Zonemaster::Engine::Normalization::Errors->new('INVALID_ASCII' => {label => $label});
+        push @messages, Zonemaster::Engine::Normalization::Errors->new(INVALID_ASCII => {label => $label});
 
         return \@messages, undef;
     } elsif (Zonemaster::LDNS::has_idn) {
         try {
             $alabel = Zonemaster::LDNS::to_idn($label);
         } catch {
-            push @messages, Zonemaster::Engine::Normalization::Errors->new('INVALID_U_LABEL' => {label => $label});
+            push @messages, Zonemaster::Engine::Normalization::Errors->new(INVALID_U_LABEL => {label => $label});
 
             return \@messages, undef;
         }
@@ -81,7 +81,7 @@ sub normalize_label {
     }
 
     if ( length($alabel) > 63) {
-        push @messages, Zonemaster::Engine::Normalization::Errors->new('LABEL_TOO_LONG' => {label => $label});
+        push @messages, Zonemaster::Engine::Normalization::Errors->new(LABEL_TOO_LONG => {label => $label});
         return \@messages, undef;
     }
 
@@ -96,7 +96,7 @@ sub normalize_name {
     $uname =~ s/${WHITE_SPACES_RE}+$//;
 
     if (length($uname) == 0) {
-        push @messages, Zonemaster::Engine::Normalization::Errors->new('EMPTY_DOMAIN_NAME');
+        push @messages, Zonemaster::Engine::Normalization::Errors->new(EMPTY_DOMAIN_NAME => {});
         return \@messages, undef;
     }
 
@@ -113,17 +113,17 @@ sub normalize_name {
 
     $uname =~ s/${FULL_STOPS_RE}/${ASCII_FULL_STOP}/g;
 
-    if ( $uname eq '.' ) {
+    if ( $uname eq $ASCII_FULL_STOP ) {
         return \@messages, $uname;
     }
 
     if ($uname =~ m/^${ASCII_FULL_STOP_RE}/) {
-        push @messages, Zonemaster::Engine::Normalization::Errors->new('INITIAL_DOT');
+        push @messages, Zonemaster::Engine::Normalization::Errors->new(INITIAL_DOT => {});
         return \@messages, undef;
     }
 
     if ($uname =~ m/${ASCII_FULL_STOP_RE}{2,}/ ) {
-        push @messages, Zonemaster::Engine::Normalization::Errors->new('REPEATED_DOTS');
+        push @messages, Zonemaster::Engine::Normalization::Errors->new(REPEATED_DOTS => {});
         return \@messages, undef;
     }
 
@@ -144,7 +144,7 @@ sub normalize_name {
     my $final_name = join '.', @label_ok;
 
     if (length($final_name) > 253) {
-        push @messages, Zonemaster::Engine::Normalization::Errors->new('DOMAIN_NAME_TOO_LONG');
+        push @messages, Zonemaster::Engine::Normalization::Errors->new(DOMAIN_NAME_TOO_LONG => {});
         return \@messages, undef;
     }
 
--- a/lib/Zonemaster/Engine/Normalization.pm
+++ b/lib/Zonemaster/Engine/Normalization.pm
@@ -6,6 +6,8 @@ use utf8;
 use strict;
 use warnings;
 
+use parent 'Exporter';
+
 use Carp;
 use Encode;
 use Readonly;
@@ -14,6 +16,27 @@ use Zonemaster::LDNS;
 use Data::Dumper;
 
 use Zonemaster::Engine::Normalization::Errors;
+
+
+=head1 NAME
+
+Zonemaster::Engine::Normalization - utility functions for names normalization
+
+
+=head1 SYNOPSIS
+
+    use Zonemaster::Engine::Normalization;
+
+    my ($errors, $final_domain) = normalize_name($domain);
+
+=head1 EXPORTED FUNCTIONS
+
+=over
+=cut
+
+
+our @EXPORT      = qw[ normalize_name ];
+our @EXPORT_OK   = qw[ normalize_name normalize_label ];
 
 Readonly my $ASCII => qr/^[[:ascii:]]+$/;
 Readonly my $VALID_ASCII => qr(^[A-Za-z0-9/_-]+$);
@@ -56,6 +79,22 @@ Readonly my %AMBIGUOUS_CHARACTERS => (
     "LATIN CAPITAL LETTER I WITH DOT ABOVE" => q/\x{0130}/,
 );
 
+
+
+=item normalize_label($label)
+
+Normalize a single label from a DNS name.
+
+If the label is ASCII only, it is untouched, else it is converted according to IDNA2008.
+
+The downcasing, unicode normalization and conversion is performed by libidn2 using L<Zonemaster::LDNS/to_idn($name, ...)>.
+
+Returns a tuple C<($errors: ArrayRef[Zonemaster::Engine::Normalization::Errors], $alabel: String)>.
+
+In case of errors, the returned label will be undefined. If the method succeeded an empty error array is returned.
+
+=cut
+
 sub normalize_label {
     my ( $label ) = @_;
     my @messages;
@@ -87,6 +126,19 @@ sub normalize_label {
 
     return \@messages, $alabel;
 }
+
+=item normalize_name($name)
+
+Normalize a DNS name.
+
+
+The normalization process is detailed in the L<normalization document|https://github.com/zonemaster/zonemaster/blob/master/docs/specifications/tests/RequirementsAndNormalizationOfDomainNames.md>.
+
+Returns a tuple C<($errors: ArrayRef[Zonemaster::Engine::Normalization::Errors], $name: String)>.
+
+In case of errors, the returned name will be undefined. If the method succeeded an empty error array is returned.
+
+=cut
 
 sub normalize_name {
     my ( $uname ) = @_;
@@ -150,5 +202,9 @@ sub normalize_name {
 
     return \@messages, $final_name;
 }
+
+
+=back
+=cut
 
 1;
--- a/lib/Zonemaster/Engine/Normalization.pm
+++ b/lib/Zonemaster/Engine/Normalization.pm
@@ -52,6 +52,9 @@ Readonly my $WHITE_SPACES_RE => (sub {
     my $re = '[' . (join '', values %WHITE_SPACES) . ']';
     return qr/$re/;
 })->();
+Readonly my %AMBIGUOUS_CHARACTERS => (
+    "LATIN CAPITAL LETTER I WITH DOT ABOVE" => q/\x{0130}/,
+);
 
 sub normalize_label {
     my ( $label ) = @_;
@@ -62,14 +65,14 @@ sub normalize_label {
     if ( $label =~ $VALID_ASCII ) {
         $alabel = lc $label;
     } elsif ( $label =~ $ASCII ) {
-        push @messages, Zonemaster::Engine::Normalization::Errors->new('INVALID_ASCII' => {dlabel => $label});
+        push @messages, Zonemaster::Engine::Normalization::Errors->new('INVALID_ASCII' => {label => $label});
 
         return \@messages, undef;
     } elsif (Zonemaster::LDNS::has_idn) {
         try {
             $alabel = Zonemaster::LDNS::to_idn($label);
         } catch {
-            push @messages, Zonemaster::Engine::Normalization::Errors->new('INVALID_U_LABEL' => {dlabel => $label});
+            push @messages, Zonemaster::Engine::Normalization::Errors->new('INVALID_U_LABEL' => {label => $label});
 
             return \@messages, undef;
         }
@@ -78,7 +81,7 @@ sub normalize_label {
     }
 
     if ( length($alabel) > 63) {
-        push @messages, Zonemaster::Engine::Normalization::Errors->new('LABEL_TOO_LONG' => {dlabel => $label});
+        push @messages, Zonemaster::Engine::Normalization::Errors->new('LABEL_TOO_LONG' => {label => $label});
         return \@messages, undef;
     }
 
@@ -94,6 +97,17 @@ sub normalize_name {
 
     if (length($uname) == 0) {
         push @messages, Zonemaster::Engine::Normalization::Errors->new('EMPTY_DOMAIN_NAME');
+        return \@messages, undef;
+    }
+
+    foreach my $char_name (keys %AMBIGUOUS_CHARACTERS) {
+        my $char = $AMBIGUOUS_CHARACTERS{$char_name};
+        if ($uname =~ m/${char}/) {
+            push @messages, Zonemaster::Engine::Normalization::Errors->new(AMBIGUOUS_DOWNCASING => { unicode_name => $char_name });
+        }
+    }
+
+    if ( @messages ) {
         return \@messages, undef;
     }
 
--- a/lib/Zonemaster/Engine/Normalization.pm
+++ b/lib/Zonemaster/Engine/Normalization.pm
@@ -1,4 +1,4 @@
-package Zonemaster::Engine::Sanitization;
+package Zonemaster::Engine::Normalization;
 
 use 5.014002;
 
@@ -13,7 +13,7 @@ use Try::Tiny;
 use Zonemaster::LDNS;
 use Data::Dumper;
 
-use Zonemaster::Engine::Sanitization::Errors;
+use Zonemaster::Engine::Normalization::Errors;
 
 Readonly my $ASCII => qr/^[[:ascii:]]+$/;
 Readonly my $VALID_ASCII => qr/^[A-Za-z0-9\/\-_]+$/;
@@ -53,7 +53,7 @@ Readonly my $WHITE_SPACES_RE => (sub {
     return qr/$re/;
 })->();
 
-sub sanitize_label {
+sub normalize_label {
     my ( $label ) = @_;
     my @messages;
 
@@ -62,14 +62,14 @@ sub sanitize_label {
     if ( $label =~ $VALID_ASCII ) {
         $alabel = lc $label;
     } elsif ( $label =~ $ASCII ) {
-        push @messages, Zonemaster::Engine::Sanitization::Errors->new('INVALID_ASCII' => {dlabel => $label});
+        push @messages, Zonemaster::Engine::Normalization::Errors->new('INVALID_ASCII' => {dlabel => $label});
 
         return \@messages, undef;
     } elsif (Zonemaster::LDNS::has_idn) {
         try {
             $alabel = Zonemaster::LDNS::to_idn($label);
         } catch {
-            push @messages, Zonemaster::Engine::Sanitization::Errors->new('INVALID_U_LABEL' => {dlabel => $label});
+            push @messages, Zonemaster::Engine::Normalization::Errors->new('INVALID_U_LABEL' => {dlabel => $label});
 
             return \@messages, undef;
         }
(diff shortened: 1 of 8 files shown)


lib/Zonemaster/Engine/Sanitization.pm
```diff
@@ -0,0 +1,91 @@
...
+    return $alabel;
+}
+
+sub sanitize_name {
+    my ($name) = shift @_;
+
+    # TODO: Allow for different encoding?
+    my $uname = eval {
+        Encode::decode('UTF-8', $name, Encode::FB_CROAK);
+    };
```
mattias-p, line 52: I'm thinking this procedure should take decoded strings. After all it is a library function. Decoding from octet strings to native Perl encoding ought to be handled close to where the octet strings enter into the process. That way it's much easier to know what variables hold octet strings and which ones hold strings with Perl's native encoding. What do you think?

t/sanitization.t
```diff
@@ -0,0 +1,107 @@
+use Test::More;
```
mattias-p, line 1: We should `use utf8;` in all files to make sure Perl considers the source files to be UTF-8 encoded.

By the way I recommend using `perlcritic`. It'll probably tell you the same thing.

t/sanitization.t
```diff
@@ -0,0 +1,107 @@
...
+        '' => 'Zonemaster::Engine::Exception::DomainSanitization::EmptyDomainName',
+    );
+
+    while (($domain, $error) = each (%input_domains)) {
+        subtest "Domain: $domain ($error)" => sub {
+            throws_ok (sub {
+                Zonemaster::Engine::Sanitization::sanitize_name($domain);
+            }, $error, 'invalid domain should throw' );
+
+            print Dumper "$@", $@->{dlabel};
```
mattias-p, line 102: Have you considered using `diag` or `note` here instead of `print`?

lib/Zonemaster/Engine/Sanitization/Errors.pm
```diff
@@ -0,0 +1,211 @@
...
+
+use 5.014002;
+
+use strict;
+use warnings;
+
+use Class::Accessor "antlers";
+
+extends(qw/Zonemaster::Engine::Exception/);
+use Zonemaster::Engine::Exception;
```
mattias-p, line 11: Is it intentional to have `extends` first and `use` after? (Maybe that's how it's supposed to be, but it does look funny to me.)

lib/Zonemaster/Engine/Sanitization/Errors.pm
```diff
@@ -0,0 +1,211 @@
...
+extends(qw/Zonemaster::Engine::Exception::DomainSanitizationError/);
+
+has 'dlabel' => ( is => 'ro', isa => 'Str' );
+
+sub new {
+    my $proto = shift;
+    my $params = shift;
+
+    $params->{tag} = 'INVALID_ENCODING';
+    $params->{message} = 'Domain name has characters that can not be decoded';
```
mattias-p, line 110: `can not` -> `could not`?

Also, all the other messages end with a dot, but not this one.

lib/Zonemaster/Engine/Sanitization/Errors.pm
```diff
@@ -0,0 +1,211 @@
+package Zonemaster::Engine::Exception::DomainSanitizationError;
+
+use 5.014002;
+
+use strict;
+use warnings;
+
+use Class::Accessor "antlers";
```
mattias-p, line 8: Have you considered not using the antlers interface? I'm not sure what that would be like because I've never tried it but since this is new code I think it's appropriate to at least consider it.

mattias-p (commented):
In general this looks good to me but I have a bunch of comments.

lib/Zonemaster/Engine/Sanitization.pm
```diff
@@ -0,0 +1,91 @@
+package Zonemaster::Engine::Sanitization;
+
+use 5.014002;
+
+use strict;
+use warnings;
+
+use Encode;
+use Data::Dumper;
```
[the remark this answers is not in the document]
blacksponge: Fixed.

lib/Zonemaster/Engine/Sanitization.pm
```diff
@@ -0,0 +1,91 @@
...
+use Zonemaster::LDNS;
+
+use Zonemaster::Engine::Sanitization::Errors;
+
+my $VALID_ASCII = q/^[A-Za-z0-9\/\-_]+$/;
+my $FULL_STOP = q/\x{002E}/;
```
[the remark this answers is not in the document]
blacksponge: I did not, now it is used.

lib/Zonemaster/Engine/Sanitization.pm
```diff
@@ -0,0 +1,91 @@
...
+
+sub sanitize_label {
+    my ($label) = shift @_;
```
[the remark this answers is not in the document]
blacksponge: Fixed.

lib/Zonemaster/Engine/Sanitization.pm
```diff
@@ -0,0 +1,91 @@
...
+    my $alabel = eval {
+        Encode::encode('ascii', $label, Encode::FB_CROAK);
+    };
+
+    # Not ascii string, assume u-label
+    if ($@) {
+        $alabel = eval {
+            Zonemaster::LDNS::to_idn($label)
+        };
+        # TODO: handle when libidn not installed?
```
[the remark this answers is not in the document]
blacksponge: Yes I thought about that, I am not sure what would be the correct behaviour here.

lib/Zonemaster/Engine/Sanitization.pm
```diff
@@ -0,0 +1,91 @@
...
+
+use Zonemaster::Engine::Sanitization::Errors;
+
+my $VALID_ASCII = q/^[A-Za-z0-9\/\-_]+$/;
+my $FULL_STOP = q/\x{002E}/;
+
+sub sanitize_label {
+    my ($label) = shift @_;
+
+    my $alabel = eval {
```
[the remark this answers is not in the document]
blacksponge: No I did not, after looking into it I might keep it the way it is. What I would have needed is a `try / catch / else` statement but Try::Tiny does not seems to offer that.

lib/Zonemaster/Engine/Sanitization.pm
```diff
@@ -0,0 +1,91 @@
...
+    if ($@) {
+        die Zonemaster::Engine::Exception::DomainSanitization::InvalidEncoding->new();
+    }
+
+    if (length($uname) == 0) {
+        die Zonemaster::Engine::Exception::DomainSanitization::EmptyDomainName->new();
+    }
+
+    # Replace fullwidth full stop, ideographic full stop and halfwidth ideographic full stop with full stop
+    $uname =~ s/[\x{FF0E}\x{3002}\x{FF61}]/\x{002E}/g;
```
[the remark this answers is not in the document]
blacksponge: I actually did, but when I do, it replaces with the literal value `\x{002E}` and not the unicode character u+002e.

lib/Zonemaster/Engine/Sanitization.pm
blacksponge: Yes I think that in most cases the string passed to this function will be decoded. I originally this that because the spec states that `If the input *Domain Name* is not a non-empty string of [Unicode] characters this test case cannot be run.` so I wanted to make sure that it could be correctly decoded to a unicode string but this job can be delegated to the caller.

t/sanitization.t
blacksponge: > We should `use utf8;` in all files to make sure Perl considers the source files to be UTF-8 encoded.

Okay, I'll do that.

lib/Zonemaster/Engine/Sanitization/Errors.pm
blacksponge: It is not, fixed.

lib/Zonemaster/Engine/Sanitization/Errors.pm
blacksponge: I removed this error type as it is not useful any more if the strings are already decoded

t/sanitization.t
blacksponge: This was some debugging for myself, I removed it.

lib/Zonemaster/Engine/Sanitization/Errors.pm
blacksponge: I updated the module to not use the antlers interface.

lib/Zonemaster/Engine/Sanitization.pm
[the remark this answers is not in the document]
mattias-p: Hmm, interesting.

lib/Zonemaster/Engine/Sanitization.pm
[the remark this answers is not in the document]
mattias-p: Interesting. I wasn't aware of the `try / catch / else` construct. I see both Python and Ruby support it. But yeah, Try::Tiny only has the traditional `try / catch / finally`.

But in any case, I don't the need for fancy constructs here. To me something like this would be more straight forward. WDYT?

```perl
if ( $label =~ $VALID_ASCII ) {
    $alabel = lc $label;
}
elsif ( $label =~ $ASCII ) {
    die InvalidAscii->new();
}
elsif ( has_idn() ) {
    try {
        $alabel = to_idn( $label );
    }
    catch {
        die InvalidULabel->new();
    };
}
else {
    croak "non-ASCII not supported"
}
```

lib/Zonemaster/Engine/Sanitization.pm
mattias-p: Yeah, that's a bit weird. The precondition that the domain name is a Unicode string is already stated in the "Inputs" section. I don't see why we need to specify a runtime assertion to make sure the caller isn't breaking this (or any) part of the contract.

lib/Zonemaster/Engine/Sanitization.pm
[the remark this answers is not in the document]
blacksponge: Indeed this is a simpler way to do it, thanks.

lib/Zonemaster/Engine/Sanitization.pm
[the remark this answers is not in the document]
blacksponge: Okay, I reused the error message from the backend for consistency

lib/Zonemaster/Engine/Sanitization.pm
```diff
@@ -0,0 +1,82 @@
...
+
+    my $alabel = "";
+
+    if ( $label =~ $VALID_ASCII ) {
+        $alabel = lc $label;
+    } elsif ( $label =~ $ASCII ) {
+        die Zonemaster::Engine::Exception::DomainSanitization::InvalidAscii->new({ dlabel => $label });
+    } elsif (Zonemaster::LDNS::has_idn) {
+        try {
+            $alabel = Zonemaster::LDNS::to_idn($label);
```
matsduf, line 32: Before this, I expect to see down casing of all characters that can be down cased.

t/sanitization.t
```diff
@@ -0,0 +1,103 @@
...
+        # Domains with U-Labels
+        'café.example.com' => 'xn--caf-dma.example.com',
+        'エグザンプル｡example｡com' => 'xn--ickqs6k2dyb.example.com',
+        '🦈．example．com' => 'xn--7s9h.example.com',
+        'αβγδε.example.com' => 'xn--mxacdef.example.com',
+
+        # Domains with uppercase unicode
+        'CafÉ.example.com' => 'xn--caf-dma.example.com',
+        'ΑβΓΔε.example.com' => 'xn--mxacdef.example.com',
+        # TODO: special I case
```
matsduf, line 37: The special I case is only special in Turkish or Azerbaijan locale. Here we only do down casing.

Turkish and Azerbaijan locale:

- U+0049 LATIN CAPITAL LETTER I --> U+0131 LATIN SMALL LETTER DOTLESS I
- U+0130 LATIN CAPITAL LETTER I WITH DOT ABOVE --> U+0069 LATIN SMALL LETTER I

Other locale:

- U+0049 LATIN CAPITAL LETTER I --> U+0069 LATIN SMALL LETTER I
- U+0130 LATIN CAPITAL LETTER I WITH DOT ABOVE --> U+0069 LATIN SMALL LETTER I + U+0307 COMBINING DOT ABOVE

I am not sure if the last one (the combination of two code points) is legal in U-label. Maybe it is.

lib/Zonemaster/Engine/Sanitization.pm
blacksponge: It is already done in `to_idn`

t/sanitization.t
blacksponge: From what I read in the basic00 document, both u+0049 and u+00130 should be converted to u+0069, but from what I see when testing with libidn or firefox both convert u+0130 to u+0069 + u+0307 resulting in xn--i-9bb.

lib/Zonemaster/Engine/Sanitization.pm
```diff
@@ -0,0 +1,82 @@
...
+        } catch {
+            die Zonemaster::Engine::Exception::DomainSanitization::InvalidULabel->new({ dlabel => Encode::encode_utf8($label) });
+        }
+    } else {
+        croak 'The domain name contains non-ascii characters and IDNA is not installed';
```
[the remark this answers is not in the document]
blacksponge: done

lib/Zonemaster/Engine/Sanitization/Errors.pm
```diff
@@ -0,0 +1,174 @@
...
+
+__PACKAGE__->mk_ro_accessors(qw(dlabel));
+
+
+sub new {
+    my $proto = shift;
+    my $params = shift;
+
+    $params->{tag} = 'INVALID_ASCII';
+    $params->{message} = 'Domain name has an ASCII label with a character not permitted.';
```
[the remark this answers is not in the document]
blacksponge: done

t/sanitization.t
matsduf: U+0130 LATIN CAPITAL LETTER I WITH DOT ABOVE are only meaningful in a Turkish or Azerbaijan context (locale) and then it is downcased to U+0069 LATIN SMALL LETTER I (without the DOT-ABOVE). The sequence 0069+0307 is problematic.

matsduf: Down-casing LATIN CAPITAL LETTER I WITH DOT ABOVE to the sequence 0069+0307 does not make sense in a domain name context. In a normal text context it is different. The sequence 0069+0307 is a homograph of U+0069, which makes the sequence unsuitable for a domain name.

I see three ways of handling the code point:

1. Down-case to the sequence 0069+0307. That would most definitely create a label that does not match what is registered or used in DNS.
2. Down-case to the sequence U+0069. That will probably match the domain name in DNS, but will apparently mismatch with how how Firefox down-cases the code point.
3. Refuse to handle LATIN CAPITAL LETTER I WITH DOT ABOVE, and instead return an error message saying that there is no unambiguous down-casing of the code point.

Alternative 3 is probably the best way to go to create the least confusion.

The specification should be updated.

matsduf (changes requested):

t/sanitization.t
blacksponge: Will go for option 3 as discussed in meeting

blacksponge:
I have updated the implementation to match the changes in the specification. There also have been so design changes, mainly a shift from exception to returning errors. This was motivated to better match the spec and allow returning multiple errors (ex. one per invalid label as opposed to only one for the first invalid label).

matsduf (changes requested):
This looks good. Besides two small code comments, I do not see neither downcasing of non-ASCII nor conversion to NFC in the code. It might be included in `Zonemaster::LDNS::to_idn` since there are test domains with upper case non-ASCII, but then it should documented that that is the case.

In the test labels there is no example of NFC conversion. I could create one if you want.

blacksponge:
> This looks good. Besides two small code comments, I do not see neither downcasing of non-ASCII nor conversion to NFC in the code. It might be included in Zonemaster::LDNS::to_idn since there are test domains with upper case non-ASCII, but then it should documented that that is the case.

This is indeed done in `to_idn`, I will document it.

lib/Zonemaster/Engine/Normalization.pm
```diff
@@ -0,0 +1,154 @@
...
+use Encode;
+use Readonly;
+use Try::Tiny;
+use Zonemaster::LDNS;
+use Data::Dumper;
+
+use Zonemaster::Engine::Normalization::Errors;
+
+Readonly my $ASCII => qr/^[[:ascii:]]+$/;
+Readonly my $VALID_ASCII => qr/^[A-Za-z0-9\/\-_]+$/;
```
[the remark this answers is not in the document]
blacksponge: done

lib/Zonemaster/Engine/Normalization.pm
```diff
@@ -0,0 +1,154 @@
...
+        }
+    }
+
+    if ( @messages ) {
+        return \@messages, undef;
+    }
+
+    $uname =~ s/${FULL_STOPS_RE}/${ASCII_FULL_STOP}/g;
+
+    if ( $uname eq '.' ) {
```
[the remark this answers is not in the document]
blacksponge: done

blacksponge:
> In the test labels there is no example of NFC conversion. I could create one if you want.

That would be great if you can.

matsduf:
> > In the test labels there is no example of NFC conversion. I could create one if you want.
> 
> That would be great if you can.

Left side is non-NFC, right side is the equivalent NFC.

"d\x{006F}\x{0308}d" => 'död',
"aq\x{0307}\x{0323}a" => "aq\x{0323}\x{0307}a",
"aḋ\x{0323}a" => "aḍ\x{0307}a"

lib/Zonemaster/Engine/Normalization.pm
```diff
@@ -0,0 +1,210 @@
...
+    "LATIN CAPITAL LETTER I WITH DOT ABOVE" => q/\x{0130}/,
+);
+
+
+
+=item normalize_label($label)
+
+Normalize a single label from a DNS name.
+
+If the label is ASCII only, it is untouched, else it is converted according to IDNA2008.
```
matsduf, line 88: It is not untouched, it is down cased.

lib/Zonemaster/Engine/Normalization.pm
```diff
@@ -0,0 +1,210 @@
...
+
+The downcasing, unicode normalization and conversion is performed by libidn2 using L<Zonemaster::LDNS/to_idn($name, ...)>.
```
matsduf, line 90: "Downcasing of upper case non-ASCII characters, normalization to the Unicode NFC format and conversion from U-label to A-label is performed by (...)"

Please no long lines. It makes review in Github harder.

lib/Zonemaster/Engine/Normalization/Errors.pm
```diff
@@ -0,0 +1,148 @@
...
+        }
+    }
+
+    return bless $obj, $class;
+}
+
+
+=item message
+
+Returns the transated error message using the parameters given when creating the object.
```
matsduf, line 108: "transated" -> "translated"

lib/Zonemaster/Engine/Normalization.pm
```diff
@@ -0,0 +1,210 @@
...
+})->();
+Readonly my %AMBIGUOUS_CHARACTERS => (
+    "LATIN CAPITAL LETTER I WITH DOT ABOVE" => q/\x{0130}/,
+);
+
+
+
+=item normalize_label($label)
+
+Normalize a single label from a DNS name.
```
matsduf, line 86: I suggest "domain name" instead of "DNS name". It fits our terminology better.

lib/Zonemaster/Engine/Normalization.pm
```diff
@@ -0,0 +1,210 @@
...
+        push @messages, Zonemaster::Engine::Normalization::Errors->new(LABEL_TOO_LONG => {label => $label});
+        return \@messages, undef;
+    }
+
+    return \@messages, $alabel;
+}
+
+=item normalize_name($name)
+
+Normalize a DNS name.
```
matsduf, line 132: "domain name"?

lib/Zonemaster/Engine/Normalization.pm
blacksponge: fixed

lib/Zonemaster/Engine/Normalization.pm
blacksponge: done

lib/Zonemaster/Engine/Normalization/Errors.pm
blacksponge: fixed

lib/Zonemaster/Engine/Normalization.pm
blacksponge: yes, it is better indeed

lib/Zonemaster/Engine/Normalization.pm
blacksponge: done

blacksponge:
I added the NFC conversion tests and updated the documentation.

matsduf (approved):
This looks fine. I have not run any tests on it though.

tgreenx (commented):
LGTM, tested and it works as expected.
One thing though, overall I think the code could benefit from an editorial pass on the spacing in every conditional blocks (e.g. `if ( ... )` everywhere) to improve readability but I won't be blocking the PR if it is not done.

lib/Zonemaster/Engine/Normalization.pm
```diff
@@ -0,0 +1,215 @@
...
+use warnings;
+
+use parent 'Exporter';
+
+use Carp;
+use Encode;
+use Readonly;
+use Try::Tiny;
+use Zonemaster::LDNS;
+use Data::Dumper;
```
pnax, line 16: Might be safely removed

pnax (dismissed):
lgtm

lib/Zonemaster/Engine/Normalization/Error.pm
```diff
@@ -0,0 +1,147 @@
...
+
+sub tag {
+    my ( $self ) = @_;
+
+    return $self->{tag};
+}
+
+=item string
+
+Returns a string representation of the error object, equivalent to message.
```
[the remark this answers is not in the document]
blacksponge: fixed

lib/Zonemaster/Engine/Normalization.pm
blacksponge: fixed

matsduf (dismissed):

lib/Zonemaster/Engine/Normalization/Error.pm
```diff
@@ -129,7 +129,7 @@ sub tag {
 
 =item string
 
-Returns a string representation of the error object, equivalent to message.
+Returns a string representation of the error object. Euivalent to message().
```
tgreenx, line 132: Euivalent -> Equivalent

tgreenx (commented):
One typo else ready to approve

lib/Zonemaster/Engine/Normalization/Error.pm
blacksponge: Fixed

tgreenx (approved):

matsduf (approved):

matsduf:
Updated by #1157

PR #566, opened 2019-06-20
Just making some benchmarks

mattias-p:

PR #1042, opened 2022-03-03
replace moose with class accessor in Logger::Entry

blacksponge:
## Purpose

Remove Moose from Logger::Entry for better performance. The logger entries are objects that are instantiated a large number of time and which properties are accessed a lot during tests execution. In large batches having too much overhead in the methods involved can lead to performance issue. (I don't have the exact numbers anymore, those are things I tested last September but didn't get the chance to backport in mainline Zonemaster).

## Context

* Continuation of a work started in https://github.com/zonemaster/zonemaster-engine/pull/566
* Partially ported from https://gitlab.rd.nic.fr/zonemaster/zonemaster-engine/-/blob/nightly-ronde/lib/Zonemaster/Engine/Logger/Entry.pm

## Changes

Replace Moose with Class::Accessor in Logger::Entry

## How to test this PR

* Tests can be executed as usual:  `zonemaster-cli --level info zonemaster.net --show-module --show-testcase`
* JSON generation is not impacted: `zonemaster-cli --level info zonemaster.net --json`

### DIFF
--- a/lib/Zonemaster/Engine/Logger/Entry.pm
+++ b/lib/Zonemaster/Engine/Logger/Entry.pm
@@ -7,11 +7,14 @@ use warnings;
 
 use version; our $VERSION = version->declare("v1.1.8");
 
+use Carp qw( confess );
 use Time::HiRes qw[time];
 use JSON::PP;
-use Moose;
+use Class::Accessor;
 use Zonemaster::Engine;
 use Zonemaster::Engine::Profile;
+
+use base qw(Class::Accessor);
 
 use overload '""' => \&string;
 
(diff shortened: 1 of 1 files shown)


lib/Zonemaster/Engine/Logger/Entry.pm
```diff
@@ -100,6 +150,13 @@ sub _build_level {
     }
 }
 
+sub _set_level {
+    my ( $self, $level ) = @_;
+
+    $self->{_level} = $level
+}
```
mattias-p, line 157: Is this ever used? The old declaration looked a bit funny being declared as readonly but with a writer specified.

mattias-p (approved):
LGTM. But I have a question.

lib/Zonemaster/Engine/Logger/Entry.pm
blacksponge: Yes I found that odd too. It looks like it is only used in the `Logger::_check_filter` method.

https://github.com/zonemaster/zonemaster-engine/blob/087066d46a75a03d86fdd3f356288c8557be5a7a/lib/Zonemaster/Engine/Logger.pm#L51-L86

But it seems that the `logfilter` entry is no longer present in the profile. Has this feature been dropped? Is it documented somewhere? If we don't use that feature any more we may remove this method as well.

blacksponge: Okay, I found it in the profile_example.json, so I guess it is still used.

pnax: The `logfilter` entry is documented here:
https://metacpan.org/pod/Zonemaster::Engine::Profile#logfilter

matsduf:
@blacksponge, I am not a Moose fan, but if we replace Moose here with Class::Accessor should we then replace it with Class::Accessor everywhere? What do we win and what do we loose?

blacksponge:
> What do we win and what do we loose?

We remove some complex abstraction and it might improve performance in some case but we lose the abstraction / type checking and need to do that by hand if we still want it, meaning more boilerplate code.  
Maybe moving away from Moose is a good idea, but I don't think it need to be a priority for the modules for which we will not see to substantial gain.

matsduf (approved):
I am fine with the change. I approve.

PR #1043, opened 2022-03-03
Use ns cache early

blacksponge:
## Purpose

Improve overall performance by returning the Nameserver object from local cache as early as possible. The main purpose is to avoid creating new instances of Net::IP object when it is not needed. I have recently profiled Zonemaster running a batch of a ~100 domains. It turns out that most time was not spent on querying but on other operations. Creating countless Net::IP objects was one of the heaviest ones, and that can easily be avoided for the most part.

Here are some profiling info for one test as measured by `PERL5OPT=-d:NYTProf ./zonemaster-cli --level info zonemaster.net --show-module --show-testcase`

<table>
    <thead>
        <tr>
            <th></th>
            <th colspan="3">Net::IP</th>
            <th></th>
            <th colspan="3">Net::IP::XS</th>
        </tr>
        <tr>
            <th></th>
            <th>calls</th>
            <th>exclusive time</th>
            <th>inclusive time</th>
            <th></th>
            <th>calls</th>
            <th>exclusive time</th>
            <th>inclusive time</th>
        </tr>
    </thead>
    <tbody>
        <tr>
            <td></td>
            <td colspan="3">Net::IP::new</td>
            <td></td>
            <td colspan="3">Net::IP::XS::new (xsub)</td>
        </tr>
        <tr>
            <th>before</th>
            <td>9640</td>
            <td>48.5ms</td>
            <td>3090ms </td>
            <th></th>
            <td>9640</td>
            <td>51.9ms</td>
            <td>51.9ms</td>
        </tr>
        <tr>
            <th>after</th>
            <td>382</td>
            <td>1.91ms</td>
            <td>181ms</td>
            <th></th>
            <td>382</td>
            <td>3.99ms</td>
            <td>3.99ms</td>
        </tr>
        <tr>
            <td></td>
            <td colspan="3">Zonemaster::Engine::Nameserver::new</td>
            <td></td>
            <td colspan="3">Zonemaster::Engine::Nameserver::new</td>
        </tr>
        <tr>
            <th>before</th>
            <td>9467</td>
            <td>302ms</td>
            <td>3640ms</td>
            <th></th>
            <td>9467</td>
            <td>337ms</td>
            <td>776ms</td>
        </tr>
        <tr>
            <th>after</th>
            <td>9467</td>
            <td>97.6ms</td>
            <td>378ms</td>
            <th></th>
            <td>9467</td>
            <td>118ms</td>
            <td>316ms</td>
        </tr>
    </tbody>
</table>

We can also note the gain from just using Net::IP::XS instead of the pure perl implementation.

## Context

Ported from https://gitlab.rd.nic.fr/zonemaster/zonemaster-engine/-/blob/nightly-ronde/lib/Zonemaster/Engine/Nameserver.pm

## Changes

Return Zonemaster::Backend::Nameserver object from cache as early as possible.

## How to test this PR

Tests should work as usual: `./zonemaster-cli --level info zonemaster.net --show-module --show-testcase`

### DIFF
--- a/lib/Zonemaster/Engine/Nameserver.pm
+++ b/lib/Zonemaster/Engine/Nameserver.pm
@@ -49,6 +49,8 @@ has 'blacklisted' => ( is => 'rw' );
 ###
 
 our %object_cache;
+our %address_object_cache;
+our %address_repr_cache;
 
 ###
 ### Build methods for attributes
@@ -67,7 +69,6 @@ sub new {
     # Required arguments
     confess "Attribute \(address\) is required"
       if !exists $attrs->{address};
-    my $address = $attrs->{address};
 
     # Type coercions
     $attrs->{name} = Zonemaster::Engine::DNSName->from_string( $attrs->{name} )
@@ -76,18 +77,25 @@ sub new {
     my $name = lc( q{} . $attrs->{name} );
     $name = '$$$NONAME' unless $name;
 
-    if (blessed $address && $address->isa( 'Zonemaster::Engine::Net::IP' )) {
-        $address = $address->short;
+    my $address;
+
+    # Use a object cache for IP type coercion (don't parse IP unless it is needed)
+    if (!blessed $attrs->{address} || !$attrs->{address}->isa( 'Zonemaster::Engine::Net::IP' )) {
+        if (!exists $address_object_cache{$attrs->{address}}) {
+            $address_object_cache{$attrs->{address}} = Zonemaster::Engine::Net::IP->new($attrs->{address});
+            $address_repr_cache{$attrs->{address}} = $address_object_cache{$attrs->{address}}->ip;
+        }
+        # Fetch IP object from the address cache (avoid object creation and method call)
+        $address = $address_repr_cache{$attrs->{address}};
+        $attrs->{address} = $address_object_cache{$attrs->{address}};
+    } else {
+        $address = $attrs->{address}->ip;
     }
 
+    # Return Nameserver object as soon as possible
     if ( exists $object_cache{$name}{$address} ) {
         return $object_cache{$name}{$address};
     }
-
-    # Remaining type coercions (don't parse IP unless it is needed)
-    $attrs->{address} = Zonemaster::Engine::Net::IP->new( $attrs->{address} )
-      if exists $attrs->{address}
-      && ( !blessed $attrs->{address} || !$attrs->{address}->isa( 'Zonemaster::Engine::Net::IP' ) );
 
     # Type constraints
     confess "Argument must be coercible into a Zonemaster::Engine::DNSName: name"
@@ -678,6 +686,8 @@ sub axfr {
 
 sub empty_cache {
     %object_cache = ();
+    %address_object_cache = ();
+    %address_repr_cache = ();
 
     Zonemaster::Engine::Nameserver::Cache::empty_cache();
 
--- a/lib/Zonemaster/Engine/Nameserver.pm
+++ b/lib/Zonemaster/Engine/Nameserver.pm
@@ -67,10 +67,24 @@ sub new {
     # Required arguments
     confess "Attribute \(address\) is required"
       if !exists $attrs->{address};
+    my $address = $attrs->{address};
 
     # Type coercions
     $attrs->{name} = Zonemaster::Engine::DNSName->from_string( $attrs->{name} )
       if !blessed $attrs->{name} || !$attrs->{name}->isa( 'Zonemaster::Engine::DNSName' );
+
+    my $name = lc( q{} . $attrs->{name} );
+    $name = '$$$NONAME' unless $name;
+
+    if (blessed $address && $address->isa( 'Zonemaster::Engine::Net::IP' )) {
+        $address = $address->short;
+    }
+
+    if ( exists $object_cache{$name}{$address} ) {
+        return $object_cache{$name}{$address};
+    }
+
+    # Remaining type coercions (don't parse IP unless it is needed)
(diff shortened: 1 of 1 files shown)


mattias-p:
I'm afraid this change breaks the caching of equivalent addresses that are formatted differently, and I'm pretty sure we don't want that. Maybe there's a way to normalize IP addresses that is faster than running them through Net::IP/Net::IP::XS?

And maybe we could drop support for Net::IP and only rely on Net::IP::XS?

mattias-p:
Maybe we could check the addresses against a regexp and only run them through Zonemaster::Net::IP they don't match?

mattias-p:
Nice benchmark BTW. And nice numbers.

blacksponge:
> I'm afraid this change breaks the caching of equivalent addresses that are formatted differently, and I'm pretty sure we don't want that.

While it is true this way the NS cache will have duplicated entries I don't see that as a big issue. I think that the case of an address provided to the NS object with different formats is pretty rare. Even in this case most of the time spent in Nameserver::new is spent in parsing IPs so in case of similar addresses creating a brand new Nameserver object will come at a smaller cost than checking for an existing object, and will be reused for the subsequent queries with same address format. In the end I think that having some duplicate entries in the NS cache does not outbalance the benefit of droping the IP parsing. Also it may be worth saying that this only affect the NS object cache and not in any case the query/response cache (Nameserver::Cache) that still rely on normalized addresses.

mattias-p:
The query/response cache respecting equality makes me feel a lot better about this. But I'm still a bit worried that we'll get confused when debugging by there being multiple entries in the cache for the same name server. If we could avoid doppelgängers without a significant performance hit, that would be really nice.

matsduf:
> The query/response cache respecting equality makes me feel a lot better about this. But I'm still a bit worried that we'll get confused when debugging by there being multiple entries in the cache for the same name server. If we could avoid doppelgängers without a significant performance hit, that would be really nice.

@mattias-p, can you describe a scenario when we could get a "doppelgänger"?

matsduf:
@blacksponge, what is the difference between "exclusive time" and "inclusive time" in your table?

lib/Zonemaster/Engine/Nameserver.pm
```diff
@@ -67,10 +67,24 @@ sub new {
...
     confess "Attribute \(address\) is required"
       if !exists $attrs->{address};
+    my $address = $attrs->{address};
 
     # Type coercions
     $attrs->{name} = Zonemaster::Engine::DNSName->from_string( $attrs->{name} )
       if !blessed $attrs->{name} || !$attrs->{name}->isa( 'Zonemaster::Engine::DNSName' );
+
+    my $name = lc( q{} . $attrs->{name} );
+    $name = '$$$NONAME' unless $name;
```
matsduf, line 77: What is the purpose of "$$$NONAME" string? And the choice of it?

blacksponge: I guess it is for Nameservers instantiated without a name, only an address. I did not choose this string, it was already there, https://github.com/zonemaster/zonemaster-engine/pull/1043/files#diff-7f293915c98c514d9ef944a59d0dab6a10d83ed2fbaf5de298d12a691a0a9409L118 and has been there since the start it seems.

matsduf:
If I understand it correctly, the "doppelgänger" issue will only come when IP addresses has been provided as input to the program, not when IP addresses have been fetched from a DNS response. Is that correct? -- I would also guess that the risk of getting IPv4 addresses in various format is very small.

What would the cost be of expanding all IPv6 addresses to explicit format? It should be possible to do that without Net::IP is that is too expensive.

lib/Zonemaster/Engine/Nameserver.pm
matsduf: It is still mysterious, but for a different PR.

blacksponge:
> what is the difference between "exclusive time" and "inclusive time" in your table?

Exclusive time is only does not include time spend in call to other function but inclusive time does.

matsduf:
Caching name server IP address by name over several tests is only possible if the same IP address is used for all tests. In the simple case of multiple tests of type "delegated" where none of the tested domains have any glue records in the delegation.

* If multiple tests are created in a batch then only one set of "undelegated" NS and glue records can be provided. However, if separate tests are created then the same name server name could be given different IP address values. Is the caching for batches only?
* If the delegation of a domain name has glue records it might be the case that the glue records have different IP address value from the authoritative record. How is that handled by the cache?
* There could be multiple domain names with the same set of name server names, where some have glue records for some or all name server names, and others do not have any glue records. Any inconsistency between glue and authoritative data is correctly reported where glue is provided, but not where no glue is included in the delegation.

matsduf:
@blacksponge, will this caching affect batches only?

blacksponge:
> Caching name server IP address by name over several tests is only possible if the same IP address is used for all tests. 

Nameserver are cached by name and IP and I did not changed that. I cache IP object by string representation and this can not change.

> If the delegation of a domain name has glue records it might be the case that the glue records have different IP address value from the authoritative record. How is that handled by the cache?

The cache was there before, I don't change any preexisting behavior here.

> If multiple tests are created in a batch then only one set of "undelegated" NS and glue records can be provided. However, if separate tests are created then the same name server name could be given different IP address values. Is the caching for batches only?

Same comment, and no the cache is not used for batch only. It might be worth mentioning that the way tests are executed by the backend makes this cache local to a single test, there is no sharing of nameservers objects between tests.

matsduf (approved):
I approve.

mattias-p (approved):
Looking good!

PR #1044, opened 2022-03-03
cache test level config in logger::entry

blacksponge:
## Purpose

Reduce the amount of time used cloning the test levels config.  
Each call to `Zonemaster::Engine::Profile->effective->get` clone the result if it is not a scalar. For the test levels, this means that there will be a lot of calls to clone a fairly big hash. Here are the profiling info for a single test measured with `PERL5OPT=-d:NYTProf ./zonemaster-cli --level info zonemaster.net --show-module --show-testcase`.
<table>
    <thead>
        <tr>
            <th></th>
            <th>calls</th>
            <th>exclusive time</th>
            <th>inclusive time</th>
        </tr>
    </thead>
    <tbody>
        <tr>
            <td></td>
            <td colspan="3">Zonemaster::Engine::Logger::Entry::level</td>
        </tr>
        <tr>
            <th>before</th>
            <td>5841</td>
            <td>411ms</td>
            <td>3390ms</td>
        </tr>
        <tr>
            <th>after</th>
            <td>5842</td>
            <td>102ms</td>
            <td>122ms</td>
        </tr>
        <tr>
            <td></td>
            <td colspan="3">Zonemaster::Engine::Profile::get</td>
        </tr>
        <tr>
            <th>before</th>
            <td>35873</td>
            <td>361ms</td>
            <td>3570ms</td>
        </tr>
        <tr>
            <th>after</th>
            <td>24838</td>
            <td>245ms</td>
            <td>726ms</td>
        </tr>
    </tbody>
</table>

## Context

Ported from https://gitlab.rd.nic.fr/zonemaster/zonemaster-engine/-/blob/nightly-ronde/lib/Zonemaster/Engine/Logger/Entry.pm

## Changes

* Cache the test levels for the time of the test

## How to test this PR

Testing a domain should work as usual:  `./zonemaster-cli --level info zonemaster.net --show-module --show-testcase`

### DIFF
--- a/lib/Zonemaster/Engine.pm
+++ b/lib/Zonemaster/Engine.pm
@@ -171,7 +171,7 @@ sub can_continue {
     my ( $class ) = @_;
 
     return 1;
-    
+
 }
 
 sub save_cache {
@@ -203,6 +203,7 @@ sub start_time_now {
 
 sub reset {
     Zonemaster::Engine::Logger->start_time_now();
+    Zonemaster::Engine::Logger->reset_config();
     Zonemaster::Engine::Nameserver->empty_cache();
     $logger->clear_history() if $logger;
     Zonemaster::Engine::Recursor->clear_cache();
--- a/lib/Zonemaster/Engine/Logger.pm
+++ b/lib/Zonemaster/Engine/Logger.pm
@@ -90,6 +90,11 @@ sub start_time_now {
     return;
 }
 
+sub reset_config {
+    Zonemaster::Engine::Logger::Entry->reset_config();
+    return;
+}
+
 sub clear_history {
     my ( $self ) = @_;
 
@@ -210,6 +215,10 @@ Set the logger's start time to the current time.
 =item clear_history()
 
 Remove all known log entries.
+
+=item reset_config()
+
+Clear the test level cached configuration.
 
 =back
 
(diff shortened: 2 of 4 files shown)


pnax (dismissed):
Nice improvement. Just a thought, do we really need to clone the data even once? Can't we just pass a ref around?

matsduf (approved):

mattias-p:
> Just a thought, do we really need to clone the data even once? Can't we just pass a ref around?

I designed the Profile module interface to avoid leaking references to internal mutable data structures. The idea is that it'll make our lives easier.

mattias-p:
Won't this break the ability to switch between profiles in Backend? Maybe reset_config() could be called from the "run" methods in Zonemaster::Engine::Test or something?

matsduf:
> Won't this break the ability to switch between profiles in Backend? 

@blacksponge, has switching been tested?

blacksponge:
> Won't this break the ability to switch between profiles in Backend? Maybe reset_config() could be called from the "run" methods in Zonemaster::Engine::Test or something?

As the profile is only set in the child process this is not breaking profile switching. To be sure I tried to run two consecutive tests with different profile and it was correctly executed.

matsduf:
I will re-review when the conflict has been resolved.

blacksponge:
I have rebased on top of develop, please re-review

matsduf (approved):
I approve.

PR #1047, opened 2022-04-04
Corrects argument name in Delegation.pm

matsduf:
## Purpose

The default message string for SOA_NOT_EXISTS expects argument "ns":
```
  15.12 ERROR     DELEGATION06   Empty NOERROR response to SOA query was received from {ns}.
```
but the tag is outputted with argument "nsname":
```
   1.00 ERROR     DELEGATION06   SOA_NOT_EXISTS   nsname=ns.ripe.net/2001:67c:e0::6
```

The correct argument name is "ns" (see the specification in "[Arguments for test case messages]").

## Changes

Argument name is updated in Delegation.pm. Nothing else is directly affected, but all PO files must be
updated with correct argument name.

## How to test this PR

Find a resolver that will accept a SOA query for iis.se and then the same query with "+norec" set to dig.
REPLACE RESOLVER-IP with the IP of that resolver and then verify that the message outputted with
the name server named and IP, and not "{ns}".

```
zonemaster-cli --ns nsp.dnsnode.net --ns resolver.namn.se/RESOLVER-IP iis.se --show-testcase --test delegation/delegation06 --locale en_US.UTF-8
```

[Arguments for test case messages]: https://github.com/zonemaster/zonemaster-engine/blob/develop/docs/logentry_args.md

### DIFF
--- a/lib/Zonemaster/Engine/Test/Delegation.pm
+++ b/lib/Zonemaster/Engine/Test/Delegation.pm
@@ -755,7 +755,7 @@ sub delegation06 {
         my $p = $local_ns->query( $zone->name, $query_type );
         if ( $p and $p->rcode eq q{NOERROR} ) {
             if ( not $p->get_records( $query_type, q{answer} ) ) {
-                push @results, info( SOA_NOT_EXISTS => { nsname => $local_ns->string } );
+                push @results, info( SOA_NOT_EXISTS => { ns => $local_ns->string } );
             }
         }
 


mattias-p:
This was made against the wrong branch, right?

matsduf:
> This was made against the wrong branch, right?

@mattias-p, yes, and I have already corrected that.

tgreenx (approved):

blacksponge (approved):

mattias-p (approved):

Issue #1048, opened 2022-04-07
Fatal error in Zonemaster::Engine::Test::DNSSEC


anandb-ripencc:
I just tested a domain using the Zonemaster instance at `zonemaster.net` and it emitted a CRITICAL message, showing a perl error from the engine. The test is:

https://zonemaster.net/result/f564ffda11c336d4

I don't know what the issue is, but shouldn't the engine fail more gracefully?

matsduf:
```
zonemaster-cli --test DNSSEC/dnssec02 --show-testcase 249.200.193.in-addr.arpa --raw --level INFO
   0.00 INFO      UNSPECIFIED    GLOBAL_VERSION   version=v4.4.0
   5.43 CRITICAL  UNSPECIFIED    MODULE_ERROR   module=DNSSEC; msg=Can't call method "hexdigest" on an undefined value at /usr/local/share/perl/5.26.1/Zonemaster/Engine/Test/DNSSEC.pm line 1504.

```
This is a bug in Zonemaster-Engine for test case DNSSEC02. I notice that that the zone has one DS record created by Algorithm GOST. We will investigate.

Thank you for reporting.


```
zonemaster-cli --test DNSSEC/dnssec01 --show-testcase 249.200.193.in-addr.arpa --raw --level INFO
   0.00 INFO      UNSPECIFIED    GLOBAL_VERSION   version=v4.4.0
   4.13 WARNING   DNSSEC01       DS_ALGO_SHA1_DEPRECATED   algo_mnemo=SHA-1; algo_num=1; domain=249.200.193.in-addr.arpa; keytag=13114; ns=ns3.afrinic.net/2001:500:14:6100:ad::1
   4.13 INFO      DNSSEC01       DS_ALGORITHM_OK   algo_mnemo=SHA-384; algo_num=4; domain=249.200.193.in-addr.arpa; keytag=13114; ns=ns3.afrinic.net/2001:500:14:6100:ad::1
   4.13 ERROR     DNSSEC01       DS_ALGORITHM_DEPRECATED   algo_mnemo=GOST R 34.11-94; algo_num=3; domain=249.200.193.in-addr.arpa; keytag=13114; ns=ns3.afrinic.net/2001:500:14:6100:ad::1
[...]

PR #1051, opened 2022-04-14
Implementation update for DNSSEC02

### LINKED ISSUE #1049: Update DNSSEC02
# Summary

This is for the implementation of the updated [DNSSEC02] specification (updated by zonemaster/zonemaster#1056). The specification has been updated in a number of respects:

* Ignore unimplemented digest algorithms. The code should not crash and the there should be no message tag outputted if the digest algorithm cannot be handled by the Zonemaster implementation. That applies for any digest algorithm code that a DS record might have.
* The default level of two messages has been lowered from ERROR to WARNING.

The overall logic has not been changed.

# Handling of unimplemented digest algorithm

The bug resported in #1048 is due to the incorrect handling of unknown (unimplemented) digest algorithms.

The following code (https://github.com/zonemaster/zonemaster-engine/blob/develop/lib/Zonemaster/Engine/Test/DNSSEC.pm#L194-L199) assumes what digest algorithms are supported, but that depends on the underlying OpenSSL.
```
Readonly::Hash our %LDNS_digest_algorithms_supported => (
    1 => q{sha1},
    2 => q{sha256},
    3 => q{gost},
    4 => q{sha384},
);
```

In the following loop (https://github.com/zonemaster/zonemaster-engine/blob/develop/lib/Zonemaster/Engine/Test/DNSSEC.pm#L1501-L1510) `$tmp_ds` will be undefind if the digest algorithm is not supported, and then `$tmp_ds->hexdigest()` will crash.

```

                foreach my $matching_keytag_dnskey ( @matching_keytag_dnskeys ) {
                    if ( exists $LDNS_digest_algorithms_supported{$ds->digtype()} ) {
                        my $tmp_ds = $matching_keytag_dnskey->ds($LDNS_digest_algorithms_supported{$ds->digtype()});
                        if ( $tmp_ds->hexdigest() eq $ds->hexdigest() ) {
                            $matching_dnskey = $matching_keytag_dnskey;
                            $match_ds_dnskey = 1;
                            last;
                        }
                    }
                }
```

Update the code so that the test (compare if the digest in the DS is correct) is ignored if the digest algorithm is not supported. No message should be outputted, at least nothing INFO or higher. DEBUG messages could be outputted.

The output of the test case should be the same as if the DS record could be verified against the DNSKEY.

[DNSSEC02]: https://github.com/zonemaster/zonemaster/blob/develop/docs/specifications/tests/DNSSEC-TP/dnssec02.md

tgreenx:
## Purpose

The purpose of this PR is three-fold, according to [#1056](https://github.com/zonemaster/zonemaster/pull/1056):
 - Fixes handling of unimplemented digest algorithm.
 - Changes default level of two messages from ERROR to WARNING.
 - Adds checks for DNS responses.

## Context

Fixes #1049
Fixes [#1036](https://github.com/zonemaster/zonemaster/issues/1036)

## How to test this PR

```
zonemaster-cli --test DNSSEC/dnssec02 --show-testcase 249.200.193.in-addr.arpa --raw --level INFO
   0.00 INFO      UNSPECIFIED    GLOBAL_VERSION   version=v4.4.0
   3.96 ERROR     DNSSEC02       DS02_NO_MATCH_DS_DNSKEY   keytag=13114; ns_ip_list=94.228.98.98;94.228.99.99
```


matsduf:
> ## How to test this PR
> 
> ```
> zonemaster-cli --test DNSSEC/dnssec02 --show-testcase 249.200.193.in-addr.arpa --raw --level INFO
>    0.00 INFO      UNSPECIFIED    GLOBAL_VERSION   version=v4.4.0
>    3.96 ERROR     DNSSEC02       DS02_NO_MATCH_DS_DNSKEY   keytag=13114; ns_ip_list=94.228.98.98;94.228.99.99
> ```

There is also recorded data in #1048 to be used if the problematic DS record has been removed from the parent zone.

But the result here does not follow the specification. The problematic DS recod should just be ignored.

matsduf (changes requested):
A DS record that is used with a digest algorithm that Zonemaster does not have any support for should be ignored.

tgreenx:
Changes addressed in latest commit, please re-review @matsduf.

`zonemaster-cli --test DNSSEC/dnssec02 --show-testcase 249.200.193.in-addr.arpa --raw --level INFO
   0.00 INFO      UNSPECIFIED    GLOBAL_VERSION   version=v4.4.0`

matsduf:
I am sorry, my comments were unclear. The DS record should be ignored in respect to its digest matching the DNSKEY. The test case should still verify the DNSKEY that the DS record points at.

In step 5.3 we check if Zonemaster can handle the digest algorithm. If not (5.3.1) we skip verification. If we do (5.3.2 we will take note if the digest check fails. In both cases we continue with steps 5.4 to 5.9.

There is no reason to handle digest algorithms 1-4 special. The specification says that we should handle all digest algorithms the same.

```
    5. For each DS in *DS Records*, do:
       1. Find the equivalent DNSKEY in *DNSKEY RRs* by key ID (key tag). If
          there is more than one such DNSKEY, select the correct one.
       2. If matching DNSKEY is not found add DS key tag and name server IP to
          the *No DNSKEY for DS* set and go to next DS.
       3. Verify if the Zonemaster installation has support for the digest
          algorithm that created the DS:
          1. If no support, then ignore the following test if the DS matches
             the DNSKEY.
          2. Else, if the DS values (algorithm and digest) do not match the
             DNSKEY record then add DS key tag and name server IP to the
             *No match DS DNSKEY* set.
       4. If bit 7 of the DNSKEY flags field is unset (value 0), then add DS key
          tag and name server IP to the *DNSKEY not for zone signing* set.
       5. If bit 15 of the DNSKEY flags field is unset (value 0), then add the
          DNSKEY record key tag and name server IP to the *DNSKEY not SEP*
          set.
       6. Find the equivalent RRSIG in *DNSKEY RRSIG* by key ID (key tag). If
          there is more than one such RRSIG record, select the correct one.
       7. If matching RRSIG is not found, add DNSKEY record key tag and name
          server IP to the *No matching DNSKEY RRSIG* set.
       8. Else, if the Zonemaster installation does not have support for the
          DNSKEY algorithm that created the RRSIG, then add name server IP,
          DNSKEY algorithm and DNSKEY key tag to the *Algo Not Supported By ZM*
          set.
       9. Else, if the RRSIG values (algorithm and signature) do not match
          the DNSKEY then add the key tag from the RRSIG record and name server
          IP to the *RRSIG not valid by DNSKEY* set.
```
(From https://raw.githubusercontent.com/zonemaster/zonemaster/develop/docs/specifications/tests/DNSSEC-TP/dnssec02.md)

tgreenx:
I believe I have made the appropriate fix, please re-review @matsduf.

```
zonemaster-cli --test DNSSEC/dnssec02 --show-testcase 249.200.193.in-addr.arpa --raw --level INFO
0.00 INFO      UNSPECIFIED    GLOBAL_VERSION   version=v4.4.0
```

PS: The specification says, [line 156](https://github.com/zonemaster/zonemaster/blame/develop/docs/specifications/tests/DNSSEC-TP/dnssec02.md#L156): `If there is more than one such DNSKEY, select the correct one.` How is that choice supposed to be made? In the implementation, that choice seems to be made from this loop, [lines 1501 to 1510](https://github.com/zonemaster/zonemaster-engine/blob/master/lib/Zonemaster/Engine/Test/DNSSEC.pm#L1501-L1510), which looks at the digest between the DNSKEY and the DS. In our case (i.e., when that check can't be made due to unsupported digest algorithms) I have made the change so that the first DNSKEY with the appropriate keytag is selected for further testing, even if others (with same keytag) exist. Is that ok?

matsduf:
> PS: The specification says, [line 156](https://github.com/zonemaster/zonemaster/blame/develop/docs/specifications/tests/DNSSEC-TP/dnssec02.md#L156): `If there is more than one such DNSKEY, select the correct one.` How is that choice supposed to be made?

The DS record uses the key tag to refer to the DNSKEY its digest is based on. The key tag is calculated from the DNSKEY. If the DNSKEY record does not match the DS by key tag, then it is ignored. If key tag matches, DNSKEY algorithm must also match.

Key tag is a 16-bit integer, which means that the could are many DNSKEY records with the same key tag. And it is possible that there are more than one DNSKEY for the same zone with that same algorithm and key tag, even if it is unlikely. If so, even if it is unlikely, the DS comparison should be done against both (all) to select the correct one, and then that should be selected.

Will the code get too complex for an unlikely case? Should we warn in DNSSEC05 and then ignore one key in other test case if a zone has more than one key with the same algorithm and key tag?

> In the implementation, that choice seems to be made from this loop, [lines 1501 to 1510](https://github.com/zonemaster/zonemaster-engine/blob/master/lib/Zonemaster/Engine/Test/DNSSEC.pm#L1501-L1510), which looks at the digest between the DNSKEY and the DS. In our case (i.e., when that check can't be made due to unsupported digest algorithms) I have made the change so that the first DNSKEY with the appropriate keytag is selected for further testing, even if others (with same keytag) exist. Is that ok?

I think that is OK if they have the same algorithm, but then that is a design issue, and we should update DNSSEC05 an commented above.

tgreenx:
[...]
> ~I think that is OK if they have the same algorithm, but then that is a design issue, and we should update DNSSEC05 an commented above.~ After thinking about it again I think we must have more correct approach. So what should the test case do if If two or more DNSKEY have the same key tag as the DS record?
> 
>     * If only one of the DNSKEY records has the same algorithm as pointed out by the DS it is simple, select that DNSKEY.
> 
>     * If two or more of the DNSKEY records have the same algortithm as pointed out by the DS, then a comparison with the DS digest must be done.
>       
>       * If none matches, then report that.
>       * Else take the first that matches.

Thanks for the explanation. I understand, however here I am talking about the specific case where that digest comparison can't be done (because the digest algorithm isn't supported by ZM, which is the root cause of this PR). So to be clear: if we have several DNSKEY records that match a DS record by keytag and have the same algorithm, and if a comparison of the digests can't be done, then the first DNSKEY record is selected for further testing.

matsduf:
> Thanks for the explanation. I understand, however here I am talking about the specific case where that digest comparison can't be done (because the digest algorithm isn't supported by ZM, which is the root cause of this PR). So to be clear: if we have several DNSKEY records that match a DS record by keytag and have the same algorithm, and if a comparison of the digests can't be done, then the first DNSKEY record is selected for further testing.

Yes, that could be fine. We should update the specification. This will, however, be very rare.

pnax (approved):
lgtm

lib/Zonemaster/Engine/Test/DNSSEC.pm
```diff
@@ -1498,16 +1498,19 @@ sub dnssec02 {
...
                     if ( exists $LDNS_digest_algorithms_supported{$ds->digtype()} ) {
                         my $tmp_ds = $matching_keytag_dnskey->ds($LDNS_digest_algorithms_supported{$ds->digtype()});
-                        if ( $tmp_ds->hexdigest() eq $ds->hexdigest() ) {
+
+                        if ( not $tmp_ds or $tmp_ds->hexdigest() eq $ds->hexdigest() ) {
                             $matching_dnskey = $matching_keytag_dnskey;
                             $match_ds_dnskey = 1;
                             last;
                         }
                     }
```
matsduf, line 1511: Let us say that we have one domain with DS digest algorithm 3, not supported by the installation. Then on line 1506 the code will ignore that problem and do not output a message that there is no match. Fine.

Let us say that we have another domain with DS digest algorithm 5, also no supported. But because it is not included in %LDNS_digest_algorithms_supported the code goes another way, and we will get another message outputted, DS02_NO_MATCH_DS_DNSKEY

There should be no difference.

lib/Zonemaster/Engine/Test/DNSSEC.pm
```diff
@@ -1498,16 +1498,19 @@ sub dnssec02 {
...
                 }
+
                 if ( scalar @matching_keytag_dnskeys == 1 and not $match_ds_dnskey) {
```
matsduf, line 1514: If exact 1, but if 2 or more?

matsduf (changes requested):

lib/Zonemaster/Engine/Test/DNSSEC.pm
tgreenx: Actually in that case it will output DS02_NO_DNSKEY_FOR_DS, which is the current behavior of DNSSEC02 (unchanged by this PR). Is that not expected?

tgreenx: Fixed, behavior will be consistent for any unsupported digest algorithms (either because of unknown algorithm or missing library)

lib/Zonemaster/Engine/Test/DNSSEC.pm
tgreenx: Fixed. First DNSKEY in matching set will be selected for further testing in case none were correctly selected before.

lib/Zonemaster/Engine/Test/DNSSEC.pm
```diff
@@ -1498,17 +1498,25 @@ sub dnssec02 {
                 my $matching_dnskey = undef;
                 my @matching_keytag_dnskeys = grep { $ds->keytag == $_->keytag } @dnskey_rrs;
                 my $match_ds_dnskey = 0;
+
                 foreach my $matching_keytag_dnskey ( @matching_keytag_dnskeys ) {
                     if ( exists $LDNS_digest_algorithms_supported{$ds->digtype()} ) {
```
pnax, line 1503: I think the `if ( exists ... )` could be moved outside the `foreach`.

lib/Zonemaster/Engine/Test/DNSSEC.pm
```diff
@@ -1498,17 +1498,25 @@ sub dnssec02 {
...
                             $matching_dnskey = $matching_keytag_dnskey;
                             $match_ds_dnskey = 1;
                             last;
                         }
                     }
+                    else{
+                        $matching_dnskey = $matching_keytag_dnskey;
+                        $match_ds_dnskey = 1;
+                        last;
+                    }
```
pnax, line 1516: In case the `if ( exists ... )` is moved outside the foreach, I think this `else` block can be safely removed.

lib/Zonemaster/Engine/Test/DNSSEC.pm
pnax, line 1503: maybe add a comment that we need to check that Zonemaster supports the algorithm to look for the correct DNSKEY in case there are several ones.

pnax (approved):
beside small suggestions, this looks good

PR #1056, opened 2022-04-28
Use libidn2 instead of libidn

pnax:
## Purpose

Replace libidn with libidn2.

## Context

https://github.com/zonemaster/zonemaster-ldns/pull/133

## Changes

Update Installation.md document and Github Action file.

## How to test this PR

Install Zonemaster-LDNS that uses libidn2 and all tests should pass.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -47,7 +47,7 @@ jobs:
             automake \
             build-essential \
             gettext \
-            libidn11-dev \
+            libidn2-dev \
             libssl-dev \
             libtool \
             m4 \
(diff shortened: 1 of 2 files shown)


matsduf (approved):

pnax:
The failure in Github Action comes from the fact that we test against Zonemaster::LDNS master state which does not support libidn2 yet. This is at this stage only available in develop. To be sure all tests are successful against develop, I opened #1058 and removed the `latest` matrix entry. [The tests pass](https://github.com/zonemaster/zonemaster-engine/runs/6211903137) when performed with Zonemaster::LDNS develop state.
So I'm merging this PR even though the tests are marked as failed here.

Issue #899, opened 2021-05-11
GitHub Actions workflows modified (remove changes as possible)


vlevigneron:
- In order to get checks completed for zonemaster/zonemaster-engine#896 PR, worflows had been modified to only work on 'develop' branch instead 'develop' + 'latest'.
- Restore 'develop' + 'latest' checks as soon as possible.

pnax:
The same problem occurred again recently. So 'latest' has been disabled again with #1062. What is the point in testing Zonemaster::Engine develop branch against Zonemaster::LDNS master branch?

The change has been introduced in 6b81954b from PR #890. This PR contains a discussion on testing against Zonemaster::LDNS from CPAN (see https://github.com/zonemaster/zonemaster-engine/pull/890#issuecomment-806567263 and following comments).

Since this change, this check against "latest" has been removed twice: https://github.com/zonemaster/zonemaster-engine/commit/88075c0c193fd9175727c7d1a558230d85869937 (from #896) and https://github.com/zonemaster/zonemaster-engine/commit/569dc1a7a0decda8a9507c3025449eb71b77a593 (from #1062). The first removal occurred during the same release cycle as the feature insertion.

It looks like we need to question the need for such test. Or think about the way we change things between Zonemaster::LDNS and Zonemaster::Engine.

matsduf:
> The same problem occurred again recently. So 'latest' has been disabled again with #1062. What is the point in testing Zonemaster::Engine develop branch against Zonemaster::LDNS master branch?

I am not sure, but it could be that if it fails against develop branch and not against master branch then you might find the cause.

In the Travis model develop branch is tested against develop branch (and master branch against master branch).

I suggest that we permanently keep "stable" commented, and only uncomment in case of need.

matsduf:
Let's see if @mattias-p har any insights into it.

The behavior we want, I think, is that PRs against develop branch in Engine depend on Zonemaster-LDNS in develop branch to be able to handle changes that are not backward compatible.

PR #1062, opened 2022-05-04
Momentarily comment out "latest" from Github Action

pnax:
## Purpose

Momentarily drop the unit tests relying on fetching Zonemaster::LDNS from CPAN due to an incompatibility with the required version of libidn. The CPAN module requires libidn11-dev, but current develop migrated to libidn2-dev.

## Context

Github Action is failing since merging #1056.

## Changes

Github Action file.

## How to test this PR

Tests should pass.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,7 +18,7 @@ jobs:
       matrix:
         compatibility:
           - develop
-          - latest
+          #- latest
         perl:
           - '5.32'
           - '5.26'


matsduf:
Travis is better in that respect. There we test develop branch against develop branch.

To resolve this kind of issue last time we commented master branch. I suggest that we do that again instead, because then we will know that develop build against develop without any modifications.

pnax:
> because then we will know that develop build against develop without any modifications.

Good point. I've commented out the part that will configure the failing tests.

matsduf (approved):
Please create an issue that we should re-enable latest, or maybe question to have it.

pnax:
The issue already exists: #899

PR #1045, opened 2022-03-23
Remove some unintended network dependencies from the unit tests

mattias-p:
## Purpose

This PR makes the unit tests don't autocomplete fake glue by sending out DNS requests.

## Context

Fixes #888.

## Changes

A new method Zonemaster::Engine->add_fake_delegation_raw() is factored out from Zonemaster::Engine->add_fake_delegation(). The new method performs the meat of things but without the bells and whistles like autocompleting fake glue and returning a boolean. (I don't think the return value is ever used and it's kind of nonsense anyway. Maybe I should have removed it altogether.)

## How to test this PR

1. In one terminal run `sudo tcpdump -nn port 53` and keep it open.
2. In another terminal run `make test` and wait for it to complete.
3. Make sure tcpdump didn't list any DNS requests from the tests.

PR #1068, opened 2022-05-18
Fix breakage in Backend tests

mattias-p:
## Purpose

* This PR fixes a breakage in Backend as seen in zonemaster/zonemaster-backend#1008.

## Context

Fixes a bug that was introduced in #1045.

## Changes

Makes `undef` an alias for `[]` in the arguments to add_fake_delegation().

## How to test this PR

Run the Backend unit tests using this version of Engine.

### DIFF
--- a/lib/Zonemaster/Engine.pm
+++ b/lib/Zonemaster/Engine.pm
@@ -116,8 +116,9 @@ sub add_fake_delegation {
     foreach my $name ( keys %{$href} ) {
         $name =~ /[^.]$|^\.$/
           or croak 'Each key of argument $href must omit the trailing dot, or it must be a single dot';
-        ref $href->{ $name } eq 'ARRAY'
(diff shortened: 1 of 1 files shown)


lib/Zonemaster/Engine.pm
```diff
@@ -323,6 +324,8 @@ This method adds some fake delegation information to the system.
 The arguments are a domain name, and a hashref with delegation information.
 The keys in the hash are nameserver names, and the values are arrayrefs of IP
 addresses for their corresponding nameserver.
+Alternatively the IP addresses may be specified as an `undef` which is handled
+the same as an empty arrayref.
```
matsduf, line 328: Is "empty arrayref" correct in strict meaning? Shouldn't it be "reference to an empty array"?

pnax (approved):

tgreenx (approved):

PR #1069, opened 2022-05-18
Make query timeout configurable in profile

blacksponge:
## Purpose

The timeout of queries is currently not configurable. This PR makes it configurable in the profile.

## Context

Asked on the mailing list.
>  I wanted to change this to just 1 query. Is it configurable?

## Changes

* Make the timeout configurable.
* The default is the current default (LDNS default)

## How to test this PR

* Query a NS that does not respond, check for the time taken to fail (two times the timeout as there is a retry)
   `../zonemaster-cli/script/zonemaster-cli --level debug --ns localhost/127.0.0.1 zonemaster.net`
* Lower the timeout, check that it fails faster.


pnax (approved):
works as expected

pnax (approved):

tgreenx (approved):

matsduf (approved):

PR #1097, opened 2022-07-19
Missing update of ns args in SYSTEM messages

pnax:
## Purpose

Commit 766c1116 (PR #792) missed some messages. This fixes it (and reverted the argument based on the conversation here, from "ns_list" to "ns").
Also the query type is logged as well.

## Context

Follow-up on #792 

## Changes

* Test/Zone.pm
* Zone.pm
* Translator.pm: the msgid for SKIP_IPV[46]_DISABLED is updated
* Some refactoring as well

## How to test this PR

The following command should not display `{ns_list}` anymore:
```
zonemaster-cli --no-ipv4 --test zone nic.fr --level DEBUG --show-testcase --show-testcase --locale en_US.UTF-8 | grep "disabled"
```
And the query type should also be logged.

### DIFF
--- a/lib/Zonemaster/Engine/Test/Zone.pm
+++ b/lib/Zonemaster/Engine/Test/Zone.pm
@@ -771,8 +771,8 @@ sub _is_ip_version_disabled {
     if ( not Zonemaster::Engine::Profile->effective->get( q{net.ipv4} ) and $ns->address->version == $IP_VERSION_4 ) {
         Zonemaster::Engine->logger->add(
             SKIP_IPV4_DISABLED => {
-                ns_list => $ns->string,
-                rrtype  => $type
+                ns     => $ns->string,
+                rrtype => $type
             }
         );
         return 1;
@@ -781,8 +781,8 @@ sub _is_ip_version_disabled {
     if ( not Zonemaster::Engine::Profile->effective->get( q{net.ipv6} ) and $ns->address->version == $IP_VERSION_6 ) {
         Zonemaster::Engine->logger->add(
             SKIP_IPV6_DISABLED => {
-                ns_list => $ns->string,
-                rrtype  => $type
+                ns     => $ns->string,
+                rrtype => $type
             }
         );
         return 1;
--- a/lib/Zonemaster/Engine/Test/Zone.pm
+++ b/lib/Zonemaster/Engine/Test/Zone.pm
@@ -769,12 +769,12 @@ sub _is_ip_version_disabled {
     my $ns = shift;
 
     if ( not Zonemaster::Engine::Profile->effective->get( q{net.ipv4} ) and $ns->address->version == $IP_VERSION_4 ) {
-        Zonemaster::Engine->logger->add( SKIP_IPV4_DISABLED => { ns => "$ns" } );
+        Zonemaster::Engine->logger->add( SKIP_IPV4_DISABLED => { ns_list => "$ns" } );
         return 1;
     }
 
     if ( not Zonemaster::Engine::Profile->effective->get( q{net.ipv6} ) and $ns->address->version == $IP_VERSION_6 ) {
-        Zonemaster::Engine->logger->add( SKIP_IPV6_DISABLED => { ns => "$ns" } );
+        Zonemaster::Engine->logger->add( SKIP_IPV6_DISABLED => { ns_list => "$ns" } );
         return 1;
     }
 
--- a/lib/Zonemaster/Engine/Test/Zone.pm
+++ b/lib/Zonemaster/Engine/Test/Zone.pm
@@ -323,7 +323,7 @@ sub zone01 {
 
                 my $ns = Zonemaster::Engine::Nameserver->new( { name => $soa_mname, address => $ip_address->short } );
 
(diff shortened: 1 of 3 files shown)


lib/Zonemaster/Engine/Test/Zone.pm
```diff
@@ -769,12 +769,12 @@ sub _is_ip_version_disabled {
     my $ns = shift;
 
     if ( not Zonemaster::Engine::Profile->effective->get( q{net.ipv4} ) and $ns->address->version == $IP_VERSION_4 ) {
-        Zonemaster::Engine->logger->add( SKIP_IPV4_DISABLED => { ns => "$ns" } );
+        Zonemaster::Engine->logger->add( SKIP_IPV4_DISABLED => { ns_list => "$ns" } );
```
matsduf, line 772: Isn't $ns a single name server? If so it should be "ns" not "ns_list".

matsduf: I agree, but I think the correction should be the other way around. It should be "ns" in all places, not "ns_list". See https://github.com/zonemaster/zonemaster-engine/blob/master/docs/logentry_args.md

```
    SKIP_IPV4_DISABLED => sub {
        __x    # SYSTEM:SKIP_IPV4_DISABLED
          "IPv4 is disabled, not sending query to {ns_list}.", @_;
    },
    SKIP_IPV6_DISABLED => sub {
        __x    # SYSTEM:SKIP_IPV6_DISABLED
          "IPv6 is disabled, not sending query to {ns_list}.", @_;
    },
```
(https://github.com/zonemaster/zonemaster-engine/blob/develop/lib/Zonemaster/Engine/Translator.pm#L74-L81)

```
$ grep SKIP_IPV4_DISABLED lib/Zonemaster/Engine/Zone.pm lib/Zonemaster/Engine/Test/Zone.pm
lib/Zonemaster/Engine/Zone.pm:            Zonemaster::Engine->logger->add( SKIP_IPV4_DISABLED => { ns_list => "$ns" } );
lib/Zonemaster/Engine/Zone.pm:        Zonemaster::Engine->logger->add( SKIP_IPV4_DISABLED => { ns_list => ( join ';', map { "$_" } @nope ) } );
lib/Zonemaster/Engine/Zone.pm:            Zonemaster::Engine->logger->add( SKIP_IPV4_DISABLED => { ns_list => "$ns" } );
lib/Zonemaster/Engine/Zone.pm:            Zonemaster::Engine->logger->add( SKIP_IPV4_DISABLED => { ns_list => "$ns" } );
lib/Zonemaster/Engine/Test/Zone.pm:        Zonemaster::Engine->logger->add( SKIP_IPV4_DISABLED => { ns => "$ns" } );
```

pnax: > ```
>  [...]->add( SKIP_IPV4_DISABLED => { ns_list => ( join ';', map { "$_" } @nope ) } );
> ```

This pretty much looks like `@nope` is a list of NS here.

lib/Zonemaster/Engine/Translator.pm
```diff
@@ -73,11 +73,11 @@ Readonly my %TAG_DESCRIPTIONS => (
     },
     SKIP_IPV4_DISABLED => sub {
         __x    # SYSTEM:SKIP_IPV4_DISABLED
-          "IPv4 is disabled, not sending query to {ns_list}.", @_;
+          "IPv4 is disabled, not sending \"{rrtype}\" query to {ns_list}.", @_;
     },
     SKIP_IPV6_DISABLED => sub {
         __x    # SYSTEM:SKIP_IPV6_DISABLED
-          "IPv6 is disabled, not sending query to {ns_list}.", @_;
+          "IPv6 is disabled, not sending \"{rrtype}\" query to {ns_list}.", @_;
```
matsduf, line 80: "ns_list" => "ns" both for IPv4 and IPv6

matsduf (changes requested):

lib/Zonemaster/Engine/Test/Zone.pm
pnax: Ok I found a solution. Instead of joining the NS into a single list, I chose to display on message per item.

matsduf: When a message (msgid) is created it is important to know if it is a list of NS or just a single NS. If it is a list it must be put at the end to make the message readable. If you know that it is just a single NS then you can make a more readable message by putting it earlier.

If a message is by default on INFO or higher level I prefer list to prevent the result to have 5-10 message with the same thing. On DEBUG level I think that multiple messages with single NS makes more sense.

tgreenx (approved):

matsduf (approved):

matsduf:
Do you plan to do anything more in this area?

1. Not all test cases output IPVx_DISABLED.
2. There is another message, SKIP_IPV6_DISABLED
   1. Should there be two different messages?
   2. That message actually means the opposite that it sounds, SKIP_IPV6_DISABLED == IPV6_DISABLED. It is not skipped disabling, it is skipped IPv6.

mattias-p (approved):
LGTM

PR #1098, opened 2022-07-19
Refactoring, avoid code duplication

pnax:
## Purpose

Remove some code redundancy when dealing with message (especially IPV4_DISABLED and IPV6_DISABLED).

## Context

n/a

## Changes

* remove some trailing spaces
* Test/DNSSEC.pm
* Test/Consistency.pm
* Test/Connectivity.pm
* Test/Delegation.pm
* Test/Nameserver.pm
* Test/Basic.pm
* Test/Syntax.pm

## How to test this PR

Refactoring only. Everything should work as before. Tests should pass.

### DIFF
--- a/lib/Zonemaster/Engine/Test/Basic.pm
+++ b/lib/Zonemaster/Engine/Test/Basic.pm
@@ -286,6 +286,61 @@ sub version {
     return "$Zonemaster::Engine::Test::Basic::VERSION";
 }
 
+sub _ip_disabled_message {
+    my ( $results_array, $ns, @rrtypes ) = @_;
+
+    if ( not Zonemaster::Engine::Profile->effective->get(q{net.ipv4}) and $ns->address->version == $IP_VERSION_4 ) {
+        push @$results_array, map {
+          info(
+            IPV4_DISABLED => {
+                ns     => $ns->string,
+                rrtype => $_,
+            }
+          )
+        } @rrtypes;
+        return 1;
+    }
+
+    if ( not Zonemaster::Engine::Profile->effective->get(q{net.ipv6}) and $ns->address->version == $IP_VERSION_6 ) {
+        push @$results_array, map {
+          info(
+            IPV6_DISABLED => {
+                ns     => $ns->string,
+                rrtype => $_,
+            }
+          )
+        } @rrtypes;
+        return 1;
+    }
+    return 0;
+}
+
+sub _ip_enabled_message {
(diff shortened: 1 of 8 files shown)


lib/Zonemaster/Engine/Test/Basic.pm
```diff
@@ -286,6 +286,61 @@ sub version {
...
 
+sub _ip_disabled_message {
+    my ( $results_array, $ns, @rrtypes ) = @_;
+
+    if ( not Zonemaster::Engine::Profile->effective->get(q{net.ipv4}) and $ns->address->version == $IP_VERSION_4 ) {
+        push @$results_array, map {
+          info(
+            IPV4_DISABLED => {
+                ns     => $ns->string,
+                rrtype => $_,
```
matsduf, line 297: rrtype is not a defined argument name in "[Arguments for test case messages]". Could you include an update that defines it while you are at it?

[Arguments for test case messages]: https://github.com/zonemaster/zonemaster-engine/blob/master/docs/logentry_args.md

matsduf (commented):
It looks good, but wouldn't it be even better to defined the function in one place instead of repeating it in every level method?

pnax:
> It looks good, but wouldn't it be even better to defined _ip_disabled_message() and _ip_enabled_message() in one place instead of repeating it in every level method?

Totally. I wasn't sure where to put such functions. I think I'll put them in Util.pm

pnax:
> > It looks good, but wouldn't it be even better to defined _ip_disabled_message() and _ip_enabled_message() in one place instead of repeating it in every level method?
>
> Totally. I wasn't sure where to put such functions. I think I'll put them in Util.pm

Well in fact it can't be done easily as the info() call should happen in the test module because of:
https://github.com/zonemaster/zonemaster-engine/blob/968f7d942d2b2ecfb5168778c160d2f869676a74/lib/Zonemaster/Engine/Logger/Entry.pm#L116-L122

So I won't factorize the code further not touch the Util.pm file.

pnax:
> From the change one can see that the message is not consistently outputted from the test cases. E.g. Nameserver10 does not output any message, and there are more of them.

Indeed. I'll look to harmonize all this in a follow-up PR to avoid overloading this one. If you think this one is ready, we could merge it so I'll move forward on the harmonization.

marc-vanderwal:
Unit tests pass and I have witnessed no regression during manual testing.

PR #1102, opened 2022-07-20
Consistently output IPV4_DISABLED/IPV6_DISABLED message in DEBUG

pnax:
## Purpose

Such messages have a DEBUG level and were not always displayed. This
significantly increases the number of DEBUG messages.

## Context

Follow up from #1098, see https://github.com/zonemaster/zonemaster-engine/pull/1098#issuecomment-1190342319

## Changes

* Test/Nameserver.pm

## How to test this PR

Playing with `--no-ipv4` and `--no-ipv6` should give more DEBUG messages:
```
zonemaster-cli --no-ipv4 --test nameserver --level DEBUG --show-module --show-testcase --show-testcase --locale en_US.UTF-8 nic.fr | grep IP
```

### DIFF
--- a/lib/Zonemaster/Engine/Test/Nameserver.pm
+++ b/lib/Zonemaster/Engine/Test/Nameserver.pm
@@ -450,14 +450,9 @@ sub nameserver01 {
         @nss = values %nss;
     }
 
-    if ( not Zonemaster::Engine::Profile->effective->get(q{net.ipv6}) ) {
-        @nss = grep { $_->address->version != $IP_VERSION_6 } @nss;
-    }
-    if ( not Zonemaster::Engine::Profile->effective->get(q{net.ipv4}) ) {
-        @nss = grep { $_->address->version != $IP_VERSION_4 } @nss;
-    }
-
     for my $ns ( @nss ) {
+
+        next if ( _ip_disabled_message( \@results, $ns, q{A} ) );
 
         my $response_count = 0;
         my $nxdomain_count = 0;
@@ -510,9 +505,7 @@ sub nameserver02 {
     foreach
       my $local_ns ( @{ Zonemaster::Engine::TestMethods->method4( $zone ) }, @{ Zonemaster::Engine::TestMethods->method5( $zone ) } )
     {
-        next if ( not Zonemaster::Engine::Profile->effective->get(q{net.ipv6}) and $local_ns->address->version == $IP_VERSION_6 );
-
-        next if ( not Zonemaster::Engine::Profile->effective->get(q{net.ipv4}) and $local_ns->address->version == $IP_VERSION_4 );
+        next if ( _ip_disabled_message( \@results, $local_ns, q{SOA} ) );
 
         next if $nsnames_and_ip{ $local_ns->name->string . q{/} . $local_ns->address->short };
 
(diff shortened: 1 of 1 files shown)


tgreenx:
As discussed privately, this will now add messages to the Engine::Logger object `@results` which will disturb checks such as the one found in Nameserver04, L653: `and not grep { $_->tag ne q{TEST_CASE_START} } @results )`. Thus a small refactoring is needed for test cases with such a check.

tgreenx (changes requested):
As discussed privately, this will now add messages to the Engine::Logger object @results which will disturb checks such as the one found in [Nameserver04, L621](https://github.com/zonemaster/zonemaster-engine/blob/master/lib/Zonemaster/Engine/Test/Nameserver.pm#L621): and not grep { $_->tag ne q{TEST_CASE_START} } @results ). Thus a small refactoring is needed for test cases with such a check (Nameserver02, 04 and 07 from what I can see), otherwise the results will differ. (Note that unit tests don't capture such an issue since they are run without any options)

matsduf:
> Such messages have a DEBUG level and were not always displayed. This significantly increases the number of DEBUG messages.

They are DEBUG execpt in Basic where they are INFO, on purpose (I actually think they should be NOTICE in Basic). The increase number of DEBUG messages is not a large problem, but the increase of INFO messages from Basic is maybe a problem. Ideally I think Basic04 should output such a message on NOTICE level and the other Basic test cases on DEBUG level. That could be achieved if we have specific messages from Basic04 and then make the standard message DEBUG.

matsduf:
> As discussed privately,  (...)

@tgreenx, could you elaborate and explain? I do not see the issue.

pnax:
> could you elaborate and explain? I do not see the issue.

The following commands do not output the same results, and they should:
```
zonemaster-cli --test nameserver/nameserver04 --show-module --show-testcase --show-testcase --locale en_US.UTF-8 --level INFO afnic.fr
zonemaster-cli --no-ipv4 --test nameserver/nameserver04 --show-module --show-testcase --show-testcase --locale en_US.UTF-8 --level INFO afnic.fr
```

matsduf:
Now I see the issue. The code wants to check if any DIFFERENT_SOURCE_IP message has been outputted. I think it would be simpler to have a scalar the is increased by 1 each time such a message is outputted. I think that would be more stable.

pnax:
> this will now add messages to the Engine::Logger object `@results` which will disturb checks

This should now be fixed. Please re-review.

matsduf:
I think it is OK, except in Basic, because there the message is on INFO level. There will be too many messages for the normal user. 

What we need is to have a special message in Basic04 which is outputted at the most once per IP address, maybe just once, on INFO or even NOTICE level, and then we could have the normal messages per query not sent on DEBUG level.

This PR is just about Nameserver, but I assume the plan is to implement it in all modules.

matsduf (approved):

pnax:
> I think it is OK, except in Basic, because there the message is on INFO level. There will be too many messages for the normal user.
> What we need is to have a special message in Basic04 which is outputted at the most once per IP address, maybe just once, on INFO or even NOTICE level, and then we could have the normal messages per query not sent on DEBUG level.

The behavior your are describing is already present in master. If this is really an issue, maybe create an issue and/or a PR with a fix.

> This PR is just about Nameserver, but I assume the plan is to implement it in all modules.

I think this will be the case already now that this PR is merged, since most of the work was conducted in #1098.

matsduf:
> The behavior your are describing is already present in master. If this is really an issue, maybe create an issue and/or a PR with a fix.

I will look into it.

PR #1108, opened 2022-07-26
Fixes bug in Net::IP with a work around. Issue #1107

### LINKED ISSUE #1107: Net::IP creates wrong IPv4 reverse name when last octet is "0"
Net::IP::XS is used by Zonemaster i.e. to create reverse names of IP addresses. If Net::IP::XS is not available, Net::IP is used instead. The issue reported here is affecting Net::IP and not Net::IP::SX, which means that OSs that do load Net::IP::SX are not affected (FreeBSD, Debian, Ubuntu).

Affected OSs are CentOS 7 Linux and Docker. It is unknown if Rocky Linux is affected.

A work around for affected OSs can be, if possible, to install Net::IP::XS. Version 0.21 works with Zonemaster. If lower version also works has not been tested.

## Issue
Creating the reverse name using Net::IP when the last octet **is not** "0":
```
$ perl -E 'use Net::IP; $ip = new Net::IP ("10.11.12.13"); say $ip->reverse_ip'
13.12.11.10.in-addr.arpa
```
Creating the reverse name using Net::IP when the last octet **is** "0":
```
$ perl -E 'use Net::IP; $ip = new Net::IP ("10.11.12.0"); say $ip->reverse_ip'
12.11.10.in-addr.arpa.
```
*Note that the first label, that corresponds to the last octet, is missing.*

The same thing using Net::IP::XS:
```
$ perl -E 'use Net::IP::XS; $ip = new Net::IP::XS ("10.11.12.13"); say $ip->reverse_ip'
13.12.11.10.in-addr.arpa.

$ perl -E 'use Net::IP::XS; $ip = new Net::IP::XS ("10.11.12.0"); say $ip->reverse_ip'
0.12.11.10.in-addr.arpa.
```
*Note that both reverse names are correct.*

## IPv6 reverse
IPv6 reverse is not affected by the issue.
```
$ perl -E 'use Net::IP; $ip = new Net::IP ("2a05:d016:c84:7600::1"); say $ip->reverse_ip'
1.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.6.7.4.8.c.0.6.1.0.d.5.0.a.2.ip6.arpa.

$ perl -E 'use Net::IP; $ip = new Net::IP ("2a05:d016:c84:7600::0"); say $ip->reverse_ip'
0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.6.7.4.8.c.0.6.1.0.d.5.0.a.2.ip6.arpa.
```

## Ubuntu

When Ubuntu 22.04 is installed using the CPAN method, Net::IP::XS is not installed, but it is till not affected. Net::IP is claimed to be the same version (1.26) but it has been updated in the relevant part of the code. Correct output on Ubuntu:
```
$ perl -E 'use Net::IP; $ip = new Net::IP ("10.11.12.0"); say $ip->reverse_ip'
0.12.11.10.in-addr.arpa.
```

## Reference

This issue was discovered and reported by @anandb-ripencc in #1105.

The bug in Net::IP is reported in https://rt.cpan.org/Public/Bug/Display.html?id=129964

matsduf:
## Purpose

Net::IP does not make correct reverse name when last octet is a "0". This is a work-around the bug.

## Context

Resolves #1107

## How to test this PR

On a system where Net::IP::XS is not installed (e.g. CentOS 7) run the following command and verify that one of the last lines have a PTR query where the IP address ends with "0" and that the name is correct.
```
zonemaster-cli 15.248.151.in-addr.arpa --raw --show_testcase --test address/address02 --level DEBUG
```

### DIFF
--- a/lib/Zonemaster/Engine/Net/IP.pm
+++ b/lib/Zonemaster/Engine/Net/IP.pm
@@ -61,6 +61,12 @@ sub print {
 
 sub reverse_ip {
     my $self = shift;
+    my $addr = $self->{_inner}->short;
+    # Work around for IPv4 address due to a bug in Net::IP when the last octet is "0"
+    # https://github.com/zonemaster/zonemaster-engine/issues/1107
+    if ($addr =~ /^\d{1,3}\.\d{1,3}\.\d{1,3}\.0$/) { # IPv4 x.x.x.0
+        return join ('.', reverse (split (/\./, $addr))) . '.in-addr.arpa.';
+    };
     return $self->{_inner}->reverse_ip();
 }
 


anandb-ripencc:
Can be fixed by explicitly installing Net::IP::XS

matsduf:
> Can be fixed by explicitly installing Net::IP::XS

The fix is still needed. Firstly, we know that the Docker installation is affected. It is unknown if the Rocky Linux is affected. Secondly, as long as Zonemaster supports running with Net::IP, which is an alternative dependency, the bug must be worked around.

pnax:
> as long as Zonemaster supports running with Net::IP

Couldn't Zonemaster depend on Net::IP::XS instead of Net::IP?

matsduf:
> Couldn't Zonemaster depend on Net::IP::XS instead of Net::IP?

Maybe. But as far as I can see Alpine Linux does not have any package with Net::IP::XS.

My solution is a work-around that I think the code should have as long as it uses Net::IP as an alternative dependency. If that is removed, the work-around is not needed.

pnax:
> But as far as I can see Alpine Linux does not have any package with Net::IP::XS.

Can't we install it from CPAN? It looks like we already install some packages from CPAN.

matsduf:
> Can't we install it from CPAN? It looks like we already install some packages from CPAN.

Possibly. When I saw the issue I also saw a simple work-around that would be safe add. If Net::IP is kept, I think the work-around is a correct solution, if Net::IP is dropped, the work-around is not needed and should not be kept or added.

anandb-ripencc:
I don't understand the problem with the Docker image. In the Dockerfile, I see that you're installing some perl modules from packages, and then some directly from CPAN using cpanm. So you can just remove the `perl-net-ip` package, and instead add Net::IP::XS to the list of modules installed directly from CPAN.

To add to this discussion, our RIPE NCC Zonemaster build does NOT use any perl modules from the base distribution. We're using cpanm to install EVERYTHING into a perl local::lib directory. We do this, because we don't like mixing OS provided perl modules and CPAN provided perl modules, because it causes massive confusion.

To summarise, I think you should remove Net::IP as a dependency, and specify only Net::IP::XS as a dependency. If it's available as a package, a user can install it. Otherwise, it comes from CPAN.

matsduf:
This is a work-around that is required if Net::IP is used. Nothing else.

blacksponge:
> To summarise, I think you should remove Net::IP as a dependency, and specify only Net::IP::XS as a dependency. If it's available as a package, a user can install it. Otherwise, it comes from CPAN.

I agree with this, even on a performance aspect Net::IP::XS performs way better, if we could have only this one as a dependency it would be nice.

mvw-afnic (approved):
Looks good to me.

anandb-ripencc:
This PR will be obsoleted once #1119 is merged, so it's better to close this. There's no need for this work-around.

matsduf:
@mvw-afnic, can you review #1119 instead? When that is merged I will close this.

matsduf:
Replaced by #1119

mvw-afnic:
Will do!

Issue #1071, opened 2022-05-18
Fallback to EDNS or TCP can be in conflict with test cases


matsduf:
https://github.com/zonemaster/zonemaster-engine/blob/master/lib/Zonemaster/Engine/Profile.pm#L588-L597 says

> resolver.defaults.fallback
A boolean. If true, UDP queries that get responses with the C<TC>
flag set will be automatically resent over TCP or using EDNS. Default 
true.
In ldns-1.7.0 (NLnet Labs), in case of truncated answer when UDP is used,
the same query is resent with EDNS0 and TCP (if needed). If you
want the original answer (with TC bit set) and avoid this kind of
replay, set this flag to false.

I am not sure that this should be configurable here, since this can be in conflict with the test cases. It is probably more reasonable that it is set to false in the calls to LDNS, and then handled by the Zonemaster code.

PR #1109, opened 2022-07-27
Update implementation Zone09

### LINKED ISSUE #1103: Implement updated zone09
Test case Zone09 has been updated. See https://github.com/zonemaster/zonemaster/pull/870

The implemenation must be updated.

tgreenx:
## Purpose

This PR proposes a new implementation for test case Zone09, as the specification was recently [updated](https://github.com/zonemaster/zonemaster/pull/870).

## Context

Fixes #1103 

## Changes

(per commit)

- Add new test messages
- Add new implementation
- Modify unitary tests

## How to test this PR

Checks should pass.


lib/Zonemaster/Engine/Test/Zone.pm
```diff
@@ -650,44 +706,131 @@ sub zone08 {
...
+
+        if ( not $p1 or $p1->rcode ne q{NOERROR} or not $p1->aa or not $p1->has_rrs_of_type_for_name(q{SOA}, $zone->name) ){
+            next;
+        }
+
+        my $p2 = $ns->query( $zone->name, q{MX}, { usevc => 0 } );
+
+        if ( not $p2 ){
+            $p2 = $ns->query( $zone->name, q{MX}, { usevc => 1 });
+        }
```
matsduf, line 737: Why this? It does not comply with the specification. If the TC flag is set in the response in $p2, then it should be retried over TCP. Else it should not. It would not reflect normal resolver behavior.

https://github.com/zonemaster/zonemaster/blob/develop/docs/specifications/tests/DNSQueryAndResponseDefaults.md#default-setting-in-dns-query

tgreenx: I just added this. I think I got confused then, after looking at #1071.

Should that whole code block instead be just this then, in case `fallback` becomes disable by default in profile?
```
        my $p2 = $ns->query( $zone->name, q{MX}, { fallback => 1 } );
```

matsduf: No, unfortunately, the fallback is not what we want, because we do no want EDNS unless specified. It sounds like  Zonemaster should have its own fallback method that does the following logic:
1. Send over UDP and LDNS fallback set to false.
2. Check if the response has the TC flag set.
3. Return response if TC flag was not set.
4. Retry the query with the same parameters, but over TCP.
5. Return response.
6. If no response from TCP, the orignal query is handled as no response.

I do not expect any MX queries to result in TC. 512 bytes is quite a lot.

tgreenx: I see. Please check latest commit ([3d7b643](https://github.com/zonemaster/zonemaster-engine/pull/1109/commits/3d7b6437801676d6146188db4ec120dc0a4387ce)). This should now be correctly handled as you mentioned.

lib/Zonemaster/Engine/Test/Zone.pm
```diff
@@ -650,44 +706,142 @@ sub zone08 {
...
+
+        if ( $p2_udp and not $p2_udp->tc ){
+            $p2 = $p2_udp;
+        }
+        else{
+            my $p2_tcp = $ns->query( $zone->name, q{MX}, { fallback => 0, usevc => 1 } );
+            
+            if ( $p2_tcp ){
+                $p2 = $p2_tcp;
             }
```
matsduf, line 744: If we get TC and resend by TCP, then the response by TCP should be considered to be the final response. I.e.
```
            $p2 = $ns->query( $zone->name, q{MX}, { fallback => 0, usevc => 1 } );
```

tgreenx: Fixed.

lib/Zonemaster/Engine/Test/Zone.pm
```diff
@@ -650,44 +706,131 @@ sub zone08 {
...
+
+        if ( not $p1 or $p1->rcode ne q{NOERROR} or not $p1->aa or not $p1->has_rrs_of_type_for_name(q{SOA}, $zone->name) ){
+            next;
+        }
+
+        my $p2 = $ns->query( $zone->name, q{MX}, { fallback => 0, usevc => 0 } );
+
+        if ( $p2 and $p2->tc ){
+            $p2 = $ns->query( $zone->name, q{MX}, { fallback => 0, usevc => 1 } );
+        }
```
matsduf, line 737: This should be the default querying when the specification sends a [default DNS query]. Would it make sense to implement the default query as a method? (I do not mean necessarily in this PR.)

[default DNS query]: https://github.com/zonemaster/zonemaster/blob/develop/docs/specifications/tests/DNSQueryAndResponseDefaults.md#default-setting-in-dns-query

tgreenx: I am not sure yet. I need more clarification about the behavior of LDNS, because now I see that there is a 3rd flag that can influence TCP behavior. It is unclear to me how it behaves when all of those flags are set/unset simultaneously.

https://github.com/zonemaster/zonemaster-engine/blob/master/lib/Zonemaster/Engine/Profile.pm#L610-L613

```
=head2 resolver.defaults.igntc

A boolean. If false, UDP queries that get responses with the C<TC>
flag set will be automatically resent over TCP. Default false.
```

tgreenx: @mattias-p maybe you have some info on this?

matsduf:
I have created a test zone xa with real MX on one name server and no MX on the other. I do not expect any Z09_TLD_EMAIL_DOMAIN, but I get four (it is a TLD):
```
$ zonemaster-cli --raw --show-testcase --test zone/zone09 --level INFO xa --ns zonemaster01-prd.iis.se --ns zonemaster05-prd.iis.se
   0.34 WARNING   ZONE09         Z09_INCONSISTENT_MX   
   0.34 INFO      ZONE09         Z09_NO_MX_FOUND   ns_ip_list=2001:67c:124c:7316::85;45.155.96.85
   0.34 INFO      ZONE09         Z09_MX_FOUND   ns_ip_list=2001:67c:124c:7316::81;45.155.96.81
   0.34 WARNING   ZONE09         Z09_TLD_EMAIL_DOMAIN   
   0.34 WARNING   ZONE09         Z09_TLD_EMAIL_DOMAIN   
   0.34 WARNING   ZONE09         Z09_TLD_EMAIL_DOMAIN   
   0.34 WARNING   ZONE09         Z09_TLD_EMAIL_DOMAIN  
```

matsduf:
A new test zone, xb. The two servers have different data in MX. I expecte two (not four) Z09_MX_DATA with "ns_ip_list" (not with "ns").
```
$ zonemaster-cli --raw --show-testcase --test zone/zone09 --level INFO --ns zonemaster01-prd.iis.se --ns zonemaster05-prd.iis.se xb
   0.34 WARNING   ZONE09         Z09_TLD_EMAIL_DOMAIN   
   0.34 WARNING   ZONE09         Z09_TLD_EMAIL_DOMAIN   
   0.34 WARNING   ZONE09         Z09_INCONSISTENT_MX_DATA   
   0.34 INFO      ZONE09         Z09_MX_DATA   domain_list=mahimahi.ripe.net.;molamola.ripe.net.; ns_ip_list=45.155.96.85
   0.34 INFO      ZONE09         Z09_MX_DATA   domain_list=mx1.iis.se.;mx2.iis.se.; ns_ip_list=45.155.96.81
   0.34 INFO      ZONE09         Z09_MX_DATA   domain_list=mx1.iis.se.;mx2.iis.se.; ns_ip_list=2001:67c:124c:7316::81
   0.34 INFO      ZONE09         Z09_MX_DATA   domain_list=mahimahi.ripe.net.;molamola.ripe.net.; ns_ip_list=2001:67c:124c:7316::85
```

matsduf:
New test zone, xc. One server has two normal MX and the the other has one normal and one Null-MX. I think the specification says that only the inconsistencies between the servers should be reported. If we should follow the implementation then the specification has to be updated.

Z09_NULL_MX_WITH_OTHER_MX is outputted twice, I only expected once.
```
$ zonemaster-cli --raw --show-testcase --test zone/zone09 --level INFO --ns zonemaster01-prd.iis.se --ns zonemaster05-prd.iis.se xc
   0.34 WARNING   ZONE09         Z09_TLD_EMAIL_DOMAIN   
   0.34 WARNING   ZONE09         Z09_NULL_MX_WITH_OTHER_MX   
   0.34 WARNING   ZONE09         Z09_NULL_MX_WITH_OTHER_MX   
   0.34 WARNING   ZONE09         Z09_TLD_EMAIL_DOMAIN   
   0.34 WARNING   ZONE09         Z09_INCONSISTENT_MX_DATA   
   0.34 INFO      ZONE09         Z09_MX_DATA   domain_list=.;molamola.ripe.net.; ns_ip_list=45.155.96.85
   0.34 INFO      ZONE09         Z09_MX_DATA   domain_list=.;molamola.ripe.net.; ns_ip_list=2001:67c:124c:7316::85
   0.34 INFO      ZONE09         Z09_MX_DATA   domain_list=mx1.iis.se.;mx2.iis.se.; ns_ip_list=45.155.96.81
   0.34 INFO      ZONE09         Z09_MX_DATA   domain_list=mx1.iis.se.;mx2.iis.se.; ns_ip_list=2001:67c:124c:7316::81
```

lib/Zonemaster/Engine/Test/Zone.pm
matsduf: Yes, it really looks like we want "resolver.defaults.igntc" to be set false by default, and then we should not need to worry. A test what would happen if the server responds with TC set over UDP and then not at all over TCP would be interesting.

matsduf:
Test zone xd. Both servers have the same MX, one correct Null-MX and one with preference 5. Too many messages.
```
$zonemaster-cli --raw --show-testcase --test zone/zone09 --level INFO --ns zonemaster01-prd.iis.se --ns zonemaster05-prd.iis.se xd
   0.34 WARNING   ZONE09         Z09_NULL_MX_WITH_OTHER_MX   
   0.34 WARNING   ZONE09         Z09_NULL_MX_WITH_OTHER_MX   
   0.34 NOTICE    ZONE09         Z09_NULL_MX_NON_ZERO_PREF   
   0.34 WARNING   ZONE09         Z09_NULL_MX_WITH_OTHER_MX   
   0.34 WARNING   ZONE09         Z09_NULL_MX_WITH_OTHER_MX   
   0.34 NOTICE    ZONE09         Z09_NULL_MX_NON_ZERO_PREF   
   0.34 WARNING   ZONE09         Z09_NULL_MX_WITH_OTHER_MX   
   0.34 WARNING   ZONE09         Z09_NULL_MX_WITH_OTHER_MX   
   0.34 NOTICE    ZONE09         Z09_NULL_MX_NON_ZERO_PREF   
   0.34 WARNING   ZONE09         Z09_NULL_MX_WITH_OTHER_MX   
   0.34 WARNING   ZONE09         Z09_NULL_MX_WITH_OTHER_MX   
   0.34 NOTICE    ZONE09         Z09_NULL_MX_NON_ZERO_PREF
```

matsduf (changes requested):
I have tested as reported in the comments. The test zones are available.

tgreenx:
> I have created a test zone xa with real MX on one name server and no MX on the other. I do not expect any Z09_TLD_EMAIL_DOMAIN, but I get four (it is a TLD):
> 
[ ... ]

The specification says:

> If Child Zone is a [TLD](https://github.com/matsduf/zonemaster/blob/155f654a77a36874a234d8a6d1c2122464dff42e/docs/specifications/tests/Zone-TP/zone09.md#terminology) with a [non-Null MX](https://github.com/matsduf/zonemaster/blob/155f654a77a36874a234d8a6d1c2122464dff42e/docs/specifications/tests/Zone-TP/zone09.md#terminology) then output [Z09_TLD_EMAIL_DOMAIN](https://github.com/matsduf/zonemaster/blob/155f654a77a36874a234d8a6d1c2122464dff42e/docs/specifications/tests/Zone-TP/zone09.md#Summary).

So, this zone is a TLD, and If I understand correctly, there are indeed 2 non-null MX records (`mx1.iis.se.` and `mx2.iis.se.`) from nameserver `zonemaster01-prd.iis.se` (which has an IPv4 and an IPv6 address, hence the 4 messages right now). What am I missing then? Shouldn't this message tag be outputted?

matsduf:
> The specification says:
> 
> > If Child Zone is a [TLD](https://github.com/matsduf/zonemaster/blob/155f654a77a36874a234d8a6d1c2122464dff42e/docs/specifications/tests/Zone-TP/zone09.md#terminology) with a [non-Null MX](https://github.com/matsduf/zonemaster/blob/155f654a77a36874a234d8a6d1c2122464dff42e/docs/specifications/tests/Zone-TP/zone09.md#terminology) then output [Z09_TLD_EMAIL_DOMAIN](https://github.com/matsduf/zonemaster/blob/155f654a77a36874a234d8a6d1c2122464dff42e/docs/specifications/tests/Zone-TP/zone09.md#Summary).

In the specification there is first
```
5.  For each name server IP in *Name Server IP* do:
```
which loops around the name servers and saves information into the sets. Then there is
```
10. If the *MX RRset* set is non-empty, then do:
     1. If the RRsets in *MX RRset* are not equal for all name servers then do:
     (...)
     2. Else do:
      (...)
        2. If *Child Zone* is a [TLD] with a [non-Null MX][Null MX] then
          output *[Z09_TLD_EMAIL_DOMAIN]*.
```
and that is not meant to say that the message should be outputted once per name server. Should I clarify the specification? There is no meaning of outputting that message more than once.

Secondly, as I read the specification, if the MX data are not equal in all servers then you come to `1. If the RRsets in *MX RRset* are not equal for all name servers then do:` and  never to `2. Else do:`.

When it comes to the second bullet one could have different opinions if the Z09_TLD_EMAIL_DOMAIN message should be outputted or not. If it should, then the specification should be updated, it not it should maybe be clarified?

> So, this zone is a TLD, and If I understand correctly, there are indeed 2 non-null MX records (`mx1.iis.se.` and `mx2.iis.se.`) from nameserver `zonemaster01-prd.iis.se` (which has an IPv4 and an IPv6 address, hence the 4 messages right now). What am I missing then? Shouldn't this message tag be outputted?

No, that was not my expectation, and definitely not more than once.

tgreenx:
I agree on the fact that having it outputted more than once was a mistake. It has been corrected in the latest commit (same goes for other message tags for other zones you mentioned earlier).

However I think that the specification do require an update then, because in our case, regarding zone `xa`:

1. It has two nameservers, only one of which (`zonemaster01-prd.iis.se`) has MX records (`mx1.iis.se.` and `mx2.iis.se.`).
2. Meaning that: `MX RRset` will contain those records (from nameserver `zonemaster01-prd.iis.se`) , while `No MX RRset` will contain the other nameserver (`zonemaster05-prd.iis.se`).
3. Thus: `MX RRset` is non-empty and `No MX RRset` isn't either. This validates [step 9](https://github.com/zonemaster/zonemaster/blob/develop/docs/specifications/tests/Zone-TP/zone09.md?plain=1#L190), and outputs the corresponding message tags. However this also validates [step 10](https://github.com/zonemaster/zonemaster/blob/develop/docs/specifications/tests/Zone-TP/zone09.md?plain=1#L197), as from there we look at the content of `MX RRset` exclusively (I think the problem arises from that). [ Step 10.1](https://github.com/zonemaster/zonemaster/blob/develop/docs/specifications/tests/Zone-TP/zone09.md?plain=1#L198) evaluates to false, 10.2.1 as well, thus we end up in step 10.2.2, where Z09_TLD_EMAIL_DOMAIN  is outputted (because zone `xa` is indeed a TLD).

matsduf:
> However I think that the specification do require an update then, because in our case, regarding zone `xa`:

I agree with your reading. There will be some extra messages, but only in case there is an inconsistency, so I think it is fine.

matsduf:
> I have created a test zone xa with real MX on one name server and no MX on the other. I do not expect any Z09_TLD_EMAIL_DOMAIN, but I get four (it is a TLD):

Now it looks fine:
```
   0.35 WARNING   ZONE09         Z09_INCONSISTENT_MX   
   0.35 INFO      ZONE09         Z09_NO_MX_FOUND   ns_ip_list=2001:67c:124c:7316::85;45.155.96.85
   0.35 INFO      ZONE09         Z09_MX_FOUND   ns_ip_list=2001:67c:124c:7316::81;45.155.96.81
   0.35 WARNING   ZONE09         Z09_TLD_EMAIL_DOMAIN   
```

matsduf:
> A new test zone, xb. The two servers have different data in MX. I expecte two (not four) Z09_MX_DATA with "ns_ip_list" (not with "ns").

I expected two Z09_MX_DATA, one per set, but I only get one (zone **xb**):
```
   0.34 WARNING   ZONE09         Z09_INCONSISTENT_MX_DATA   
   0.34 INFO      ZONE09         Z09_MX_DATA   domain_list=mx1.iis.se.;mx2.iis.se.;mahimahi.ripe.net.;molamola.ripe.net.;mx1.iis.se.;mx2.iis.se.;mahimahi.ripe.net.;molamola.ripe.net.; ns_ip_list=2001:67c:124c:7316::81;2001:67c:124c:7316::85;45.155.96.81;45.155.96.85
```

matsduf:
> New test zone, xc. One server has two normal MX and the the other has one normal and one Null-MX. I think the specification says that only the inconsistencies between the servers should be reported. If we should follow the implementation then the specification has to be updated.

Also here I expect two Z09_MX_DATA, one per set, but I only get one (zone **xc**):
```
   0.36 WARNING   ZONE09         Z09_INCONSISTENT_MX_DATA   
   0.37 INFO      ZONE09         Z09_MX_DATA   domain_list=mx2.iis.se.;mx1.iis.se.;molamola.ripe.net.;.;mx2.iis.se.;mx1.iis.se.;.;molamola.ripe.net.; ns_ip_list=2001:67c:124c:7316::81;2001:67c:124c:7316::85;45.155.96.81;45.155.96.85
```

matsduf:
> Test zone xd. Both servers have the same MX, one correct Null-MX and one with preference 5. Too many messages.

Now it looks fine (**xd**):
```
   0.34 WARNING   ZONE09         Z09_NULL_MX_WITH_OTHER_MX   
   0.34 NOTICE    ZONE09         Z09_NULL_MX_NON_ZERO_PREF   
```

matsduf:
I have also created the following test zones on the same server and report as expected:
* xe
* 2.0.192.in-addr.arpa
* 100.51.198.in-addr.arpa
* 113.0.203.in-addr.arpa
* normal-mx.xa
* no-mx.xa
* null-mx.xa

matsduf (changes requested):
* profile_example.json is not updated, see #1118
* You can use my test zones, and I'll document that they must be kept
* Besides the lack of separation between MX RRsets in Z09_MX_DATA I see no other issues.

tgreenx:
> * profile_example.json is not updated, see #1118
> 
>     * You can use my test zones, and I'll document that they must be kept
> 
>     * Besides the lack of separation between MX RRsets in Z09_MX_DATA I see no other issues.

All these should be solved now. I chose to add unitary tests for zones `xa` to `xe` for now, if you think the other zones you created deserve an unitary test let me know and I'll add them.

matsduf:
> * profile_example.json is not updated, see #1118

I created a PR #1120 that removes all message tags etc from profile_example.json, and if the PR is excepted, then the file should not be included here.

matsduf (approved):
It looks fine. Maybe there should be more scenarios in the unit tests to cover all messages, e.g. having root zones with normal MX, with Null MX and with no MX, respectively.

lib/Zonemaster/Engine/Test/Zone.pm
mattias-p: I agree with you @matsduf that we ought to have a method that encapsulates the default way we make DNS requests. And I'd like to expand on this idea too. For each behavior in the specifications that deviate from the default, the caller should be able to easily override that in the method parameters. (I guess possibly it could be a DNS request API that consists of more than one method.)

I agree with you @tgreenx too that this is a bit of a mess. I've tried to make sense of these flags before, but I'm afraid I've lost those notes. Though we should really deprecated and remove all properties from the Profile that correspond to LDNS flags that the DNS request method wants to set/clear.

MANIFEST
```diff
@@ -274,6 +274,16 @@ t/Test-syntax06-L.data
...
+t/Test-zone09-A.data
+t/Test-zone09-A.t
+t/Test-zone09-B.data
+t/Test-zone09-B.t
+t/Test-zone09-C.data
+t/Test-zone09-C.t
+t/Test-zone09-D.data
+t/Test-zone09-D.t
+t/Test-zone09-E.data
+t/Test-zone09-E.t
```
mattias-p, line 286: It's really nice that you continue the use of an established naming convention! However, the set of messages checked in those files is also the set of messages we need to keep stable because other people are relying on them. If we add new tests with this naming convention (and do nothing else), we might believe those messages are also covered by the stability guarantee, and that's an unnecessary maintenance burden.

I guess it would be a good idea to create a list in zonemaster/zonemaster that lists all the messages that we need to keep stable. But that's not enough. It should be visible somehow in the test files which messages are stable and which aren't. Otherwise it's really easy to forget that some messages are stable when making and reviewing updates. Maybe we could just rename the old Test-XXXX-X.t to Stable-XXXX-X.t? Maybe there are better ideas?

mattias-p (approved):
This looks good to me! Though I've opened up a couple of discussions in the comments that I don't think we should forget about. I guess I'll just create separate issues for those.

lib/Zonemaster/Engine/Test/Zone.pm
mattias-p: I created #1122 for to continue this discussion.

MANIFEST
mattias-p: I created #1123 so we don't need to stall this PR because of this discussion.

matsduf:
@tgreenx, I think this can be merged.

PR #1112, opened 2022-07-28
Update Installation.md

anandb-ripencc:
## Purpose

This PR updates installation instructions for RedHat-like systems.

## Context

Net::IP::XS is faster and less buggy than Net::IP.

## Changes

...

## How to test this PR

...

### DIFF
--- a/docs/Installation.md
+++ b/docs/Installation.md
@@ -65,7 +65,7 @@ Zonemaster::Engine, see the [declaration of prerequisites].
 4) Install packages from CPAN:
 
    ```sh
-   sudo cpanm Module::Install::XSUtil MooseX::Singleton Net::IP::XS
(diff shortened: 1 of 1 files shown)


matsduf:
It will not be enough to remove Net::IP from the list of dependencies in the installation ducument. It is specified in `Makefile.PL` so it will be pulled in anyway, now from CPAN instead of as a binary package.

https://github.com/zonemaster/zonemaster-engine/blob/develop/Makefile.PL#L31

anandb-ripencc:
I've submitted another PR to update Makefile.PL, and replace Net::IP with Net::IP::XS. So there will be no need to specify it an explicit module in the installation instructions. I will update the other PR.

There are no Net::IP::XS packages available for CentOS or Rocky. It's fine to install it from CPAN.

matsduf:
Replaced by #1130.

PR #1113, opened 2022-07-28
Update Makefile.PL

anandb-ripencc:
## Purpose

This PR replaces the dependency on Net::IP with Net::IP::XS.

## Context

Net::IP is buggy, and so it's better to use Net::IP::XS.

## Changes

...

## How to test this PR

...


matsduf:
Replaced by #1119

PR #1114, opened 2022-07-28
Update Dockerfile

anandb-ripencc:
## Purpose

This PR drops the Net::IP module from the docker installation.

## Context

Zonemaster::Engine specifies Net::IP::XS as a dependency, and it will be installed from CPAN.

## Changes

...

## How to test this PR

...

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -17,7 +17,6 @@ RUN apk add --no-cache \
     perl-lwp-protocol-https \
     perl-module-install \
     perl-moose \
-    perl-net-ip \
     perl-pod-coverage \
     perl-test-differences \
     perl-test-exception \


matsduf:
1. Have you also verified that the image will build with Net::IP::XS?
2. There is a reason why the dependencies are listen in the Docker file even if they are already in Makefile.PL. Compare with `Locale::TextDomain` which you will find in both Dockerfile and Makefile.PL.
3. Please add Net::IP::XS to the section for CPAN.

To test this PR Makefile.PL must be adjusted too.

anandb-ripencc:
Hi Mats,
What is the reason for explicitly installing `Local::TextDomain` if it's also listed in `Makefile.PL` as a dependency?

anandb-ripencc:
I built a docker image, and the build was successful. But I found one problem. The module `Zonemaster::Engine::Test` explicitly requires `Net::IP` and prevents `zonemaster-cli` and other parts of Zonemaster from running. That should either be removed or changed to `Net::IP::XS`, but I don't know what you prefer. In order to resolve it for now, I tried by changing `Net::IP` to `Net::IP::XS` and then `zonemaster-cli` runs without any errors.

anandb-ripencc:
I have one more comment. I built my docker image using modules *only* from CPAN. I did not use any alpine linux packages. I think this is a better way of managing a Zonemaster installation, instead of using some modules from the OS package system and some from CPAN.

matsduf:
> I built a docker image, and the build was successful. But I found one problem. The module `Zonemaster::Engine::Test` explicitly requires `Net::IP` and prevents `zonemaster-cli` and other parts of Zonemaster from running. That should either be removed or changed to `Net::IP::XS`, but I don't know what you prefer. In order to resolve it for now, I tried by changing `Net::IP` to `Net::IP::XS` and then `zonemaster-cli` runs without any errors.

A good catch.

I think we should have one pull request that changes the following parts:
1. dockerfile
2. Makefile.PL
3. Everywhere in the code where it says Net::IP it should say (or only say) Net::IP::XS
4. All unit tests should also be searched for Net::IP.

That PR could then be tested, and when everything passes and the Docker image passes, we come to step two, to verify that all OSs works with Net::IP::XS. And we think they do.

Step three is to update installation instructions and all other places that might refer to Net::IP.

anandb-ripencc:
Okay, let me create a new PR that combines the changes.

matsduf:
Replaced by #1119

PR #1117, opened 2022-08-02
Lower IPVx_DISABLED to DEBUG in Basic module

matsduf:
## Purpose

In all modules besides Basic the IPVx_DISABLED message is on DEBUG level. This PR alines Basic with the other module (before INFO). There is a need for some message higher level from Basic, and update zonemaster/zonemaster/pull/1090 of Basic04 adds such a message, which is less verbose than IPVx_DISABLED, especially after update #790.

## Changes

The default profile.

## How to test this PR

Run a test:
```
zonemaster-cli --raw --no-ipv6 --level INFO --test basic zonemaster.net
```
No PV6_DISABLED should be printed.
Run:
```
zonemaster-cli --raw --no-ipv6 --level DEBUG --test basic zonemaster.net | grep PV6_DISABLED
```
Messages should be printed on DEBUG level.


tgreenx (approved):

mattias-p (approved):
LGTM

PR #1119, opened 2022-08-03
Remove dependency on Net::IP, and use Net::IP::XS everywhere

### LINKED ISSUE #1107: Net::IP creates wrong IPv4 reverse name when last octet is "0"

anandb-ripencc:
## Purpose

This PR replaces Net::IP with Net::IP::XS. It doesn't get rid of old code that still checks for Net::IP. That can come in a later PR that does cleanup and also updates the installation documentation.

## Context

This replaces #1113 and #1114 and #1108

Resolves #1107 

## Changes

...

## How to test this PR

...


matsduf:
`lib/Zonemaster/Engine/Net/IP.pm` must also be updated. And `lib/Zonemaster/Engine/ASNLookup.pm` has some reference in documentation.

anandb-ripencc:
@matsduf `lib/Zonemaster/Engine/Net/IP.pm` doesn't use `Net::IP` if `Net::IP::XS` is present. Can't we handle updating it in a separate cleanup PR? There are several places from where `Net::IP` has to be removed, but this PR is only to switch from `Net::IP` to `Net::IP::XS` without also doing a cleanup.

matsduf (dismissed):

matsduf:
> @matsduf `lib/Zonemaster/Engine/Net/IP.pm` doesn't use `Net::IP` if `Net::IP::XS` is present. Can't we handle updating it in a separate cleanup PR? There are several places from where `Net::IP` has to be removed, but this PR is only to switch from `Net::IP` to `Net::IP::XS` without also doing a cleanup.

If this is to be the switch, all places should be handled. I have created a pull request for the file.

In  `lib/Zonemaster/Engine/ASNLookup.pm` it is documentation only to be updated.

Are there more places? I do not think so, but I can be wrong.

Dockerfile
```diff
@@ -17,7 +17,6 @@ RUN apk add --no-cache \
     perl-lwp-protocol-https \
     perl-module-install \
     perl-moose \
-    perl-net-ip \
```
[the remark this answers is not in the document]
anandb-ripencc: Why does `Net::IP::XS` need to be installed explicitly? It will be pulled in as a dependency when `Zonemaster::Engine` is installed, isn't it? In fact, that's what happens in the docker image when I install the engine using cpan.

Makefile.PL
```diff
@@ -28,7 +28,7 @@ requires 'Locale::TextDomain' => 1.20;
 requires 'Module::Find'       => 0.10;
 requires 'Moose'              => 2.0401;
 requires 'MooseX::Singleton'  => 0.30;
-requires 'Net::IP'            => 1.26;
+requires 'Net::IP::XS'        => 0;
```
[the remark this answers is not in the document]
anandb-ripencc: Done

anandb-ripencc:
Hi @matsduf I believe all the code changes are done now. There is only one unresolved discussion, about the Dockerfile. I do not believe it is necessary to explicitly install Net::IP::XS, because it will be pulled in as a dependency of the engine. If you disagree, I would like to understand why.

Dockerfile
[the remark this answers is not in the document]
matsduf: It is true, but that is not how Dockerfile is set up. It would be inconsistent not to list `Net::IP::XS` under cpanm, just after MooseX::Singelton. Please add it there.

matsduf:
As you can see all direct dependencies are listed in Dockerfile, either as binary package or as CPAN package. In that way there is a complete list. If Net::IP::XS is not included, then it would be inconsistent since the other dependencies of CPAN are listed there.

matsduf:
I made a mistake in my pull request to your pull request. I have added another commit to fix that. I think you have to accept to make it to be included.

Dockerfile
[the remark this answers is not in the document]
anandb-ripencc: Done

lib/Zonemaster/Engine/Net/IP.pm
```diff
@@ -70,37 +57,22 @@ sub version {
...
-    }
+    return Net::IP::XS::Error();
 }
 
 1;
 
 =head1 NAME
 
-Zonemaster::Engine::Net::IP - Net::IP/Net::IP::XS Wrapper (STILL EXPERIMENTAL)
+Zonemaster::Engine::Net::IP - Net::IP::XS Wrapper
```
mattias-p, line 75: We should probably mark this module as deprecated in this release. In a future release we can drop it entirely and just use Net::IP::XS directly instead.

matsduf: > We should probably mark this module as deprecated in this release. In a future release we can drop it entirely and just use Net::IP::XS directly instead.

If Net::IP is not dropped then #1108 should be merged to work-around the bug in Net::IP.

matsduf:
The changes in `lib/Zonemaster/Engine/Net/IP.pm` make the loading of some data files in the unit tests to crash, but not all of them. See #1121 for a test pull request where only that change is include plus a needed addition in `Makefile.PL`. Loading e.g. `t/Test-basic.data` works well, but not e.g. `t/Test-consistency05-A.data`. The error message, when loading fails, is
> Failed to build RR: Syntax error, could not parse the RR's rdata at `Zonemaster/LDNS/RR.pm` line 89.
* Are the affected data files dependent on Net::IP, or
* Is there a bug in the changed `lib/Zonemaster/Engine/Net/IP.pm`?

matsduf:
@mattias-p, @blacksponge and @pnax, can you please review? The PR passes the unit tests now.

lib/Zonemaster/Engine/Net/IP.pm
```diff
@@ -8,16 +8,13 @@ use warnings;
...
-    return q{Net::IP};
 };
 
 if ( $p_class ) {
     $p_class->import;
 }
 else {
-    die "Both Net::IP and Net::IP::XS missing?\n";
+    die "Net::IP::XS is not loaded\n";
 }
```
pnax, line 18: 2 questions here:
1. why can't we just have `use Net::IP::XS` here? I've tested it and tests are failing, it seems we can't import the module at compile time, any idea why?
2. can we get rid of the `$p_class` variable with something like:
    ```perl
    eval {
        require Net::IP::XS;
        Net::IP::XS->import;
    };
    if ( $@ ) {
        die "Net::IP::XS is not loaded\n";
    }

    Net::IP::XS->new( '192.0.2.1' );
    ```

mattias-p: 1. I very much believe we could.
2. A simple `use Net::IP::XS` would replace all of this code too.

The reason we had all this complexity in the first place was to make things work smoothly even if either one of Net::IP::XS or Net::IP was unavailable. Since we're dropping the dependency on Net::IP and promoting Net::IP::XS to a hard requirement we can just `use` that the normal way we don't need to do anything fancy at all.

mattias-p: On the other hand I'm not sure it's right to demand of @anandb-ripencc to do all the clean up. In my view we shouldn't be so hard on external contributors. I think fine if we the team clean things up to our own standards after merging an externally contributed PR that's functionally correct.

mattias-p: Oh, and the clean up should probably be to remove the entire Zonemaster::Engine::Net::IP module, and replace all its uses with direct calls into Net::IP::XS.

matsduf: In my first iteration I had
```
use Net::IP::XS;
my $p_class = q{Net::IP::XS};
```
but then many unit tests failed (see #1121).

`$p_class` is used in other places in the code.

The update in this PR is the minimal change to make it work with Net::IP::XS without Net::IP. If this PR is OK, then a clean-up could be done later.

matsduf: > Oh, and the clean up should probably be to remove the entire Zonemaster::Engine::Net::IP module, and replace all its uses with direct calls into Net::IP::XS.

I suggest that we take things in steps. Such a changes also affects Zonemaster-CLI and Zonemaster-Backend.

matsduf:
@mattias-p and @pnax, are you fine with the PR as-is? If so, please approve. The change in `Zonemaster::Engine::Net::IP` is the minimal change to make it work.

pnax (approved):
> On the other hand I'm not sure it's right to demand of @anandb-ripencc to do all the clean up. In my view we shouldn't be so hard on external contributors. I think fine if we the team clean things up to our own standards after merging an externally contributed PR that's functionally correct.

You have a very good point. Thanks @anandb-ripencc for providing this PR!

matsduf:
@mattias-p and @blacksponge, can you review too? Since I have worked on this PR I want another approval before I merge.

mattias-p (approved):
Looks good to me!
I created #1129 as a follow-up.

PR #1126, opened 2022-08-31
Define argument name "label" and "unicode_name"

matsduf:
## Purpose

Both "dlabel" and "label" are used as argument names, but none of them are defined. The former is used only once, the latter is clearer and used several times. With this PR "label" is defined in the documentation and "dlabel" is not used any more.

To make implementation consistent, "dlabel" is replaced by "label" in the affected Perl module and in the PO files.

In addition, the argument name "unicode_name" is defined (not yet in use in any code).

## Context

In https://github.com/zonemaster/zonemaster/pull/942#discussion_r951081481 it was noted that undefined "dlabel" was used in that PR. Instead of defining that, it will be replaced by "label" that this PR defines.

## Changes

* docs/logentry_args.md
* lib/Zonemaster/Engine/Test/Basic.pm
* share/da.po
* share/es.po
* share/fi.po
* share/fr.po
* share/nb.po
* share/sv.po

## How to test this PR

The main change is documentation change only. Review the change and compare with the implementation.

The other change affects the PO files. That can be tested by following the "[Instructions for translators]":

1. Update the PO files by running `./update-po da.po` and for all other PO files.
2. Check the arguments by running `../util/check-msg-args da.po` and for all other PO files.

[Instructions for translators]: https://github.com/zonemaster/zonemaster/blob/develop/docs/internal-documentation/maintenance/Instructions-for-translators.md

### DIFF
--- a/docs/logentry_args.md
+++ b/docs/logentry_args.md
@@ -71,7 +71,7 @@ and updated messages (*msgids* and *msgstr*).
 | rcode          | An RCODE Name                            | An RCODE Name (not numeric code) from [DNS RCODEs].                     |
 | rrtype         | A Resource Record TYPE Name              | A Resource Record TYPE Name (not numeric code) from [DNS RR TYPEs].     |
 | testcase       | A Zonemaster test case, or `all`         | A test case identifier.                                                 |
-
+| unicode_name   | Unicode name of a code point             | The name is a string in ASCII only and in upper case, e.g. "LATIN SMALL LETTER A"|
 
 ## Preliminary or proposed arguments
 
--- a/docs/logentry_args.md
+++ b/docs/logentry_args.md
@@ -57,6 +57,7 @@ and updated messages (*msgids* and *msgstr*).
 | ds_algo_mnemo  | Text                                     | The mnemonic of a [DS Digest algorithm].                                |
 | ds_algo_num    | Non-negative integer                     | The numeric value for a [DS Digest algorithm].                          |
 | keytag         | Non-negative integer                     | A keytag for a DNSKEY record or a keytag used in a DS or RRSIG record.  |
+| label          | Domain name label                        | A single label, i.e. the string between the dots, from a domain name.   |
 | mailtarget     | Domain name                              | The domain name of the mailserver in an MX RDATA.                       |
 | mailtarget_list| List of domain names                     | A list of name servers, as specified by "mailtarget", separated by ";". |
 | module         | A Zonemaster test module, or `all`       | The name of a Zonemaster test module.                                   |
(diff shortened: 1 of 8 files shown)


tgreenx (dismissed):
LGTM.

docs/logentry_args.md
```diff
@@ -91,7 +92,6 @@ defined *arguments*.
 || DNS packet size| The size in octets of a DNS packets.|
 || DNSKEY key length| The key length for a DNSKEY. The interpretation of this value various quite a bit with the algorithm. Be careful when using it for algorithms that aren't RSA-based.|
 || DNSSEC delegation verification failure reason| A somewhat human-readable reason why the delegation step between the tested zone and its parent is not secure.|
-| dlabel (?) | Domain name label| A single label from a domain name.|
 | dlength (?) | Domain name label length| The length of a domain name label.|
```
pnax, line 95: It seems the `dlength` is used, have you considered moving it to the `Defined arguments` table?

pnax (approved):
Beside a suggestion (not necessarily related to the PR), this looks ok.

docs/logentry_args.md
matsduf: I did consider that too, but I could not make up my mind if it should be "length" (since all lenght I know of is a positive integer) and defined as "Length of some entity" or "labellength"? If just "length" then it could be used in several cases.

In the same message ID there is also a "max". Should that be "max", "maxlength" or "maxlabellenght"?

There are several argument names to define.

pnax: I don't know either, but if we concatenate several words, I'd suggest to stick to snake case as already done for other labels. Hence it could be "label_length".

docs/logentry_args.md
```diff
@@ -70,7 +71,7 @@ and updated messages (*msgids* and *msgstr*).
 | rcode          | An RCODE Name                            | An RCODE Name (not numeric code) from [DNS RCODEs].                     |
 | rrtype         | A Resource Record TYPE Name              | A Resource Record TYPE Name (not numeric code) from [DNS RR TYPEs].     |
 | testcase       | A Zonemaster test case, or `all`         | A test case identifier.                                                 |
-
+| unicode_name   | Unicode name of a code point             | The name is a string in ASCII only and in upper case, e.g. "LATIN SMALL LETTER A"|
```
pnax, line 74: > In addition, the argument name "unicode_name" is defined (not yet in use in any code).

Do you have an example of how this argument would be used?

pnax: Found it in https://github.com/zonemaster/zonemaster/commit/8a14f99bdf364a2bf032cfb4cad20266896a9c71 (from https://github.com/zonemaster/zonemaster/pull/942)

pnax (approved):

tgreenx (approved):

PR #1128, opened 2022-09-05
Fix message to take equality into consideration

pnax:
## Purpose

In Zone02 (resp. Zone04), when the SOA 'refresh' (resp. 'retry') value equals the minimum recommended value, the message can be disconcerting:

```
SOA 'refresh' value (14400) is higher than the minimum recommended value (14400). 
SOA 'retry' value (3600) is more than the minimum recommended value (3600). 
```
This PR updates the msgid to include the equality case in the message.

## Context

Addresses https://github.com/zonemaster/zonemaster/issues/1092

## Changes

Update msgid in Zone.pm.

## How to test this PR

Running a test where the SOA 'refresh' (resp. 'retry') value equal the minimum recommended one and see look for the updated messages:
```
zonemaster-cli --test zone/zone02 --test zone/zone04 --level INFO elysee.fr
Seconds Level     Message
======= ========= =======
   0.00 INFO      Using version v4.5.1 of the Zonemaster engine.
   1.86 INFO      SOA 'refresh' value (14400) is higher than the minimum recommended value (14400).
   0.00 INFO      Using version v4.5.1 of the Zonemaster engine.
   0.52 INFO      SOA 'retry' value (3600) is more than the minimum recommended value (3600).
```

### DIFF
--- a/lib/Zonemaster/Engine/Test/Zone.pm
+++ b/lib/Zonemaster/Engine/Test/Zone.pm
@@ -180,7 +180,7 @@ Readonly my %TAG_DESCRIPTIONS => (
     },
     RETRY_MINIMUM_VALUE_OK => sub {
         __x    # ZONE:RETRY_MINIMUM_VALUE_OK
-          'SOA \'retry\' value ({retry}) is more than the minimum recommended value ({required_retry}).', @_;
+          'SOA \'retry\' value ({retry}) is more or equal than the minimum recommended value ({required_retry}).', @_;
     },
     MNAME_NO_RESPONSE => sub {
         __x    # ZONE:MNAME_NO_RESPONSE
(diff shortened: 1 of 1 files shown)


tgreenx (commented):
Minor suggestion but can we harmonize the two messages? They each use a synonym ('more' / 'higher').
Specifications are not updated to the latest Template01 format so they don't require a change first (message tags are not specified). However translation files do.

tgreenx (approved):
As discussed, this rewording ("at least equal to") sounds the best to express the intent and meaning.

lib/Zonemaster/Engine/Test/Zone.pm
```diff
@@ -180,7 +180,7 @@ Readonly my %TAG_DESCRIPTIONS => (
     },
     RETRY_MINIMUM_VALUE_OK => sub {
         __x    # ZONE:RETRY_MINIMUM_VALUE_OK
-          'SOA \'retry\' value ({retry}) is more than the minimum recommended value ({required_retry}).', @_;
+          'SOA \'retry\' value ({retry}) is at least equal to the minimum recommended value ({required_retry}).', @_;
```
matsduf, line 183: I think "equal to or higher than" is easier to read and understand.

matsduf: This message should not be higher than INFO, in fact it makes more sense if it is a DEBUG message.

PR #1132, opened 2022-09-28
Make root hints file configurable

### LINKED ISSUE #850: Break root hints data out from Perl module
Root hints data is stored as a [data section] in `Zonemaster/Engine/Recursor.pm`. For some use cases it can be useful to be able to change that to private root hints data:

* DNS training using a private DNS tree including root and TLDs.
* Zonemaster unit tests making it possible to set up both parent and child in a private tree including TLDs.

To make these use cases possible, put the root hints data was a separate data file. Also make it possible to override that file with a private hints file without updating the perl code.

[data section]: https://github.com/zonemaster/zonemaster-engine/blob/5124989b8a05263b4445aaa181478e0b1f0737a2/lib/Zonemaster/Engine/Recursor.pm#L379-L471

matsduf:
## Purpose

In current code the root hints are hard coded. This PR opens for use cases where private root must be used, see #850. The driving use case for this PR is the possibility to use it for unit tests.

## Context

Fixes #850.

## Changes

Default root hints file is found in shared dist directory, and is used in the normal case. If a root hints file is found in /etc/zonemaster or /usr/local/etc/zonemaster that is used instead. The file name must be `root-hints.json`. Highest priority has the environment variable `ZONEMASTER_ENGINE_ROOT_HINTS_FILE` that can point at any file name, and that file will be used as the root hints file.

## How to test this PR

1. Run a default configuration and validate that nothing breaks.
2. Create a private root zone file and a zone file for some non-existent TLD, e.g. .xa
3. In the root zone, delegate to .xa and load the zones.
4. Create a root hints file in /etc/zonemaster or /usr/local/etc/zonemaster that points at the private root.
5. Run `zonemaster-cli xa --level INFO` and verify that .xa is reported as an existing domain.
6. Rename the root hints file and use `ZONEMASTER_ENGINE_ROOT_HINTS_FILE` and verify that you can still test the .xa domain.

### DIFF
--- a/lib/Zonemaster/Engine/Recursor.pm
+++ b/lib/Zonemaster/Engine/Recursor.pm
@@ -22,7 +22,7 @@ our %_fake_addresses_cache;
 
 sub get_default_path {
     state $path =
-        $ENV{ZONEMASTER_ENGINE_ROOT_HINTS_FILE}              ? $ENV{ZONEMASTER_ENGINE_ROOT_HINTS_FILE}
+        length( $ENV{ZONEMASTER_ENGINE_ROOT_HINTS_FILE} )    ? $ENV{ZONEMASTER_ENGINE_ROOT_HINTS_FILE}
       : -e '/etc/zonemaster/root-hints.json'                 ? '/etc/zonemaster/root-hints.json'
       : -e '/usr/local/etc/zonemaster/root-hints.json'       ? '/usr/local/etc/zonemaster/root-hints.json'
       :                                                        eval { dist_file( 'Zonemaster-Engine', 'root-hints.json' ) };
@@ -460,7 +460,7 @@ Class method to empty the cache of responses to recursive queries (but not the o
 =item root_servers()
 
 Returns a list of ns objects representing the root servers. The list of root servers is found in an external
-file..
+file.
 
 =back
 
--- a/lib/Zonemaster/Engine/Recursor.pm
+++ b/lib/Zonemaster/Engine/Recursor.pm
@@ -7,6 +7,8 @@ use warnings;
 
 use Carp;
 use Class::Accessor "antlers";
+use File::ShareDir qw[dist_file];
+use File::Slurp qw( read_file );
 use JSON::PP;
 use Zonemaster::Engine::Util;
 use Zonemaster::Engine::Net::IP;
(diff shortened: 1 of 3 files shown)


lib/Zonemaster/Engine/Recursor.pm
```diff
@@ -17,10 +19,14 @@ my $seed_data;
...
 our %_fake_addresses_cache;
 
-{
-    local $/;
-    my $json = <DATA>;
-    $seed_data = decode_json $json;
+
+sub get_default_path {
+    state $path =
+        $ENV{ZONEMASTER_ENGINE_ROOT_HINTS_FILE}              ? $ENV{ZONEMASTER_ENGINE_ROOT_HINTS_FILE}
```
mvw-afnic, line 25: Suppose I want to load a root hints file called `0` (unlikely, but still): if I set `ZONEMASTER_ENGINE_ROOT_HINTS_FILE=0` in my environment, that variable won’t be taken into account because "0" is coerced to boolean false.

Using `length $ENV{ZONEMASTER_ENGINE_ROOT_HINTS_FILE}` should work as a test for both existence and non-emptiness of the environment variable.

lib/Zonemaster/Engine/Recursor.pm
```diff
@@ -409,118 +459,9 @@ Class method to empty the cache of responses to recursive queries (but not the o
 
 =item root_servers()
 
-Returns a list of ns objects representing the root servers. The list of root servers is hardcoded into this module.
+Returns a list of ns objects representing the root servers. The list of root servers is found in an external
+file..
```
mvw-afnic, line 463: There’s an extraneous full stop at the end of this line.

mvw-afnic (changes requested):

lib/Zonemaster/Engine/Recursor.pm
matsduf: I think we had some similar case when the zone name was `0`. I will fix that. I copied the solution from Backend, where loading backend configuration file has the same issue, but I guess it is more theoretical since it is very unlikely that you would name the file `0`.

mvw-afnic (approved):

lib/Zonemaster/Engine/Recursor.pm
```diff
@@ -368,6 +383,41 @@ delegations (pre-publication tests).
...
+the same value is returned no matter if environment variables or the file system
+have changed.
+
+The following checks are made in order:
+
+=over 4
+
+=item $ZONEMASTER_ENGINE_ROOT_HINTS_FILE
+
+If this environment variable is set ot a truthy value, that path is returned.
```
pnax, line 402: typo: "ot" -> "at"

lib/Zonemaster/Engine/Recursor.pm
```diff
@@ -368,6 +383,41 @@ delegations (pre-publication tests).
...
+If a file exists at such a path, it is returned.
+
+=item DIST_DIR/root-hints.json
+
+If a file exists at this path, it is returned.
+DIST_DIR is wherever File::ShareDir installs the Zonemaster-Engine dist.
+
+=back
+
+=head2 Other mothods
```
pnax, line 419: typo: "mothods" -> "methods"

Did you consider promoting all following methods to `=head2` instead?

lib/Zonemaster/Engine/Recursor.pm
matsduf: Re-written.

lib/Zonemaster/Engine/Recursor.pm
matsduf: I did, but took the easy path. Updated with `=head2` now.

pnax (approved):
Nice feature!

matsduf:
See #1134 for a more complete solution.

PR #1147, opened 2022-11-16
Fix EDNS behavior for queries

tgreenx:
## Purpose

This PR fixes bugs regarding EDNS behavior in Engine.

## Context

While implementing #1034, I noticed odd behaviors when playing around EDNS queries and responses.

Also, note that Test Cases nameserver10 to 14 (14 being NYI, see #1033) don't have unitary tests which can partly explain why such an issue was not seen before. I created issue #1146 for that.

## Changes

- EDNS handling for queries has been updated - the main issue was that setting only the version of EDNS would not enable it as an EDNS query. Some refactoring as well in that regard.
- Default EDNS query packet size has been updated to 512 in profile.pm, as per the specification ([DNSQueryAndResponseDefaults.md](https://github.com/zonemaster/zonemaster/blob/master/docs/specifications/tests/DNSQueryAndResponseDefaults.md#default-setting-in-edns-query)).
- Some test cases in Test/Nameserver.pm were not checking if the OPT record was present in the responses to EDNS queries (default behavior as per the same specification). 
- Unitary tests updated accordingly.

## How to test this PR

Tests should pass.


matsduf:
> @matsduf I included this change initally but now I removed it because I am not entirely sure if it is needed. Looking at specifications, Nameserver02 is testing whether OPT records are present in responses of EDNS queries ("`If an OPT record is present in a received request, compliant responders MUST include an OPT record in their respective responses`"). This can be seen in the implementation also (using `has_edns()`). Subsequent specifications, for example [Nameserver10](https://github.com/zonemaster/zonemaster/blob/master/docs/specifications/tests/Nameserver-TP/nameserver10.md), say "`Issues covered by [...]` [Nameserver02](https://github.com/zonemaster/zonemaster/blob/master/docs/specifications/tests/Nameserver-TP/nameserver02.md)` (basic EDNS issues) will not result in messages from this test case.`" and so in implementations the OPT record is not checked in responses.

Yes, in [Nameserver10](https://github.com/zonemaster/zonemaster/blob/master/docs/specifications/tests/Nameserver-TP/nameserver10.md) if a server does not respond properly to EDNS vers 0 queries they should just be ignored.

> But default behavior (as specified in [DNSQueryAndResponseDefaults.md](https://github.com/zonemaster/zonemaster/blob/master/docs/specifications/tests/DNSQueryAndResponseDefaults.md#default-setting-in-edns-query) should apply to all specifications, shouldn't it? Should that check be included for all Test Cases checking for different EDNS parameters?

The [DNSQueryAndResponseDefaults.md](https://github.com/zonemaster/zonemaster/blob/master/docs/specifications/tests/DNSQueryAndResponseDefaults.md#default-setting-in-edns-query) applies to all specifications that have been updated to refer to it.

The idea is that we should not report the same problem i several test cases. If a server cannot handle EDNS at all, then only [Nameserver02](https://github.com/zonemaster/zonemaster/blob/master/docs/specifications/tests/Nameserver-TP/nameserver02.md) should report that.

tgreenx:
> Do I understand it correctly that that some queries were mean to be EDNS queries but were sent as non-EDNS queries?

Yes that's right, in Nameserver10.

lib/Zonemaster/Engine/Nameserver.pm
```diff
@@ -298,22 +298,9 @@ sub query {
     my $p;
     my $md5 = Digest::MD5->new;
     my $edns_special_case = 0;
```
matsduf, line 300: Can this control variable "$edns_special_case" defined and explained.

lib/Zonemaster/Engine/Nameserver.pm
```diff
@@ -298,22 +298,9 @@ sub query {
...
+
     if ( defined $href->{edns_details} ) {
```
matsduf, line 302: Where does "edns_details" come from? Can it be defined and explained?

lib/Zonemaster/Engine/Nameserver.pm
```diff
@@ -455,37 +446,35 @@ sub _query {
         );
     }
     else {
-        if ( $edns_special_case ) {
+        if ( $edns_special_case and defined $href->{edns_details} ) {
```
matsduf, line 454: Is it possible for "$edns_special_case" to be true and "$href->{edns_details}" be undefined? See lines 302-304?

t/Test-nameserver.t
```diff
@@ -73,9 +73,6 @@ zone_gives( 'nameserver01', $zone, [q{NO_RECURSOR}] );
 zone_gives_not( 'nameserver01', $zone, [q{IS_A_RECURSOR}] );
 
 # nameserver02
-$zone = Zonemaster::Engine->zone( 'flagday.rootcanary.net' );
-zone_gives( 'nameserver02', $zone, ['BREAKS_ON_EDNS']);
```
matsduf, line 77: Why is this unit test removed? It works as expected (tested with the code of this PR):
```
# zonemaster-cli --show-testcase --level info --test nameserver/nameserver02 flagday.rootcanary.net --raw
  21.68 ERROR     NAMESERVER02   BREAKS_ON_EDNS   domain="flagday.rootcanary.net"; ns=flagday.rootcanary.net/2a04:b900:0:100::9
```

matsduf (changes requested):

matsduf (commented):
In general, I have no problem with this PR, but we do lack good test zones.

t/Test-nameserver.t
tgreenx: Not sure anymore, it works correctly on my end now too but it wasn't the case a few days ago. I'll add it back.

matsduf: It emphasizes the need for internal test zones, i.e. the work in https://github.com/zonemaster/zonemaster/pull/1113

lib/Zonemaster/Engine/Nameserver.pm
tgreenx: No, but it was here still as a safeguard. Anyway now I refactored and removed control variable `$edns_special_case`.

lib/Zonemaster/Engine/Nameserver.pm
tgreenx: In fact, after the initial refactor it is no longer needed so I refactored again. See latest commit https://github.com/zonemaster/zonemaster-engine/pull/1147/commits/b85fe49426de0833c2a4a8479acc540277eff9ff

lib/Zonemaster/Engine/Nameserver.pm
tgreenx: This is a parameter of a query, used in Test Cases implementations. It enables the query as an EDNS query. See for example [here](https://github.com/zonemaster/zonemaster-engine/blob/develop/lib/Zonemaster/Engine/Test/Nameserver.pm#L1022): `my $p = $ns->query( $zone->name, q{SOA}, { edns_details => { version => 0 } } );`

EDNS queries will now be correctly enabled using any `edns_details` parameters ("version", "z", "do", "udp_size", etc). The main issue before this PR was that using `edns_details->version = 0` would not enable the query as an EDNS query. See lines 302 and 460.
`edns_details->udp_size` and `edns_size` have the same effect, but `udp_size` takes precedence if both are provided.
Setting `edns_size` exclusively will have no effect (i.e. not enable the query as an EDNS query) unless a parameter of `edns_details` is also specified.

matsduf (approved):

lib/Zonemaster/Engine/Nameserver.pm
```diff
@@ -297,44 +297,28 @@ sub query {
...
 
     $md5->add( q{NAME}    , $name );
     $md5->add( q{TYPE}    , "\U$type" );
     $md5->add( q{CLASS}   , "\U$class" );
     $md5->add( q{DNSSEC}  , $dnssec );
     $md5->add( q{USEVC}   , $usevc );
     $md5->add( q{RECURSE} , $recurse );
-    if ( $edns_special_case ) {
+    
+    if ( defined $href->{edns_details} ) {
```
mattias-p, line 308: It seems the `edns_details` parameter is completely undocumented in the API documentation for query(). I guess that means this PR doesn't break any promises with regard to behavior. That's always convenient, I guess. But this PR does change the implemented semantics for that parameter. Wouldn't this be a good time to document the new semantics?

lib/Zonemaster/Engine/Nameserver.pm
```diff
@@ -297,44 +297,28 @@ sub query {
...
         $md5->add( q{EDNS_VERSION}        , $href->{edns_details}{version} ? $href->{edns_details}{version} : 0 );
         $md5->add( q{EDNS_Z}              , $href->{edns_details}{z} ? $href->{edns_details}{z} : 0 );
         $md5->add( q{EDNS_EXTENDED_RCODE} , $href->{edns_details}{extended_rcode} ? $href->{edns_details}{extended_rcode} : 0 );
         $md5->add( q{EDNS_DATA}           , $href->{edns_details}{data} ? $href->{edns_details}{data} : q{} );
-        $md5->add( q{EDNS_UDP_SIZE}       , $href->{edns_details}{udp_size} ? $href->{edns_details}{udp_size} : 0 );
+        $md5->add( q{EDNS_UDP_SIZE}       , $href->{edns_details}{udp_size} ? $href->{edns_details}{udp_size} : $edns_size );
```
mattias-p, line 313: I realize this is not of your making, but don't these conditions look a bit buggy? E.g. if you try to set `$href->{edns_details}{data}` to `0`, `EDNS_DATA` would end up being set to the empty string. I would have expected the conditions to use `defined`.

Since we're already changing the semantics of `edns_details` maybe fix this too at the same time?

lib/Zonemaster/Engine/Nameserver.pm
```diff
@@ -429,11 +413,20 @@ sub _query {
...
-    if ( defined $href->{edns_details} and $href->{edns_details}{udp_size} ) {
-        $flags{q{edns_size}} = $href->{edns_details}{udp_size};
+    
+    if ( defined $href->{edns_details} ) {
+        if ( defined $href->{edns_details}{udp_size} ) {
+            $flags{q{edns_size}} = $href->{edns_details}{udp_size};
+        }
+        else {
+            $flags{q{edns_size}} = $href->{q{edns_size}} // Zonemaster::Engine::Profile->effective->get( q{resolver.defaults.edns_size} );
+        }
```
mattias-p, line 424: Couldn't this be shortened like so?

```perl
$flags{q{edns_size}} = $href->{edns_details}{udp_size}
  // $href->{q{edns_size}}
  // Zonemaster::Engine::Profile->effective->get( q{resolver.defaults.edns_size} );
```

To me that's more readable.

mattias-p (commented):
I haven't looked at all the calls to query() to understand the exact consequences of this change, but I do like the approach.

lib/Zonemaster/Engine/Nameserver.pm
tgreenx: Yes, good catch! Updated.

tgreenx:
@mattias-p @matsduf please re-review.

lib/Zonemaster/Engine/Nameserver.pm
```diff
@@ -833,6 +819,21 @@ If set to true, incoming response packets with the TC flag set fall back to EDNS
...
+Set the EDNS0 UDP maximum size.
+
+Does not enable on its own the query to be an EDNS query. Value discarded if 'edns_details' is not provided.
+Value overridden by 'edns_details->udp_size' (if also given). More details below.
+
+=item edns_details
+
+A hash. The currently supported keys for this flag are 'version', 'z', 'do', 'extended_rcode', 'udp_size' and 'data'.
+If any of these are set the query will be an EDNS query, constructed with the values (if any) for each of these keys.
+See L<Zonemaster::LDNS::Packet> for more details (key names prefixed with 'edns_').
```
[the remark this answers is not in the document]
tgreenx: > `edns_udp_size` is not documented or used in Zonemaster::LDNS::Packet.

Indeed, it is called 'edns_size' instead. Another exception is 'extended_rcode' which is called 'edns_rcode' in Zonemaster::LDNS::Packet. I updated the documentation to account for those two exceptions.

> There is also the `do` flag that can be set directly in $flagref to query(). That setting will also enable EDNS, won't it? And what is the relationship between $flagreg->{do} and $flagref->{edns_details}{do}?

The profile flag is called 'dnssec'. I overlooked this one; they have the same effect as well, and now I have given explicit precedence to 'edns_details->do' when both are defined. I updated the documentation as well.

> In LDNS.xs it says about edns_data that "It can only deal with option codes with OPTION-LENGTH=0 (see 6.1.2 section of RFC 6891) which means OPTION-DATA is always empty." Does that mean that Zonemaster can neither read nor write data for an option code?

Not sure right now, I will have to investigate more on that.

t/Test-dnssec.t
```diff
@@ -218,7 +218,7 @@ zone_gives( 'dnssec08', $zone, [qw{DS08_DNSKEY_RRSIG_EXPIRED DS08_RRSIG_NOT_VALI
 zone_gives_not( 'dnssec08', $zone, [qw{DS08_ALGO_NOT_SUPPORTED_BY_ZM DS08_DNSKEY_RRSIG_NOT_YET_VALID DS08_MISSING_RRSIG_IN_RESPONSE DS08_NO_MATCHING_DNSKEY}] );
 
 $zone = Zonemaster::Engine->zone( 'dnssec08-dnskey-signature-not-ok.zut-root.rd.nic.fr' );
-zone_gives( 'dnssec08', $zone, [qw{DS08_RRSIG_NOT_VALID_BY_DNSKEY}] );
+#zone_gives( 'dnssec08', $zone, [qw{DS08_RRSIG_NOT_VALID_BY_DNSKEY}] );
```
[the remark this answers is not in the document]
tgreenx: Yes, done.

tgreenx (commented):
Tests are failing, I am not sure why exactly. My latest change on the cache shouldn't impact all of them. Investigating...

tgreenx:
Issue has been identified (checking with `defined()` on a hash key of an undefined hash). An alternative was possible (using `exists()`) but this would overcomplicate the code for no reason. Instead, I went with a change in line with the previous changes of this PR. This modifies the md5 digest created for each query so this requires an update to all unitary tests data. @blacksponge will take care of that.

tgreenx:
After reconsideration, this proves too much of a hassle to update all unitary tests data files. Another good reason to move to thethe upcoming https://github.com/zonemaster/zonemaster/pull/1113. I went with the less elegant solution, that doesn't impact the query cache.

tgreenx:
@matsduf @mattias-p please re-review

lib/Zonemaster/Engine/Nameserver.pm
```diff
@@ -297,44 +299,35 @@ sub query {
...
     $md5->add( q{TYPE}    , "\U$type" );
     $md5->add( q{CLASS}   , "\U$class" );
-    $md5->add( q{DNSSEC}  , $dnssec );
+
+    if ( defined $href->{edns_details} and defined $href->{edns_details}{do} ) {
+        $md5->add( q{DNSSEC} , $href->{edns_details}{do} );
+    }
+    else {
+        $md5->add( q{DNSSEC} , $dnssec );
+    }
```
matsduf, line 312: It takes more space but it is well readable.

t/Test-dnssec.t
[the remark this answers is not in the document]
matsduf: You write "Yes, done". What was done? I the test is just commented it could as well be removed. If there is a TODO we will be reminded. An issue could also do.

It also applies for other tests below.

tgreenx:
@matsduf @mattias-p please re-review

matsduf (dismissed):

matsduf:
Please rebase, and I will re-approve.

lib/Zonemaster/Engine/Nameserver.pm
```diff
@@ -842,6 +840,24 @@ If set to true, incoming response packets with the TC flag set fall back to EDNS
...
+
+=item edns_details
+
+A hash. The currently supported keys for this flag are 'version', 'z', 'do', 'extended_rcode', 'udp_size' and 'data'.
+If any of these are set the query will be an EDNS query, constructed with the values (if any) for each of these keys.
+See L<Zonemaster::LDNS::Packet> for more details (key names prefixed with 'edns_'; note that 'udp_size' is abbreviated
+to 'size' and 'extended_rcode' to 'rcode').
+Note that a profile flag 'edns_size' also exists (see above) and has the same effect as 'edns_details->udp_size', although the value of the 
+latter will take precedence if both are given. Solely setting 'edns_size' will not enable the query as an EDNS query.
+Similarly, note that a profile flag 'dnssec' also exists (see above) and has the same effect as 'edns_details->do', although the value of the
```
mattias-p, line 858: Maybe add an empty line before this one to make it easier to navigate?

I feel it's confusing to speak of these argument flags as profile flags. Maybe change the wording? E.g. by referring to argument flags or to $flagref. (This goes for edns_size a couple of lines up too.)

Maybe also add that if this flag is omitted, a default value is taken from the profile property 'resolver.defaults.dnssec'?

lib/Zonemaster/Engine/Nameserver.pm
```diff
@@ -842,6 +840,24 @@ If set to true, incoming response packets with the TC flag set fall back to EDNS
...
+
+Does not enable on its own the query to be an EDNS query. Value discarded if 'edns_details' is not provided.
+Value overridden by 'edns_details->udp_size' (if also given). More details in 'edns_details' below.
+
+=item edns_details
+
+A hash. The currently supported keys for this flag are 'version', 'z', 'do', 'extended_rcode', 'udp_size' and 'data'.
+If any of these are set the query will be an EDNS query, constructed with the values (if any) for each of these keys.
+See L<Zonemaster::LDNS::Packet> for more details (key names prefixed with 'edns_'; note that 'udp_size' is abbreviated
+to 'size' and 'extended_rcode' to 'rcode').
```
mattias-p, line 855: In addition to specifying the meaning of the different flags as you do here, maybe also specify what default values they get if they're omitted from the arguments?

mattias-p:
It seems without this PR we get wrong behaviors for some EDNS queries, but with this PR we get some other weird behaviors in DNSSEC08, DNSSEC10, DNSSEC13 and DNSSEC17. Maybe we should push it to 2023.1?

tgreenx (commented):
Rebased on develop and addressed comments. Please re-review.

tgreenx:
Unitary tests will need another pass as my machine didn't have an IPv6 connectivity

matsduf:
>     * `edns_details->size` and `edns_size` have the same effect - `edns_details->size` takes precedence if both are provided. The first is a query parameter used in Test Cases, while the latter is a profile parameter (for queries, too).

There is really no reason to have `edns_size` as a profile parameter since the EDNS settings must be under the control of the the test case.

>     * Same goes for `edns_details->do` and `dnssec` - `edns_details->do` takes precedence.

The same thing here, the DO setting must be under the control of the test case.

I suggest that this PR **deprecates** those two profile settings. Keep them, but make it impossible to change the default values ("dnssec" : false, "edns_size" : 0) via profile setting. Then remove them completely in v2023.2.

lib/Zonemaster/Engine/Profile.pm
```diff
@@ -596,6 +596,10 @@ Default 3.
 
 A boolean. If true, sets the DO flag in queries. Default false.
 
+=head2 resolver.defaults.edns_size
+
+An integer. The EDNS0 UDP size used in EDNS queries. Default 512.
```
matsduf, line 601: In the default profile it is set to 0, but I guess that is in practice the same as 512 since 512 is the lowest possible value. However, this setting should rather be removed since it makes no sense to set it to another value. The value must be under control of the test case.

https://github.com/zonemaster/zonemaster-engine/blob/develop/share/profile.json#L18

lib/Zonemaster/Engine/Constants.pm
```diff
@@ -282,6 +284,12 @@ C<$UDP_PAYLOAD_LIMIT>
 
 =item *
 
+C<$UDP_EDNS_PAYLOAD_LIMIT>
+
+The EDNS0 UDP maximum size.
```
marc-vanderwal, line 289: I don’t quite get the difference between `UDP_PAYLOAD_LIMIT`, `UDP_EDNS_PAYLOAD_LIMIT` and `UDP_COMMON_EDNS_LIMIT` from reading this file. Can you elaborate, and document those two other constants as well?

marc-vanderwal (commented):
Just one minor remark about documentation.

lib/Zonemaster/Engine/Constants.pm
matsduf: From the value, `UDP_EDNS_PAYLOAD_LIMIT` is the lowest value of UDP payload in EDNS. If lower, it must be treated as 512. `UDP_EDNS_PAYLOAD_MIN` is maybe a better name? There is also a maximum value. Should that be considered?

matsduf: `UDP_PAYLOAD_LIMIT` is not used to check incoming DNS packets, but only used in DELEGATION03 to check if a constructed packet with the assumed referral for the domain name would exceed 512 Bytes or not.

`UDP_COMMON_EDNS_LIMIT` is used to generate a message out of test case if the packet data is larger than `UDP_COMMON_EDNS_LIMIT` (as defined above) which feels quite arbitrarily. There are two sizes that could be valid to react on:
* Larger than the value set in the query. That is not permitted.
* Larger than 1232 because the greatly increases the risk of fragmented UDP.

The new `UDP_EDNS_PAYLOAD_LIMIT` is used as the UDP size setting in the default query. I suggest the name `UDP_EDNS_QUERY_DEFAULT` or something like that.

lib/Zonemaster/Engine/Nameserver.pm
```diff
@@ -808,11 +801,14 @@ Send the query via TCP (only).
...
 
-The retransmission interval
+The retransmission interval.
 
 =item dnssec
 
-Set the DO flag in the query.
+Set the DO flag in the query. Defaults to false.
+
+Enables the query to be an EDNS query if set to true.
```
matsduf, line 810: Unclear. Should it say that EDNS is automatically enabled if the DO flag is set?

matsduf:
@tgreenx, what is the status of this PR?

lib/Zonemaster/Engine/Constants.pm
```diff
@@ -282,6 +284,12 @@ C<$UDP_PAYLOAD_LIMIT>
 
 =item *
 
+C<$UDP_EDNS_QUERY_DEFAULT>
+
+The EDNS0 UDP maximum size.
```
matsduf, line 289: $UDP_EDNS_QUERY_DEFAULT is the default setting in EDNS in queries from Zonemaster. 512 is also the lower limit of what you may announce in EDNS.

marc-vanderwal (approved):
Looks good to me.

It’s a shame we cannot investigate the failing unit tests further. It does seem, however, that the test zones themselves are at fault and not the code, because I ran `ZONEMASTER_RECORD=1 prove -l t/Test-dnssec.t` with an unrelated branch and got even more errors.

matsduf (approved):

PR #1178, opened 2023-01-12
Fix bug in Zone01

### LINKED ISSUE #1175: Zone01 crashes if two nameservers serve different SOA serials
Consider the following test zone, served by two nameservers but with different SOA serials:

```zone
$TTL 3600
$ORIGIN mname-not-master.zone01.xa.

@                       IN  SOA  (
                                   ns1.mname-not-master.zone01.xa.
                                   root.nic.test.
                                   2022122300 ; ns1 serves this serial, ns2 a slightly higher one
                                   86400
                                   14400
                                   3600000
                                   3600
                                 )

                            NS   ns1.mname-not-master.zone01.xa.
                            NS   ns2.mname-not-master.zone01.xa.

ns1                         A    127.0.1.1
ns2                         A    127.0.1.2
```

When running `zonemaster-cli` on such a broken zone, the test unfortunately crashes with a fatal error:

```
$ zonemaster-cli --test zone/zone01 --hints local.hints mname-not-master.zone01.xa
Seconds Level     Message
======= ========= =======
   0.04 CRITICAL  Fatal error in Zone: Undefined subroutine &Zonemaster::Engine::Test::Zone::max called at [redacted]/Zonemaster/Engine/Test/Zone.pm line 561.
```

tgreenx:
## Purpose

This PR proposes an update to Test Case Zone01 due to an oversight in #1035 which led to Engine crashing when several nameservers are found to serve different SOA serials.

## Context

Fixes #1175 

## Changes

- Fix missing library import
- Fix code for message tag Z01_MNAME_NOT_MASTER
- Add unit tests
- Add function `serial_gt` in Zonemaster::Engine::Util, and its documentation
- Refactoring

## How to test this PR

Unit tests should pass.

But if you want to test manually with zonemaster-cli, create a file (e.g. `local.hints`) with the following content:
```
$ more local.hints
.                              3600000      NS    IBDNS01.LABS.PRIVE.NIC.FR.
IBDNS01.LABS.PRIVE.NIC.FR.     3600000      A     10.1.72.23
.                              3600000      NS    IBDNS01-24.LABS.PRIVE.NIC.FR.
IBDNS01-24.LABS.PRIVE.NIC.FR.  3600000      A     10.1.72.24
```
Then you can do:
```
$ zonemaster-cli --test zone/zone01 --hints local.hints mname-not-master.zone01.xa --level=INFO
Seconds Level     Message
======= ========= =======
   0.00 INFO      Using version v4.6.0 of the Zonemaster engine.
   0.07 WARNING   SOA MNAME name server(s) "ns1.mname-not-master.zone01.xa/10.1.72.23" do not have the highest SOA SERIAL (expected "2022122300" but got "2022122300;2022122301")
```

Also, taken from https://github.com/zonemaster/zonemaster-engine/issues/1175#issuecomment-1380227232:
(note that I disable IPv6 due to my machine limitations at the moment)
```
$ zonemaster-cli --test zone/zone01 hwcdn.net --level=INFO --no-ipv6
Seconds Level     Message
======= ========= =======
   0.00 INFO      Using version v4.6.0 of the Zonemaster engine.
   0.56 WARNING   SOA MNAME name server(s) "ns1.hwcdn.net/69.16.174.10" do not have the highest SOA SERIAL (expected "1673538139" but got "1673538139;1673538142")
```


lib/Zonemaster/Engine/Test/Zone.pm
```diff
@@ -543,7 +544,7 @@ sub zone01 {
                 }
 
                 foreach my $serial ( uniq @serial_ns ){
-                    if ( $serial > $mname_serial and ( ($serial - $mname_serial) < 2**($serial_bits - 1) ) ){
```
marc-vanderwal, line 546: That was the only use of the `$serial_bits` variable defined on line 536. It isn’t used anymore, so it can be removed.

marc-vanderwal (commented):
Just one small detail, otherwise it looks good to me.

marc-vanderwal (approved):
Looks good to me.

lib/Zonemaster/Engine/Util.pm
```diff
@@ -236,6 +236,12 @@ sub parse_hints {
     return \%hints;
 }
 
+sub serial_gt {
+    my ( $sa, $sb ) = @_;
+
+    return ( $sa > $sb and ( ($sa - $sb) < 2**(32 - 1) ) );
+}
```
matsduf, line 243: Let us say 
$sa == 1
$sb == (2^32) -1 = 4,294,967,295

serial_gt ($sa, $sb) will return false, but $sa is a higher SOA serial than $sb.

lib/Zonemaster/Engine/Util.pm
```diff
@@ -236,6 +236,12 @@ sub parse_hints {
     return \%hints;
 }
 
+sub serial_gt {
+    my ( $sa, $sb ) = @_;
+
+    return ( $sa > $sb and ( ($sa - $sb) < 2**(32 - 1) ) );
```
matsduf, line 242: Why `2**(32 - 1)` instead of `2**31`?

matsduf (changes requested):

lib/Zonemaster/Engine/Util.pm
tgreenx: Yes, there are undefined results. From [RFC1982, section 3.2](https://www.rfc-editor.org/rfc/rfc1982#section-3.2):

>    Note that there are some pairs of values s1 and s2 for which s1 is
>    not equal to s2, but for which s1 is neither greater than, nor less
>    than, s2.  An attempt to use these ordering operators on such pairs
>    of values produces an undefined result.

matsduf: My example is not undefined. $sa is greater than $sb when compared as SOA serial values.

What is undefined is if two SOA serial values, s1 and s2 meet the following equation:
```
| s1 - s2 | = 2^31
```

lib/Zonemaster/Engine/Util.pm
matsduf: If there as a constant $SERIAL_BITS = 32 then is would make more sense. Maybe references to the RFC could be added?

lib/Zonemaster/Engine/Util.pm
marc-vanderwal: @matsduf is right: this code won’t take into account rollovers. With serial arithmetic, 4 294 967 295 + 2 equals 1, and according to the definition, 1 > 4 294 967 295.

The correct code is:

```perl
return ( $sa < $sb and ( ($sb - $sa) > 2**31 ) ) or
       ( $sa > $sb and ( ($sa - $sb) < 2**31 ) );
```

marc-vanderwal (changes requested):

lib/Zonemaster/Engine/Util.pm
tgreenx: > If there as a constant $SERIAL_BITS = 32 then is would make more sense. Maybe references to the RFC could be added?

I added such a constant and updated documentation to refer to the relevant RFC section.

marc-vanderwal (approved):
Looks good to me.

lib/Zonemaster/Engine/Constants.pm
```diff
@@ -262,7 +264,11 @@ C<$IP_VERSION_6>
 
 =item *
 
-C<$MAX_SERIAL_VARIATION>
+C<$SERIAL_BITS>
```
matsduf, line 267: I can see that none of the constants are defined or described, but I think it would make sense to have a short definition or description of the constant. Preferably here in the POD, or else in the code where it is set.

matsduf (approved):

lib/Zonemaster/Engine/Constants.pm
tgreenx: Added documentation

matsduf (approved):

marc-vanderwal (approved):

PR #1195, opened 2023-02-15
Check PO files

pnax:
## Purpose

Add unit test for PO files.

## Context

Improve our PO files.

## Changes

* new `check-po` rule
* new unit test in t/po-files.t

#### `check-po` rule

1. add a trailing space to a msgstr in a PO file
2. run `make check-po`, an error should be raised
3. remove the trailing space and verify that no error is raised anymore

#### unit tests

* Do the same but this time run `prove t/po-files.t`

* t/po-files.t
    1. alter a PO file so that it's untidied
    2. run `prove t/po-files.t`, an error should be raised
    3. remove the alteration, no more error should be raised


pnax:
Tests are failing because the PO files are not tidied. Hopefully this will be the case after v2022.2.2.

matsduf:
> Tests are failing because the PO files are not tidied. Hopefully this will be the case after v2022.2.2.

I always mean to run `./update-po xx.po` after updating a PO file before doing `git add` but it is easy to forget. There is also the `../../zonemaster-engine/util/check-msg-args xx.po` command to run before adding the PO file to git.

pnax:
> Will make check-po just check for imbalanced trailing new lines between msgid and msgstr or will it check for something else?

It will check based on `mgfmt`, see https://www.gnu.org/software/gettext/manual/html_node/msgfmt-Invocation.html#index-_002dc_002c-msgfmt-option. So it's not looking for trailing newline per se, but checking that there is the same number of format specifier.

pnax:
> > There is also the ../../zonemaster-engine/util/check-msg-args xx.po command to run before adding the PO file to git.
>
> I'll look into it.

In fact with the current implementation of `check-po` relying on `msgfmt -c`, the `check-msg-args` script becomes unnecessary.
Applying the following diff:
```diff
diff --git a/share/fr.po b/share/fr.po
index 4438d59..94713c7 100644
--- a/share/fr.po
+++ b/share/fr.po
@@ -34,7 +34,7 @@ msgstr "Le nom de domaine commence par un point."
 msgid ""
 "Domain name has an ASCII label (\"{label}\") with a character not permitted."
 msgstr ""
-"Le nom de domaine contient un label (\"{label}\") avec des caractères "
+"Le nom de domaine contient un label (\"{lbel}\") avec des caractères "
 "interdits."
 
 #, perl-brace-format
```
And running `make check-po`, I get:
```
$ make check-po
fr.po:36: a format specification for argument 'label' doesn't exist in 'msgstr'
msgfmt: found 1 fatal error
make: *** [GNUmakefile:39: check-po] Error 1
```

This gives a little bit less context than the script. However it properly detects the error. Do we need to keep the `check-msg-args` script?

If we really want to keep the script, it needs to be updated to exit with a proper error code.

matsduf:
Could `share/GNUmakefile` be update with some description of what the command does? Or possibly put that in the translation instruction document.

Will `make check-po` warn if there is an argument in msgstr that is not in msgid as check-msg-args does?

```
check-msg-args sv.po 

# sv.po line 14
msgid "Ambiguous downcaseing of character \"{unicode_name}\" in the domain name. Use all lower case instead {lisa}."
msgstr "Domännamnet innehåller den versala (stora) bokstaven \"{unicode_name}\". Det är oklart {nisse} hur den ska omvandlas till den motsvarande gemen (lilla) bokstaven. Använd bara gemener (små bokstäver) istället."
  Only in msgid: {lisa}
  Only in msgstr: {nisse}
```

Definitely better description than in your example. Yes, I think that we should keep check-msg-args, and the error code is probably not very important.

pnax:
> Will make check-po warn if there is an argument in msgstr that is not in msgid as check-msg-args does?

`msgfmt -c` seems to only check for argument present in msgid. I've added `check-msg-args` to the `check-po` recipe.

matsduf:
1. If the error is that an argument in msgid is not used in msgstr then the first part of "check-po" will report, and that report is not as good as the one that `check-msg-args` gives. 
```
$ make check-po POFILES=sv.po
gmake[1]: Entering directory '/usr/home/matsd/git/zonemaster-engine/share'
sv.po:18: a format specification for argument 'unicode_name' doesn't exist in 'msgstr'
msgfmt: found 1 fatal error
gmake[1]: *** [GNUmakefile:39: check-po] Error 1
gmake[1]: Leaving directory '/usr/home/matsd/git/zonemaster-engine/share'
*** Error code 2

Stop.
make: stopped in /usr/home/matsd/git/zonemaster-engine/share
```
```
$ ../../util/check-msg-args sv.po
-bash: ../../util/check-msg-args: No such file or directory
[matsd@freebsd13-build ~/git/zonemaster-engine/share]$ ../util/check-msg-args sv.po

# sv.po line 14
msgid "Ambiguous downcaseing of character \"{unicode_name}\" in the domain name. Use all lower case instead."
msgstr "Domännamnet innehåller den versala (stora) bokstaven. Det är oklart hur den ska omvandlas till den motsvarande gemen (lilla) bokstaven. Använd bara gemener (små bokstäver) istället."
  Only in msgid: {unicode_name}
```

2. `check-po` does not work will if `POFILES` has not been specified. Here I have created error both in fr.po and sv.po but only the error in fr.po is shown. It should be required that `POFILES` is non-empty.
```
$ make check-po
gmake[1]: Entering directory '/usr/home/matsd/git/zonemaster-engine/share'

# fr.po line 15
msgid "Ambiguous downcaseing of character \"{unicode_name}\" in the domain name. Use all lower case instead."
msgstr "Ambiguïté lors du passage en minuscule du caractère \"{unicode_name}\" dans le nom de domaine. Utiliser plutôt seulement des minuscules {nisse}."
  Only in msgstr: {nisse}
gmake[1]: *** [GNUmakefile:39: check-po] Error 1
gmake[1]: Leaving directory '/usr/home/matsd/git/zonemaster-engine/share'
*** Error code 2

Stop.
make: stopped in /usr/home/matsd/git/zonemaster-engine/share
```

matsduf:
I suggest that `check-msg-args` is moved to the share directory to make it easier to use.

matsduf:
Could `POFILES` be shortened to `PO`?

matsduf:
If the `make check-po` is to be run by a unit test, then it should preferably find all errors at once else you will fix one error just to discover that there are more errors.

1. Remove ../util/check-msg-args from `check-po`.
2. Run both `make check-po` and `make check-msg-args`.
3. Run the tests explicitly for each available PO file.

pnax:
I've rebased to make `make check-msg-args` a test on its own.

matsduf (approved):
I still think there should be some documentation in GNUmakefile. Zonemaster needs more documentation. But I say OK to this.

pnax:
> Tests are failing because the PO files are not tidied. Hopefully this will be the case after v2022.2.2.

It does not seem to be the case. So commits were added to first prepare the PO files. Then the PR has been rebased on top of develop to include all latest changes from v2022.2.2.

matsduf:
> > Tests are failing because the PO files are not tidied. Hopefully this will be the case after v2022.2.2.
> 
> It does not seem to be the case. So commits were added to first prepare the PO files. Then the PR has been rebased on top of develop to include all latest changes from v2022.2.2.

I guess I missed that.

matsduf (approved):
It is OK, but please add some documentation in GNUmakefile for the new `check-po` command.

pnax:
> It is OK, but please add some documentation in GNUmakefile for the new check-po command.

TBH I don't know how to properly document this rule without paraphrasing the code. For now this is the only thing I have in mind (and I really don't think this is worth putting into documentation):
```
Run "msgfmt -c" on all PO files
```

matsduf:
> > It is OK, but please add some documentation in GNUmakefile for the new check-po command.
> 
> TBH I don't know how to properly document this rule without paraphrasing the code. For now this is the only thing I have in mind (and I really don't think this is worth putting into documentation):
> 
> ```
> Run "msgfmt -c" on all PO files
> ```

I mean to document what `check-po` does.

PR #1200, opened 2023-02-24
Avoid unnecessary calls to Profile::get in logger

pnax:
## Purpose

Store logfilter property instead of querying it at every method's call. This will improve performance.

## Context

Taken from https://gitlab.rd.nic.fr/zonemaster/zonemaster-engine/-/commit/8b303b6c5d2e1a148c592325839256cc85f828b8

Credits to @blacksponge for the proposal

## Changes

Logger.pm

## How to test this PR

Tests should pass.

To assess the performance improvements, run:

```
perl -d:NYTProf -MZonemaster::Engine -E 'say join "\n", Zonemaster::Engine->test_zone("zonemaster.net")'
nytprofhtml --open
```
And check that the time passed in the `_check_filter` routine has decreased.

### DIFF
--- a/lib/Zonemaster/Engine/Logger.pm
+++ b/lib/Zonemaster/Engine/Logger.pm
@@ -23,6 +23,8 @@ has 'entries' => (
 );
 has 'callback' => ( is => 'rw', isa => 'CodeRef', required => 0, clearer => 'clear_callback' );
 
+my $logfilter;
+
 sub add {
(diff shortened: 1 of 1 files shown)


matsduf:
Could you please add some documentation to what method `_check_filter` does? Preferably with POD.

matsduf (approved):

Issue #1129, opened 2022-09-16
Clean up after dropping support for Net::IP


mattias-p:
After we drop support for Net::IP in favor of Net::IP::XS in #1119 there are some things we need to clean up.

We need to take care of dependencies for new installations and upgrades. Also, the whole purpose of Zonemaster::Engine::Net::IP is gone so it should be deprecated and eventually removed.

* The installation instructions should be updated install Net::IP::XS instead of Net::IP. (#1130)
* An upgrade instruction should be added directing users to install Net::IP::XS when upgrading to the new version of Zonemaster::Engine.
* The API documentation for Zonemaster::Engine::Net::IP should state clearly that the module is deprecated and that it will be removed. It should also encourage people to migrate their code to use Net::IP::XS directly instead.

matsduf:
>     * The installation instructions should be updated install Net::IP::XS instead of Net::IP.

See #1130.

pnax:
> * The API documentation for Zonemaster::Engine::Net::IP should state clearly that the module is deprecated and that it will be removed.

Done with #1202

PR #1202, opened 2023-02-27
Deprecate Zonemaster::Engine::Net::IP module

pnax:
## Purpose

Replace internal use of Zonemaster::Engine::Net::IP with Net::IP::XS module. And mark Zonemaster::Engine::Net::IP as deprecated.

## Context

#1129 

## Changes

* Internal change
* update documentation with deprecation note

## How to test this PR

* Unit tests should pass.
* No more internal use of the Zonemaster::Engine::Net::IP module

### DIFF
--- a/lib/Zonemaster/Engine/ASNLookup.pm
+++ b/lib/Zonemaster/Engine/ASNLookup.pm
@@ -7,12 +7,12 @@ use warnings;
 use version; our $VERSION = version->declare( "v1.0.10" );
 
 use Zonemaster::Engine;
-use Zonemaster::Engine::Net::IP;
 use Zonemaster::Engine::Nameserver;
 use Zonemaster::Engine::Profile;
 
 use IO::Socket;
 use IO::Socket::INET;
+use Net::IP::XS;
 
 our @db_sources;
 our $db_style;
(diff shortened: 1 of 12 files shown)


matsduf (approved):
This looks fine. I see no issues with this change.

* Do we really need to go through deprecation process, or could we consider this to be a purely internal change? I think we could consider this to be an internal change without deprecation.
* If deprecation is reasonable, then I suggest that we shorten it to v2023.2
* If no deprecation is needed, then "V-Patch" is reasonable.

A general comment, not only for this PR, we should probably create a deprecation registry to keep track of all deprecations.

blacksponge:
Unit test are passing and the only reference to the module is from the module it self.

```
$ grep -r Zonemaster::Engine::Net::IP /usr/local/share/perl/5.34.0/Zonemaster/
/usr/local/share/perl/5.34.0/Zonemaster/Engine/Net/IP.pm:package Zonemaster::Engine::Net::IP;
/usr/local/share/perl/5.34.0/Zonemaster/Engine/Net/IP.pm:Zonemaster::Engine::Net::IP - Net::IP::XS Wrapper
/usr/local/share/perl/5.34.0/Zonemaster/Engine/Net/IP.pm:    my $ip = Zonemaster::Engine::Net::IP->new( q{0.0.0.0/8} );
```

PR #1207, opened 2023-03-03
Fix RIPE ASN lookup results parsing

pnax:
## Purpose

When using RIPE ASN lookup style, the following errors are raised:
```
$ perl -I lib/ -MZonemaster::Engine -E 'say join "\n", Zonemaster::Engine->test_module("CONNECTIVITY", "paris")'
Argument "2484/2486" isn't numeric in sort at lib/Zonemaster/Engine/Test/Connectivity.pm line 482.
Argument "8561/50611/48519" isn't numeric in sort at lib/Zonemaster/Engine/Test/Connectivity.pm line 482.
Argument "8561/50611/48519" isn't numeric in sort at lib/Zonemaster/Engine/Test/Connectivity.pm line 520.
Argument "2486/2484" isn't numeric in sort at lib/Zonemaster/Engine/Test/Connectivity.pm line 520.
Argument "2484/2486" isn't numeric in sort at lib/Zonemaster/Engine/Test/Connectivity.pm line 535.
Argument "8561/50611/48519" isn't numeric in sort at lib/Zonemaster/Engine/Test/Connectivity.pm line 535.
Argument "8561/50611/48519" isn't numeric in sort at lib/Zonemaster/Engine/Test/Connectivity.pm line 537.
...
```
This fixes the parsing of the result from [RISwhois](https://www.ripe.net/analyse/archived-projects/ris-tools-web-interfaces/riswhois)

## Context

n/a

## Changes

ASNLookup.pm

## How to test this PR

Check that there is no more error:
```
perl -I lib/ -MZonemaster::Engine -E 'say join "\n", Zonemaster::Engine->test_module("CONNECTIVITY", "paris")'
```

PR #1208, opened 2023-03-06
Fixes broken handling of return codes from ASN lookup

matsduf:
## Purpose

The the test case specification of CONNECTIVITY03 it specified that certain errors in the lookup responses should be handled by the test case. That was halfway implemented. The return codes were never sent to the test case code. This PR fixes that.

## How to test this PR

Erroneous response from a ASN database must be created. Unit tests will be added to this PR.

### DIFF
--- a/lib/Zonemaster/Engine/ASNLookup.pm
+++ b/lib/Zonemaster/Engine/ASNLookup.pm
@@ -49,12 +49,13 @@ sub get_with_prefix {
         die "ASN database sources undefined";
     }
 
-    my ( $asnref, $prefix, $raw );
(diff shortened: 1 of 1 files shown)


pnax:
> The return codes were never sent to the test case code.

They were before I implemented #1207. I see I forgot to return `$ret_code`.
This looks good to me. If you think it's ready and mark it as such I'll approve it.

matsduf:
> They were before I implemented #1207. I see I forgot to return `$ret_code`. This looks good to me. If you think it's ready and mark it as such I'll approve it.

I will try to do some testing first, and then I will make it ready.

PR #1209, opened 2023-03-09
Accept YAML input for profile

pnax:
## Purpose

This PR allows to pass the profile as a YAML file.
This is the first step towards replacing the current JSON input format. JSON does not support comments, and it would be nice to provide a profile document that supports comments for values and properties.

## Context

Meeting discussion on 2023-03-09.

## Changes

* new `from_yaml()` and `to_yaml()` methods in Profile.pm
* new `share/profile.yaml`
* update MANIFEST
* upgrade script `util/json2yaml.pl`

## How to test this PR

Unit test should pass.


matsduf:
Can you add POD documentation for from_yaml() and to_yaml()?

Why is profile.json kept?

I cannot see that profile.yaml is read.

lib/Zonemaster/Engine/Profile.pm
```diff
@@ -383,6 +386,18 @@ sub to_json {
...
+sub from_yaml {
+    my ( $class, $yaml ) = @_;
+    my $data = Load( $yaml );
+    return $class->from_json( encode_json( $data ) );
+}
+
+sub to_yaml {
+    my ( $self ) = @_;
+
+    return Dump( $self->{q{profile}} );
```
marc-vanderwal, line 398: The automatic exports of `Dump()` and `Load()` by `YAML::XS` makes this code slightly confusing, because `Dump()` sounds like a function that could have been exported by `Data::Dumper` (which, by the way, does provide a `Dump` function)…

I suggest changing `use YAML::XS;` to `use YAML::XS qw();` to prevent the automatic exports, then calling the `Dump` function with its fully-qualified name, `YAML::XS::Dump()`.

pnax: Good point. I'll provide an update.

pnax:
> Can you add POD documentation for from_yaml() and to_yaml()? Some documentation for json2yaml.pl is needed. Can you add that as POD into the script?

Done.

> Why is profile.json kept?
> I cannot see that profile.yaml is read.

This PR is about accepting YAML input, not replacing the JSON input.

> profile_additional_properties.json should also be provided in yaml format, shouldn't it?

Not necessarily. Since the end goal of using YAML as the input format, is to have only one profile file with commented values, I think this can be left for a future work (when a new unified `profile.yaml` file is created and used for instance).

matsduf:
> > Can you add POD documentation for from_yaml() and to_yaml()? Some documentation for json2yaml.pl is needed. Can you add that as POD into the script?
> 
> Done.

I can see the documentation in Profile.pm. Thanks. Could you provide some documentation for json2yaml.pl too?

> > Why is profile.json kept?
> > I cannot see that profile.yaml is read.
> 
> This PR is about accepting YAML input, not replacing the JSON input.

I cannot see that the profile.yaml is used. Is it? If not, what is the purpose of it?

> > profile_additional_properties.json should also be provided in yaml format, shouldn't it?
> 
> Not necessarily. Since the end goal of using YAML as the input format, is to have only one profile file with commented values, I think this can be left for a future work (when a new unified `profile.yaml` file is created and used for instance).

That is fine, but then profile_additional_properties.json should be incorporated in profile.yaml, shouldn't they?

pnax:
> Could you provide some documentation for json2yaml.pl too?

I forgot to add it. I'll do it.

> I cannot see that the profile.yaml is used. Is it? If not, what is the purpose of it?

Not yet. It's current purpose it to document how we could use it and start from there.

> That is fine, but then profile_additional_properties.json should be incorporated in profile.yaml, shouldn't they?

This is the idea. Nevertheless I'd like to keep this PR about providing the basic methods to deal with a YAML formatted profile.

PR #1212, opened 2023-03-21
Update Basic01

tgreenx:
## Purpose

This PR proposes an update of Basic01 to the latest specification (https://github.com/zonemaster/zonemaster/pull/1082).

TBD:
- [x] Update unitary tests

## Context

Fixes #568

## Changes

 - Update to latest specification (https://github.com/zonemaster/zonemaster/pull/1082)

## How to test this PR
```
$ zonemaster-cli --show-testcase --test basic/basic01 --level INFO zonemaster.net
Seconds Level     Testcase       Message
======= ========= ============== =======
   0.00 INFO      UNSPECIFIED    Using version v4.6.2 of the Zonemaster engine.
   4.27 INFO      BASIC01        The parent zone is "net." as returned from name servers "a.gtld-servers.net/192.5.6.30;a.gtld-servers.net/2001:503:a83e::2:30;b.gtld-servers.net/192.33.14.30;b.gtld-servers.net/2001:503:231d::2:30;c.gtld-servers.net/192.26.92.30;c.gtld-servers.net/2001:503:83eb::30;d.gtld-servers.net/192.31.80.30;d.gtld-servers.net/2001:500:856e::30;e.gtld-servers.net/192.12.94.30;e.gtld-servers.net/2001:502:1ca1::30;f.gtld-servers.net/192.35.51.30;f.gtld-servers.net/2001:503:d414::30;g.gtld-servers.net/192.42.93.30;g.gtld-servers.net/2001:503:eea3::30;h.gtld-servers.net/192.54.112.30;h.gtld-servers.net/2001:502:8cc::30;i.gtld-servers.net/192.43.172.30;i.gtld-servers.net/2001:503:39c1::30;j.gtld-servers.net/192.48.79.30;j.gtld-servers.net/2001:502:7094::30;k.gtld-servers.net/192.52.178.30;k.gtld-servers.net/2001:503:d2d::30;l.gtld-servers.net/192.41.162.30;l.gtld-servers.net/2001:500:d937::30;m.gtld-servers.net/192.55.83.30;m.gtld-servers.net/2001:501:b1f9::30".
   4.27 INFO      BASIC01        The zone "zonemaster.net" is found.

$ zonemaster-cli --show-testcase --test basic/basic01 --level INFO zonemaster.fr
Seconds Level     Testcase       Message
======= ========= ============== =======
   0.00 INFO      UNSPECIFIED    Using version v4.6.2 of the Zonemaster engine.
[...]

PR #1213, opened 2023-04-05
Add DNAME support (to Zone objects)

### LINKED ISSUE #1075: The built-in recursor in Zonemaster does not handle DNAME records.
I just saw someone on Stack Overflow recommend `DNAME` as a solution to a weird DNS setup problem, and was reminded that I never implemented that in Zonemaster's recursor. Given that those records seem to be extremely rare in the real world, fixing this is probably very low priority, and this issue is just intended to inform the project that the deficiency exists.

tgreenx:
## Purpose

This PR proposes an implementation for the support of DNAME by adding a new attribute to `Zonemaster::Engine::Zone` objects, instead of modifying `Zonemaster::Engine::Recursor`.

This means that this PR does not make recursive lookups of Zonemaster follow DNAME records (similar to #1288 for CNAME), but that querying zone data (e.g glue, name servers, etc) for any name (zone) that is an alias (DNAME) will return the proper data for that zone.

TBD:
- [ ] Discuss this implementation. 
- [ ] Add and update unitary tests, and re-record unit test data

## Context

Relates to https://github.com/zonemaster/zonemaster/issues/1075 and https://github.com/zonemaster/zonemaster/issues/472

Depends on https://github.com/zonemaster/zonemaster-ldns/pull/170

## Changes

 - Add new attribute 'dname' to `Zonemaster::Engine::Zone` objects
 - Update implementation to account for that new attribute and make queries to the appropriate zone name, if any

## How to test this PR

Unit tests are to be created. So far many existing unit tests will fail because of incomplete unit test data (I haven't re-recorded them yet).

Manual testing (zone: `xn--mori-qsa.nz`):
```
$ perl -MZonemaster::Engine -E 'my $zone = Zonemaster::Engine::Zone->new({ name => Zonemaster::Engine::DNSName->from_string("xn--mori-qsa.nz") }); say "zone: ", $zone->name; $zone->dname ? say "dname: ", $zone->dname->name : say "dname: undef"; say "ns:\n\t", join "\n\t", @{$zone->ns} if $zone->dname;'
zone: xn--mori-qsa.nz
dname: maori.nz
ns:
        ns1.dns.net.nz/2001:dce:2000:2::130
        ns1.dns.net.nz/202.46.190.130
        ns2.dns.net.nz/2001:dce:7000:2::130
        ns2.dns.net.nz/202.46.187.130
        ns3.dns.net.nz/2001:dce:d453::53
        ns3.dns.net.nz/202.46.188.130
        ns4.dns.net.nz/2001:dce:d454::53
        ns4.dns.net.nz/202.46.189.130
        ns5.dns.net.nz/185.159.197.130
        ns5.dns.net.nz/2620:10a:80aa::130
        ns6.dns.net.nz/185.159.198.130
        ns6.dns.net.nz/2620:10a:80ab::130
        ns7.dns.net.nz/194.146.106.54
        ns7.dns.net.nz/2001:67c:1010:13::53
```

### DIFF
--- a/lib/Zonemaster/Engine/Zone.pm
+++ b/lib/Zonemaster/Engine/Zone.pm
@@ -3,7 +3,7 @@ package Zonemaster::Engine::Zone;
 use v5.16.0;
 use warnings;
 
-use version; our $VERSION = version->declare("v1.1.9");
+use version; our $VERSION = version->declare("v1.2.0");
 
 use Carp qw( confess croak );
 use List::MoreUtils qw[uniq];
@@ -11,7 +11,7 @@ use List::MoreUtils qw[uniq];
 use Zonemaster::Engine::DNSName;
 use Zonemaster::Engine::Recursor;
 use Zonemaster::Engine::NSArray;
-use Zonemaster::Engine::Constants qw[:ip];
+use Zonemaster::Engine::Constants qw[:ip :dname];
 
 sub new {
     my ( $class, $attrs ) = @_;
@@ -44,6 +44,16 @@ sub parent {
     }
 
     return $self->{_parent};
+}
+
+sub dname {
+    my ( $self ) = @_;
+
+    if ( !exists $self->{_dname} ) {
+        $self->{_dname} = $self->_build_dname;
+    }
+
+    return $self->{_dname};
 }
 
 sub glue_names {
@@ -96,6 +106,7 @@ sub glue_addresses {
     return $self->{_glue_addresses};
 }
 
+
 ###
 ### Builders
 ###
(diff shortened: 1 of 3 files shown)


lib/Zonemaster/Engine/Zone.pm
```diff
@@ -38,25 +39,63 @@ sub _build_parent {
...
+    foreach my $rr ( $p->get_records( 'DNAME', q{answer} ) ) {
+        if ( index( $self->name->fqdn, lc($rr->owner) ) != -1 ) {
+            $dname = Zonemaster::Engine::DNSName->new( lc($rr->dname) );
+            $owner = Zonemaster::Engine::DNSName->new( lc($rr->owner) );
+
+            splice @name_labels, scalar @name_labels - scalar @{$owner->labels}, scalar @name_labels, @{$dname->labels};
+
+            push @dnames, Zonemaster::Engine::DNSName->new( lc( join( q{.}, @name_labels ) ) );
+        }
+    }
```
matsduf, line 67: 1. It is not permitted to have more than one DNAME record with the same owner name. One approach is to do as here and let the calling function do what it was.
2. The code gets the owner name of the RR in the answer section, doesn't it? The owner name should (must) be the same as in the query.

matsduf:
I am not convinced that advanced support of DNAME is needed. It is at least not for BASIC01.

matsduf:
As conluded in https://github.com/zonemaster/zonemaster-engine/pull/1212#issuecomment-1501126691 BASIC01 does not require advanced DNAME function.

If a DNAME record is present in `example.com` (with `target.xa` as target) then that will not be seen in queries for records in `example.com` (unless DNAME is the query type). Queries for `sub.example.com` (or `next.sub.example.com`) will result in the DNAME record plus a synthesized CNAME record being included. The CNAME record for our exampel will be `sub.example.com. CNAME sub.target.xa.` (`next.sub.example.com. CNAME next.sub.target.xa.`).

Besides the delegation issue handled by BASIC01, Zonemaster can meet synthesized CNAME records for host names (right side) of MX or NS or the SOA MNAME. In those cases Zonemaster can ignore the DNAME and just follow the CNAME until we have a test case that should validate those.

Just as with normal CNAME, DNAME can create a loop. DNAME loop can be handles as a CNAME loop since we get the synthesized CNAME.

The is one special "loop" of DNAME. If we have `example.com. DNAME sub.example.com.` then that seems to be handled differently on different name server software. If Zonemaster needs any special handling of DNAME then possibly that target of  the DNAME record must not be a subdomain of owner name of the record. And such a test should then be done any time a DNAME record is included in the response. That would then be a SYSTEM message since that is not bound to a specific test case.

lib/Zonemaster/Engine/Zone.pm
```diff
@@ -38,25 +39,63 @@ sub _build_parent {
     return __PACKAGE__->new( { name => $pname } );
 }
 
+sub _build_dname {
```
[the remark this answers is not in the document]
marc-vanderwal: I don’t think so. The code follows the pattern of having methods named `something` that lazily build whatever data is needed in a `_build_something` function that is meant to be private. If anything, the method that should be documented is the newly-introduced `dname`.

lib/Zonemaster/Engine/Zone.pm
```diff
@@ -113,25 +124,63 @@ sub _build_parent {
...
+
+    return [] if not defined $p;
+
+    my @name_labels = @{$self->name->labels};
+    my $dname;
+    my $owner;
+    my @dnames;
+
+    foreach my $rr ( $p->get_records( 'DNAME', q{answer} ) ) {
+        if ( index( $self->name->fqdn, lc($rr->owner) ) != -1 ) {
```
marc-vanderwal, line 144: Careful with that: you don’t want `ab.example.com` to match `b.example.com`. `$self->name->fqdn` should be either equal to or a subdomain of `$rr->owner`. Don’t we have a function that tests whether a domain is equal to or a subdomain of another?

lib/Zonemaster/Engine/Zone.pm
matsduf: But it is possible to have a chain of DNAME records, and possibly with some CNAME records in the chain too.

lib/Zonemaster/Engine/Zone.pm
```diff
@@ -113,25 +124,108 @@ sub _build_parent {
...
+    }
+
+    unless ( scalar @original_rrs == scalar @dname_rrs ) {
+        @dname_rrs = @original_rrs;
+    }
+
+    # Break if there are too many records
+    if ( scalar @dname_rrs > $DNAME_MAX_RECORDS ) {
+        return undef;
+    }
```
marc-vanderwal, line 163: We could do this check earlier, before we start removing duplicate DNAME resource records maybe.

lib/Zonemaster/Engine/Zone.pm
```diff
@@ -113,25 +124,108 @@ sub _build_parent {
...
+    while ( $dnames{$target} ) {
+        return undef if $dname_counter > $DNAME_MAX_RECORDS; # Loop protection (for good measure only - data in %dnames is sanitized already)
+        $target = $dnames{$target};
+        $dname_counter++;
+    }
+
+    # Make sure that the DNAME chain from the RRs is not broken
+    if ( $dname_counter != scalar @dname_rrs ) {
+        return undef;
+    }
```
marc-vanderwal, line 197: Is this necessary? If the chain is broken, wouldn’t it make more sense to just follow it until we can’t?

marc-vanderwal (commented):
The new code lacks unit tests. It could be made more testable if `_build_dname()` were broken up in two pieces: one function doing the DNAME query and another processing the response packet. The second function could then be tested using synthetic response packets.

lib/Zonemaster/Engine/Zone.pm
matsduf: What does broken DNAME chain mean? Does it just mean that the domain name in RDATA does not exist? Another way a DNAME chain (or DNAME record) can be broken is that it points to itself or below itself.
```
foo.xa. DNAME foo.xa.
foo.xa. DNAME www.foo.xa.
```
And of course, we can create a loop, just like with CNAMEs.
```
foo.xa. DNAME bar.xa.
bar.xa. DNAME foo.xa.
```

matsduf: > For example, if there are three (_validated_) DNAME records in the RRset, it should loop three times.

What do you mean by that? There cannot be more than one DNAME in a DNAME RRset. A name can only point at one other name.

PR #1257, opened 2023-07-18
Refactor ASNLookup code and documentation

tgreenx:
## Purpose

This PR is four-fold: 
- It updates the ASN Lookup function of Zonemaster by using standard recursive DNS lookups instead of expecting dedicated zones.
- It fixes the backup source functionality (see https://github.com/zonemaster/zonemaster-engine/pull/1311#issuecomment-1845380408)
- It removes documentation and code for deprecated profile parameter `asnroots`
- It updates documentation and unitary tests for profile parameters `asn_db.style` and `asn_db.sources`

## Context

Fixes #1255, #1052, #1053, #835

Also follows up on https://github.com/zonemaster/zonemaster-engine/pull/1311#issuecomment-1845380408

Requires #1288

## Changes

- Update code logic in `Zonemaster::Engine::ASNLookup->_cymru_asn_lookup()`
- Add message tag `ASN_LOOKUP_SOURCE` (with `level = DEBUG` , `module = System` and `testcase = Connectivity{03,04}`)
- Fix backup source functionality in `Zonemaster::Engine::ASNLookup`
- Remove `asnroots` profile property
- Refactoring
- Update documentation
- Add and update unitary tests

## How to test this PR

Tests should pass.

**Manual testing:**
1. Add ASNLookup servers `fake.asnlookup.zonemaster.net` and `asnlookup.dufberg.se` in the default profile (or just create a new one):
```
diff --git a/share/profile.json b/share/profile.json
index 2340c9fa..5ee8dc17 100644
  --- a/share/profile.json
  +++ b/share/profile.json
@@ -2,7 +2,7 @@
     "asn_db" : {
         "style" : "Cymru",
         "sources" : {
 -            "Cymru" : [ "asnlookup.zonemaster.net", "asn.cymru.com" ],
 +            "Cymru" : [ "fake.asnlookup.zonemaster.net", "asnlookup.dufberg.se" ],
             "RIPE" : [ "riswhois.ripe.net" ]
         }
     },
```
2. Start a test for Connectivity03 and Connectivity04: they should be able to reach the working ASN database server (asnlookup.dufberg.se), after skipping the non-working one (fake.asnlookup.zonemaster.net):
```
$ zonemaster-cli --show-module --show-testcase --test connectivity03 --test connectivity04 zonemaster.net --raw --level DEBUG | egrep -i "ASN_LOOKUP_SOURCE|_DIFFERENT_"
[...]

PR #1288, opened 2023-09-07
Add CNAME followage in Recursor

tgreenx:
## Purpose

This PR makes the recursive lookup functionality of Zonemaster able to follow CNAME redirections.

Attempts to follow CNAMEs will be made when all of the following are true:
- the response has RCODE "NoError"
- the answer section of the response does not contain records of the queried type, but does contain at least one CNAME record for the query name
- the answer section of the response does not contain multiple CNAME records with the same owner name
- the final target of the CNAME record(s) chain has not been followed before
- there are no records of the queried type with owner name as the final target of the CNAME record(s)

## Context

Necessary for https://github.com/zonemaster/zonemaster-engine/pull/1257 (https://github.com/zonemaster/zonemaster-engine/pull/1257#issuecomment-1710444437)

Test Zones specification: https://github.com/zonemaster/zonemaster/pull/1220

## Changes

- Update recursive lookup code (`lib/Zonemaster/Engine/Recursor.pm`)
- Add system debug message tags `CNAME_START`, `CNAME_RECORDS_DUPLICATES`, `CNAME_CHAIN_TOO_LONG`, `CNAME_LOOP_INNER`, `CNAME_LOOP_OUTER`, `CNAME_NO_MATCH`, `CNAME_RECORDS_CHAIN_BROKEN`, `CNAME_MULTIPLE_FOR_NAME` and `CNAME_RECORDS_TOO_MANY`
- Add constants `CNAME_MAX_CHAIN_LENGTH` and `CNAME_MAX_RECORDS`
- Update test cases related to CNAME (where applicable): Syntax06 and Zone07
- Add and update unitary tests
- Add and update documentation

## How to test this PR

Unit tests are updated and should pass (based on https://github.com/zonemaster/zonemaster/pull/1220).

**Manual testing:**
You can can test any zone by providing it to `Zonemaster::Engine->recurse()`, e.g. with any of the zones from https://github.com/zonemaster/zonemaster/pull/1220. See below:
```
$ perl -MZonemaster::Engine -E 'Zonemaster::Engine->add_fake_delegation( ( "." => { ns1 => [ "127.1.0.1", "fda1:b2:c3::127:1:0:1" ], ns2 => [ "127.1.0.2", "fda1:b2:c3::127:1:0:2" ] } ) ); my $p = Zonemaster::Engine->recurse("good-cname-1.cname.recursor.engine.xa"); for my $entry ( @{ Zonemaster::Engi
ne->logger->entries } ) { say "\n", $entry if index($entry->tag, "CNAME") != -1 }; say "\n", $p->string if defined $p'

;; ->>HEADER<<- opcode: QUERY, rcode: NOERROR, id: 24524
;; flags: qr aa ; QUERY: 1, ANSWER: 2, AUTHORITY: 1, ADDITIONAL: 0
;; QUESTION SECTION:
;; good-cname-1.cname.recursor.engine.xa.       IN      A

;; ANSWER SECTION:
good-cname-1.cname.recursor.engine.xa.  3600    IN      CNAME   good-cname-1-target.cname.recursor.engine.xa.
good-cname-1-target.cname.recursor.engine.xa.   3600    IN      A       127.0.0.1

;; AUTHORITY SECTION:
cname.recursor.engine.xa.       3600    IN      NS      ns1.cname.recursor.engine.xa.

;; ADDITIONAL SECTION:

;; Query time: 0 msec
;; EDNS: version 0; flags: ; udp: 512
;; SERVER: fda1:b2:c3:0:127:30:1:31
;; WHEN: Tue Nov 28 12:30:33 2023
;; MSG SIZE  rcvd: 287
```

```
$ perl -MZonemaster::Engine -E 'Zonemaster::Engine->add_fake_delegation( ( "." => { ns1 => [ "127.1.0.1", "fda1:b2:c3::127:1:0:1" ], ns2 => [ "127.1.0.2", "fda1:b2:c3::127:1:0:2" ] } ) ); my $p = Zonemaster::Engine->recurse("mult-cname.cname.recursor.engine.xa"); for my $entry ( @{ Zonemaster::Engine
->logger->entries } ) { say "\n", $entry if index($entry->tag, "CNAME") != -1 }; say "\n", $p->string if defined $p'

SYSTEM:UNSPECIFIED:CNAME_MULTIPLE_FOR_NAME name=mult-cname.cname.recursor.engine.xa
```

```
$ perl -MZonemaster::Engine -E 'Zonemaster::Engine->add_fake_delegation( ( "." => { ns1 => [ "127.1.0.1", "fda1:b2:c3::127:1:0:1" ], ns2 => [ "127.1.0.2", "fda1:b2:c3::127:1:0:2" ] } ) ); my $p = Zonemaster::Engine->recurse("looped-cname-out-of-zone.sub2.cname.recursor.engine.xa"); for my $entry ( @{ Zonemaster::Engine->logger->entries } ) { say "\n", $entry if index($entry->tag, "CNAME") != -1 }; say "\n", $p->string if defined $p'

SYSTEM:UNSPECIFIED:CNAME_LOOP_OUTER cnames=looped-cname-out-of-zone.sub2.cname.recursor.engine.xa;looped-cname-out-of-zone.sub3.cname.recursor.engine.xa; name=looped-cname-out-of-zone.sub2.cname.recursor.engine.xa; target="looped-cname-out-of-zone.sub3.cname.recursor.engine.xa"
```

### DIFF
--- a/lib/Zonemaster/Engine/Recursor.pm
+++ b/lib/Zonemaster/Engine/Recursor.pm
@@ -164,7 +164,7 @@ sub _resolve_cname {
     # Remove duplicate CNAME RRs
     my ( %duplicate_cname_rrs, @original_rrs );
     for my $rr ( @cname_rrs ) {
-        my $rr_hash = $rr->class . '/' . $rr->ttl . '/CNAME/' . $rr->owner . '/' . $rr->cname;
+        my $rr_hash = $rr->class . '/CNAME/' . lc($rr->owner) . '/' . lc($rr->cname);
 
         if ( exists $duplicate_cname_rrs{$rr_hash} ) {
             $duplicate_cname_rrs{$rr_hash}++;
@@ -201,7 +201,7 @@ sub _resolve_cname {
         }
 
         # CNAME owner name is target, or target has already been seen in this response, or owner name cannot be a target
-        if ( lc( $rr_owner ) eq lc( $rr_target ) or exists $seen_targets{lc( $rr_target )} or grep( /^$rr_target$/, keys %forbidden_targets ) ) {
+        if ( lc( $rr_owner ) eq lc( $rr_target ) or exists $seen_targets{lc( $rr_target )} or grep { $_ eq lc( $rr_target ) } ( keys %forbidden_targets ) ) {
             Zonemaster::Engine->logger->add( CNAME_LOOP_INNER => { name => join( ';', map { $_->owner } @cname_rrs ), target => join( ';', map { $_->cname } @cname_rrs ) } );
             return ( undef, $state );
         }


share/profile.json
```diff
@@ -407,6 +407,9 @@
             "CACHE_FETCHED" : "DEBUG2",
             "CACHED_RETURN" : "DEBUG3",
             "CANNOT_CONTINUE" : "CRITICAL",
+            "CNAME_LOOP_INNER" : "DEBUG",
+            "CNAME_LOOP_OUTER" : "DEBUG",
+            "CNAME_MULTIPLE_FOR_NAME" : "DEBUG",
```
matsduf, line 412: Can these three be defined or described in Recursor.pm?

matsduf:
What is the purpose of mixing case in
```
illegal1                IN  CNAME  target0
ILLEGAL1                IN  CNAME  target1
```
and others?

matsduf:
> Generally it is to test case-insensitivity of the implementation. But here it is also about testing that several CNAME records with the same owner name (with different cases) lead to an error, i.e. that the implementation outputs `CNAME_MULTIPLE_FOR_NAME` (this [part of the code](https://github.com/zonemaster/zonemaster-engine/pull/1288/files#diff-f67b3e26b5365bb9e63dc0831b878f5e86f0d66f5b1dbc50d220c586b1938bc3R198-R202)).

Then, shouldn't there be a test where the two CNAME records have identical owner names in a case sensitive comparison. too?

share/profile.json
matsduf: I do not think it is necessary to have a message ID on those, which would put a burden on the translators. Definition or description in POD would be fine.

matsduf:
@tgreenx, How should the unit tests react on NODATA and NXDOMAIN responses, respectively?

NODATA:

```
; <<>> DiG 9.18.18-0ubuntu0.22.04.1-Ubuntu <<>> @127.31.1.31 alias-nodata.recursor.core.xa A
; (1 server found)
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 38985
;; flags: qr aa rd; QUERY: 1, ANSWER: 1, AUTHORITY: 1, ADDITIONAL: 1
;; WARNING: recursion requested but not available

;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags:; udp: 1232
; COOKIE: a15d870f95dcc703 (echoed)
;; QUESTION SECTION:
;alias-nodata.recursor.core.xa.	IN	A

;; ANSWER SECTION:
alias-nodata.recursor.core.xa. 3600 IN	CNAME	target-nodata.recursor.core.xa.

;; AUTHORITY SECTION:
recursor.core.xa.	3600	IN	NS	ns1.recursor.core.xa.

;; Query time: 0 msec
;; SERVER: 127.31.1.31#53(127.31.1.31) (UDP)
;; WHEN: Sun Nov 05 23:02:53 UTC 2023
;; MSG SIZE  rcvd: 193
```

NXDOMAIN:
```
; <<>> DiG 9.18.18-0ubuntu0.22.04.1-Ubuntu <<>> @127.31.1.31 alias-nxdomain.recursor.core.xa A
; (1 server found)
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NXDOMAIN, id: 11309
;; flags: qr aa rd; QUERY: 1, ANSWER: 1, AUTHORITY: 1, ADDITIONAL: 1
;; WARNING: recursion requested but not available

;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags:; udp: 1232
; COOKIE: c82ff6fab047c6dc (echoed)
;; QUESTION SECTION:
;alias-nxdomain.recursor.core.xa. IN	A

;; ANSWER SECTION:
alias-nxdomain.recursor.core.xa. 3600 IN CNAME	target-nxdomain.recursor.core.xa.

;; AUTHORITY SECTION:
recursor.core.xa.	3600	IN	NS	ns1.recursor.core.xa.

;; Query time: 0 msec
;; SERVER: 127.31.1.31#53(127.31.1.31) (UDP)
;; WHEN: Sun Nov 05 23:02:45 UTC 2023
;; MSG SIZE  rcvd: 199
```

And how will the code handle broken chain (and here with a duplicate CNAME record)?

broken-chain.recursor.core.xa. --> broken-chain2.recursor.core.xa. --> broken-chain3.recursor.core.xa. --> NULL

NULL --> broken-chain-target.recursor.core.xa. 100 IN A	127.0.0.1

```
; <<>> DiG 9.18.18-0ubuntu0.22.04.1-Ubuntu <<>> @127.31.1.31 broken-chain.recursor.core.xa A
; (1 server found)
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 54152
;; flags: qr aa rd; QUERY: 1, ANSWER: 4, AUTHORITY: 0, ADDITIONAL: 1
;; WARNING: recursion requested but not available

;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags:; udp: 1232
; COOKIE: 83ca9efec06ee54d (echoed)
;; QUESTION SECTION:
;broken-chain.recursor.core.xa.	IN	A

;; ANSWER SECTION:
broken-chain.recursor.core.xa. 100 IN	CNAME	broken-chain2.recursor.core.xa.
broken-chain2.recursor.core.xa.	3600 IN	CNAME	broken-chain3.recursor.core.xa.
broken-chain2.recursor.core.xa.	100 IN	CNAME	broken-chain3.recursor.core.xa.
broken-chain-target.recursor.core.xa. 100 IN A	127.0.0.1

;; Query time: 0 msec
;; SERVER: 127.31.1.31#53(127.31.1.31) (UDP)
;; WHEN: Sun Nov 05 23:02:21 UTC 2023
;; MSG SIZE  rcvd: 343
```

t/recursor-A.t
```diff
@@ -0,0 +1,63 @@
...
+# Good CNAMEs
+my $p = Zonemaster::Engine->recurse( 'alias.good-cname.recursor.xa' );
+isa_ok( $p, 'Zonemaster::Engine::Packet' );
+is( scalar( $p->answer ), 1, 'one answer record' );
+is( name( ($p->answer)[0]->owner ), 'target.good-cname.recursor.xa', 'RR name ok' );
+
+$p = Zonemaster::Engine->recurse( 'chain.good-cname.recursor.xa' );
+isa_ok( $p, 'Zonemaster::Engine::Packet' );
+is( scalar( $p->answer ), 1, 'one answer record' );
+is( name( ($p->answer)[0]->owner ), 'target.good-cname.recursor.xa', 'RR name ok' );
```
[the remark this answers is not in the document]
matsduf: So if there are three A records in `target.good-cname.recursor.xa` then that would match `is( scalar( $p->answer ), 3, 'three answer records' );`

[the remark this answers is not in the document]
matsduf: I get two records in the response and the unit test fails:

```
$ perl -MZonemaster::Engine -E 'Zonemaster::Engine::Recursor->remove_fake_addresses("."); Zonemaster::Engine->add_fake_delegation( ( "." => { ns1 => [ "127.1.0.1" ] } ) ); my $p = Zonemaster::Engine->recurse("good-cname-1.cname.recursor.engine.xa", "A"); for my $entry ( @{ Zonemaster::Engine->logger->entries } ) { say "\n", $entry if index($entry->tag, "CNAME") != -1 }; say $p->string;'
;; ->>HEADER<<- opcode: QUERY, rcode: NOERROR, id: 45265
;; flags: qr aa ; QUERY: 1, ANSWER: 2, AUTHORITY: 1, ADDITIONAL: 0 
;; QUESTION SECTION:
;; good-cname-1.cname.recursor.engine.xa.	IN	A

;; ANSWER SECTION:
good-cname-1.cname.recursor.engine.xa.	3600	IN	CNAME	good-cname-1-target.cname.recursor.engine.xa.
good-cname-1-target.cname.recursor.engine.xa.	3600	IN	A	127.0.0.1

;; AUTHORITY SECTION:
cname.recursor.engine.xa.	3600	IN	NS	ns1.cname.recursor.engine.xa.

;; ADDITIONAL SECTION:

;; Query time: 0 msec
;; EDNS: version 0; flags: ; udp: 512
;; SERVER: fda1:b2:c3:0:127:30:1:31
;; WHEN: Wed Nov 15 09:40:41 2023
;; MSG SIZE  rcvd: 287
```

matsduf:
@tgreenx, see https://github.com/zonemaster/zonemaster/pull/1220

matsduf:
There are still two unit tests that fail (my update in the `t` file to print out the scenario name):
```
ok 56 - undefined as expected (LOOPED-CNAME-IN-ZONE-1)
not ok 57 - should emit CNAME_LOOP_INNER (LOOPED-CNAME-IN-ZONE-1)

#   Failed test 'should emit CNAME_LOOP_INNER (LOOPED-CNAME-IN-ZONE-1)'
#   at t/recursor-A.t line 121.
ok 58 - undefined as expected (LOOPED-CNAME-IN-ZONE-2)
not ok 59 - should emit CNAME_LOOP_INNER (LOOPED-CNAME-IN-ZONE-2)

#   Failed test 'should emit CNAME_LOOP_INNER (LOOPED-CNAME-IN-ZONE-2)'
#   at t/recursor-A.t line 127.
```

What is wrong with `LOOPED-CNAME-IN-ZONE-2` with the following response?
```
; <<>> DiG 9.18.18-0ubuntu0.22.04.1-Ubuntu <<>> @127.30.1.31 looped-cname-in-zone-2.cname.recursor.engine.xa
; (1 server found)
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 25162
;; flags: qr aa rd; QUERY: 1, ANSWER: 3, AUTHORITY: 1, ADDITIONAL: 1
;; WARNING: recursion requested but not available

;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags:; udp: 1232
; COOKIE: f6c515b68ac2fe36 (echoed)
;; QUESTION SECTION:
;looped-cname-in-zone-2.cname.recursor.engine.xa. IN A

;; ANSWER SECTION:
looped-cname-in-zone-2.cname.recursor.engine.xa. 3600 IN CNAME looped-cname-in-zone-2-a.cname.recursor.engine.xa.
looped-cname-in-zone-2-a.cname.recursor.engine.xa. 3600	IN CNAME looped-cname-in-zone-2-b.cname.recursor.engine.xa.
looped-cname-in-zone-2-b.cname.recursor.engine.xa. 3600	IN CNAME looped-cname-in-zone-2-a.cname.recursor.engine.xa.

;; AUTHORITY SECTION:
cname.recursor.engine.xa. 3600	IN	NS	ns1.cname.recursor.engine.xa.

;; Query time: 0 msec
;; SERVER: 127.30.1.31#53(127.30.1.31) (UDP)
;; WHEN: Wed Nov 29 13:41:55 UTC 2023
;; MSG SIZE  rcvd: 488
```

lib/Zonemaster/Engine/Recursor.pm
```diff
@@ -193,8 +193,65 @@ sub _recurse {
...
+
+                    $state->{cseen}{lc( $target )} = 1;
+
+                    if ( lc( $target ) eq lc( $name ) ) { # CNAME target is QNAME (inner loop)
+                        Zonemaster::Engine->logger->add( CNAME_LOOP_INNER => { name => $name, target => $target } );
+                        return ( undef, $state );
+                    }
+
+                    $state->{count} += 1;
+                    return ( undef, $state ) if $state->{count} > 100; # Loop protection
```
mattias-p, line 239: 1. Shouldn't this check be postponed until just before we actually recurse?
2. The max count is 100 here but 20 in another place. Shouldn't they be the same? Maybe define a constant for this number and refer to it in both places?

t/recursor-A.t
```diff
@@ -0,0 +1,139 @@
...
+# Common hint file (test-zone-data/COMMON/hintfile)
+Zonemaster::Engine::Recursor->remove_fake_addresses( '.' );
+Zonemaster::Engine::Recursor->add_fake_addresses( '.',
+    { 'ns1' => [ '127.1.0.1', 'fda1:b2:c3::127:1:0:1' ],
+      'ns2' => [ '127.1.0.2', 'fda1:b2:c3::127:1:0:2' ],
+    }
+);
+
+# Test zone scenarios
+# GOOD-CNAME-1
```
mattias-p, line 31: I suggest using [subtest](https://perldoc.perl.org/Test::More#subtest) here instead of simple comments. That way we'll get better diagnostics when running `prove -Ilib -v t/recursor-A.t`.

lib/Zonemaster/Engine/Recursor.pm
```diff
@@ -193,8 +193,65 @@ sub _recurse {
             return ( $p, $state );
         }
 
-        if ( $class->_is_answer( $p ) ) {    # Return answer
-            return ( $p, $state );
+        if ( $class->_is_answer( $p ) ) {    # Return answer, or follow CNAME
+            if ( not $p->has_rrs_of_type_for_name( $type, $name ) and $p->has_rrs_of_type_for_name( 'CNAME', $name ) ) {
```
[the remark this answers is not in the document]
mattias-p: Sure, I agree with all that. However my comment was not about how servers are supposed to behave but rather how Zonemaster should behave when a server misbehaves. That said I had another look at this and realized that I read the code too quickly the first time. It does in fact abort, which is reasonable.

matsduf:
How will the code react if the answer section has a CNAME pointing at a target name, but the `A` record has another owner name?

```
;; ANSWER SECTION:
cname-one.cname.recursor.engine.xa.   3600 IN CNAME cname-two.cname.recursor.engine.xa.
cname-three.cname.recursor.engine.xa. 3600 IN A 127.0.0.1
```

How will the code react if the answer section has a broken CNAME chain, i.e. a CNAME is missing in the chain?

```
;; ANSWER SECTION:
cname-one.cname.recursor.engine.xa.   3600 IN CNAME cname-two.cname.recursor.engine.xa.
cname-three.cname.recursor.engine.xa. 3600 IN CNAME cname-four.cname.recursor.engine.xa.
cname-four.cname.recursor.engine.xa.   3600 IN A 127.0.0.1
```

I could easily try to create these and other scenarios.

lib/Zonemaster/Engine/Recursor.pm
```diff
@@ -153,6 +153,54 @@ sub parent {
...
+
+    if ( $state->{tseen}{lc( $target )}  ) { # CNAME target has already been followed (outer loop); no need to recurse.
+        Zonemaster::Engine->logger->add( CNAME_LOOP_OUTER => { name => $name, target => $target, targets_seen => join( ';', keys %{ $state->{tseen} } ) } );
+        return ( undef, $state );
+    }
+
+    $state->{tseen}{lc( $target )} = 1;
+    $state->{tcount} += 1;
+
+    return ( undef, $state ) if $state->{tcount} > 100; # Safe-guard against anormaly long CNAME chains; no need to recurse
```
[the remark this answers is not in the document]
matsduf: Instead of just breaking, why not outputting e.g. CNAME_TOO_LONG_CHAIN (ERROR) at 10 (changed my mind from 12 to 10).

Maybe even a outputting e.g. CNAME_LONG_CHAIN (WARNING) at 5, but not breaking.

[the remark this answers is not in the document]
pnax: Suggestion: make this chain length configurable with an entry in the profile (e.g. `resolver.defaults.cname_chain_length`). This could be done at a later stage if this adds too much work to this PR.

lib/Zonemaster/Engine/Recursor.pm
```diff
@@ -153,6 +153,54 @@ sub parent {
...
+        return ( $p, $state );
+    }
+
+    if ( $state->{tseen}{lc( $target )}  ) { # CNAME target has already been followed (outer loop); no need to recurse.
+        Zonemaster::Engine->logger->add( CNAME_LOOP_OUTER => { name => $name, target => $target, targets_seen => join( ';', keys %{ $state->{tseen} } ) } );
+        return ( undef, $state );
+    }
+
+    $state->{tseen}{lc( $target )} = 1;
+    $state->{tcount} += 1;
```
mattias-p, line 193: I just realized we don't have any documentation of the fields of the state hash. I think we should, though. I tried to think about the right place for such documentation and I believe it's the pod documentation for _recurse(). Would you consider adding descriptions there for the new state fields added in this PR?

t/recursor-A.t
```diff
@@ -0,0 +1,150 @@
...
+    # is( $p, undef, "undefined as expected");
+    # %res = map { $_->tag => $_ } @{ Zonemaster::Engine->logger->entries };
+    # ok( $res{CNAME_LOOP_INNER}, q{should emit CNAME_LOOP_INNER} );
+# };
+
+subtest 'LOOPED-CNAME-IN-ZONE-2' => sub {
+    my $p = Zonemaster::Engine->recurse( 'looped-cname-in-zone-2.cname.recursor.engine.xa' );
+    is( $p, undef, "undefined as expected");
+    my %res = map { $_->tag => $_ } @{ Zonemaster::Engine->logger->entries };
+    ok( $res{CNAME_LOOP_INNER}, q{should emit CNAME_LOOP_INNER} );
```
matsduf, line 136: Would it be helpful if the scenario name is included in the message? E.g.
```
    is( $p, undef, "undefined as expected (LOOPED-CNAME-IN-ZONE-2)");
```
```
    ok( $res{CNAME_LOOP_INNER}, q{should emit CNAME_LOOP_INNER (LOOPED-CNAME-IN-ZONE-2)} );
```

lib/Zonemaster/Engine/Recursor.pm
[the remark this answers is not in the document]
matsduf: > It is not about the length of a CNAME chain (i.e. number of CNAME resource records) in a single response. That could be another safe-guard counter though. Are we on the same page?

Yes, sure. Doing another lookup should only be done if the target is outside the zone or zones in which the CNAME record or records reside.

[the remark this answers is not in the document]
matsduf: > Suggestion: make this chain length configurable with an entry in the profile (e.g. `resolver.defaults.cname_chain_length`). This could be done at a later stage if this adds too much work to this PR.

That could always be nice, but it makes it more complicated to create unit tests to catch the function.

matsduf:
@mattias-p wrote in https://github.com/zonemaster/zonemaster-engine/pull/1288#discussion_r1409339655:

> I believe we need to protect ourselves from an infinite loop here. E.g. when getting a response like this:
> 
> ```
> ;; QUESTION SECTION:
> ;a.example. IN A
> ;; ANSWER SECTION:
> a.example. IN CNAME b.example.
> b.example. IN CNAME a.example.
> ```

Due to a bug (I think) in CoreDNS this cannot be produced by CoreDNS unless the following is acceptable, but I do not think the code likes it.
```
;; QUESTION SECTION:
;a.example. IN A
;; ANSWER SECTION:
a.example. IN CNAME b.example.
b.example. IN CNAME a.example.
a.example. IN CNAME b.example.
b.example. IN CNAME a.example.
a.example. IN CNAME b.example.
b.example. IN CNAME a.example.
a.example. IN CNAME b.example.
b.example. IN CNAME a.example.
a.example. IN CNAME b.example.
b.example. IN CNAME a.example.
```

matsduf:
> > I could easily try to create these and other scenarios.
> 
> Yes please do.

I have added three more scenarios. I guess you will updated the unit tests.

matsduf:
Three new scenarios, LOOPED-CNAME-IN-ZONE-3, WRONG-CNAME-OWNER-NAME and EXTRA-CNAME-IN-ANSWER, have been added to the test zone (https://github.com/zonemaster/zonemaster/pull/1220).

lib/Zonemaster/Engine/Recursor.pm
```diff
@@ -153,6 +153,54 @@ sub parent {
...
+        }
+
+        if ( lc( $rr_owner ) eq lc( $rr_target ) or exists $seen_targets{lc( $rr_target )} or exists $forbidden_targets{lc( $rr_owner)} ) { # CNAME owner name is target, or target has already been seen in this response, or owner name cannot be a target
+            Zonemaster::Engine->logger->add( CNAME_LOOP_INNER => { name => join( ';', map { $_->owner } @cname_rrs ), target => join( ';', map { $_->cname } @cname_rrs ) } );
+            return ( undef, $state );
+        }
+
+        $seen_targets{lc( $rr_target )} = 1;
+        $forbidden_targets{lc( $rr_owner )} = 1;
+        $cnames{$rr_owner} = $rr_target;
```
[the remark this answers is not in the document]
matsduf: Maybe normalize in terms of upper and lower case?

t/recursor-A.t
```diff
@@ -0,0 +1,265 @@
...
+    my $p = Zonemaster::Engine->recurse( 'looped-cname-out-of-zone.sub2.cname.recursor.engine.xa' );
+    is( $p, undef, "undefined as expected");
+
+    my %res = map { $_->tag => $_ } @{ Zonemaster::Engine->logger->entries };
+    ok( $res{CNAME_START}, q{should emit CNAME_START} );
+    ok( $res{CNAME_LOOP_OUTER}, q{should emit CNAME_LOOP_OUTER} );
+};
+
+# Fails for now because there are only 10 records and the current check for CNAME_RECORDS_TOO_MANY is inclusive (> 10);
+# subtest 'TOO-LONG-CNAME-CHAIN' => sub {
```
matsduf, line 201: I suggest that the limit is set to >9 instead. 10 is too many. In that way we have test data.

t/recursor-A.t
```diff
@@ -0,0 +1,265 @@
...
+    my $p = Zonemaster::Engine->recurse( 'mult-cname.cname.recursor.engine.xa' );
+    is( $p, undef, "undefined as expected");
+
+    my %res = map { $_->tag => $_ } @{ Zonemaster::Engine->logger->entries };
+    ok( $res{CNAME_START}, q{should emit CNAME_START} );
+    ok( $res{CNAME_RECORDS_MULTIPLE_FOR_NAME}, q{should emit CNAME_RECORDS_MULTIPLE_FOR_NAME} );
+};
+
+# Untestable for now --> https://github.com/zonemaster/zonemaster/pull/1220#discussion_r1409260618
+# subtest 'LOOPED-CNAME-IN-ZONE-1' => sub {
```
matsduf, line 159: Would it be wrong to let the code reduce repeated records into one?

lib/Zonemaster/Engine/Recursor.pm
```diff
@@ -153,6 +154,97 @@ sub parent {
...
+
+    # Safe-guard against anormaly long consecutive CNAME chains; no need to recurse
+    if ( $state->{tcount} > $CNAME_MAX_CHAIN_LENGTH ) {
+        Zonemaster::Engine->logger->add( CNAME_CHAIN_TOO_LONG => { count => $state->{tcount}, max => $CNAME_MAX_CHAIN_LENGTH } );
+        return ( undef, $state );
+    }
+
+    # Make sure that the CNAME target is out of zone before making a new recursive lookup for CNAME target
+    unless ( $name->is_in_bailiwick( $target ) ) {
+        Zonemaster::Engine->logger->add( CNAME_FOLLOWED_OOB => { name => $name, target => $target } );
```
matsduf, line 237: How should OOB be interpreted? Are the new tags defined somewhere?

matsduf: I feel it is wrong to use the terms "in-bailiwick" and "out-of-bailiwick" here. In a DNS context they refer to namn server names in delegation. See the term "Bailiwick" in https://datatracker.ietf.org/doc/html/rfc8499, section 7 on page 25.

Also see my comment on Recursor.pm.

lib/Zonemaster/Engine/Recursor.pm
```diff
@@ -470,4 +587,33 @@ This list can be replaced like so:
...
+=head2 _resolve_cname()
+
+    my ( $p, $state_hash ) = _resolve_cname( $name, $type_string, $dns_class_string, $p, $state_hash );
+
+Performs CNAME resolution for the given arguments. Used by the L<recursive lookup|/_recurse()> helper method in this module.
+If CNAMEs are successfully resolved, a L<packet|Zonemaster::Engine::Packet> (which could be C<undef>) is returned along with
+one of the following message tags: CNAME_FOLLOWED_IB, CNAME_FOLLOWED_OOB.
+Note that CNAME records are also validated and, in case of an error, an empty (C<undef>) L<packet|Zonemaster::Engine::Packet>
+is returned and one of the following message tags will be logged: CNAME_CHAIN_TOO_LONG, CNAME_LOOP_INNER, CNAME_LOOP_OUTER,
+CNAME_NO_MATCH, CNAME_RECORDS_CHAIN_BROKEN, CNAME_RECORDS_MULTIPLE_FOR_NAME, CNAME_RECORDS_TOO_MANY.
```
matsduf, line 613: I suggest CNAME_FOLLOWED_IZ ("in zone") and CNAME_FOLLOWED_OOZ ("out-of-zone"), respectively, instead of  CNAME_FOLLOWED_IB and CNAME_FOLLOWED_OOB. That would capture the meaning, wouldn't it?

Could you make a list of the tags with one tag per "=item" and include a short description of what the tag stands for?

matsduf (changes requested):

matsduf:
@tgreenx, could we complete this together with [Test zones for the CNAME function in Recursor.pm in Engine]?

[Test zones for the CNAME function in Recursor.pm in Engine]: https://github.com/zonemaster/zonemaster/pull/1220

lib/Zonemaster/Engine/Recursor.pm
```diff
@@ -153,6 +154,119 @@ sub parent {
...
+    my ( $class, $name, $type, $dns_class, $p, $state ) = @_;
+    $name = name( $name );
+    Zonemaster::Engine->logger->add( CNAME_START => { name => $name, type => $type, dns_class => $dns_class } );
+
+    my @cname_rrs = $p->get_records( 'CNAME', 'answer' );
+
+    # Remove duplicate CNAME RRs
+    my ( %duplicate_cname_rrs, @original_rrs );
+    for my $rr ( @cname_rrs ) {
+        my $rr_hash = $rr->class . '/' . $rr->ttl . '/CNAME/' . $rr->owner . '/' . $rr->cname;
```
marc-vanderwal, line 167: Suppose a response contains two identical CNAME records that only differ by TTL. Should we consider them as duplicates? I’d say yes.

And how about two CNAME records that only differ by case in the owner name, CNAME name, or both? I think they should be considered duplicate too.

lib/Zonemaster/Engine/Recursor.pm
```diff
@@ -153,6 +154,119 @@ sub parent {
...
+        my $rr_target = name( $rr->cname );
+
+        # Multiple CNAME records with same owner name
+        if ( exists $forbidden_targets{lc( $rr_owner )} ) {
+            Zonemaster::Engine->logger->add( CNAME_RECORDS_MULTIPLE_FOR_NAME => { name => $rr_owner } );
+            return ( undef, $state );
+        }
+
+        # CNAME owner name is target, or target has already been seen in this response, or owner name cannot be a target
+        if ( lc( $rr_owner ) eq lc( $rr_target ) or exists $seen_targets{lc( $rr_target )} or grep( /^$rr_target$/, keys %forbidden_targets ) ) {
```
marc-vanderwal, line 204: The `grep` can have odd side-effects if `$rr_target` contains names with special characters in them. It was also missing a conversion to lowercase. It’s better to do:
```perl
grep { lc( $_ ) eq $rr_target } ( keys %forbidden_targets )
```

lib/Zonemaster/Engine/Recursor.pm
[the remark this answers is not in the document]
marc-vanderwal: We should do case-insensitive comparisons when checking whether we’ve already queried the name before (and thus, detect loops). But we should also be case-preserving: if `foo.example` is an alias to `BAR.example`, we should query `BAR.example`, not `bar.example`, even though both of these QNAMEs should in theory get us the same RRset.

lib/Zonemaster/Engine/Recursor.pm
```diff
@@ -193,8 +193,34 @@ sub _recurse {
...
+
+                    $cname = $cnames{$cname} while $cnames{$cname};
+
+                    if ( $state->{seen}{$cname} ) {
+                        Zonemaster::Engine->logger->add( CNAME_LOOP => {} );
+                        return ( undef, $state );
+                    }
+                    else {
+                        ( $p, $state ) = $class->_recurse( $cname, $type, $dns_class,
+                            { ns => [ root_servers() ], count => 0, common => 0, seen => { $cname => 1 }, glue => {} });
```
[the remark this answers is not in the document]
marc-vanderwal: Given the date of that comment: yes, apparently I forgot I already started a review eons ago.

lib/Zonemaster/Engine/Recursor.pm
matsduf: I agree.

matsduf:
Is case preservation really needed here? I cannot see that is wrong to query for `bar.xa` even it RDATA of the CNAME record was `BAR.XA`.

marc-vanderwal:
> Is case preservation really needed here? I cannot see that is wrong to query for `bar.xa` even it RDATA of the CNAME record was `BAR.XA`.

RFC 1034 does, in fact, state that:

> When you receive a domain name or label, you should preserve its case.

My understanding is that when you “receive” a CNAME, if the resolver decides to follow it, the follow-up query’s QNAME should retain the case exactly as it was received in the previous response.

So in a nutshell: it’s better to preserve case here. If not, I feel like we are adding an assumption to `Recursor` that the name server will give identical RRsets when queried for `BAR.XA` and `bar.xa`. For the initial use case of this code, it’s probably overkill. But if we decide to use the same Recursor to query untrusted nameservers, than anything goes.

marc-vanderwal (approved):

matsduf (approved):

matsduf:
I think this can be merged now.

marc-vanderwal:
Release testing for 2024.2: no issues found.

Issue #1284, opened 2023-08-30
Disabling a Test Case in the profile does not prevent it from being run if called explicitly


tgreenx:
The profile can be used to prevent Test Cases from being run (i.e. by removing them from the profile). This is true when Engine runs the whole test suite, or even specific Test modules, but not when running specific Test Cases (i.e., using [Zonemaster::Engine/test_method()](https://github.com/zonemaster/zonemaster-engine/blob/7af3518b067890247aad29c4de90c1d45ee9bdf2/lib/Zonemaster/Engine.pm#L73-L77) or [Zonemaster::Engine::Test/run_one()](https://github.com/zonemaster/zonemaster-engine/blob/7af3518b067890247aad29c4de90c1d45ee9bdf2/lib/Zonemaster/Engine/Test.pm#L155-L207)). That is because the checks are done directly in the `all()` method of each Test module (with the help of [Zonemaster::Engine::Util/should_run_test()](https://github.com/zonemaster/zonemaster-engine/blob/7af3518b067890247aad29c4de90c1d45ee9bdf2/lib/Zonemaster/Engine/Util.pm#L48-L53)). That method is not used when running specific Test Cases.

For example, using Zonemaster-CLI, after removing Address01 from the profile (note that Engine is started twice because of two `--test` arguments):
```
$ zonemaster-cli --show-testcase --test Address --test Address/address01 --level DEBUG --no-ipv6 --raw zonemaster.net | grep TEST_CASE_START
   0.00 DEBUG    ADDRESS02      TEST_CASE_START   testcase=address02
   3.33 DEBUG    ADDRESS03      TEST_CASE_START   testcase=address03
   0.00 DEBUG    ADDRESS01      TEST_CASE_START   testcase=address01
```

It could be intentional, but that seems odd to me. So, bug or feature? I would advocate for the former.

matsduf:
Yes, Engine is started twice, but I noticed when preparing data for #1281 that cache and blacklisting survive from the first run to the second run. But I think it would be better if Engine could be run once.

I think it is fine that you can include a test case not in the profile if you explicitly calls it. What is the problem of that? What do you think should happen? And why?

matsduf:
Resolved by #1294

PR #1294, opened 2023-09-20
Refactoring of function Zonemaster::Engine::Test::run_one()

### LINKED ISSUE #1278: Incoherent argument type in function 'Zonemaster::Engine::Test/run_one()'
When running the Test suite, [Zonemaster::Engine::Test](https://github.com/zonemaster/zonemaster-engine/blob/v4.7.2/lib/Zonemaster/Engine/Test.pm) will run either:
- the full suite (using [run_all_for()](https://github.com/zonemaster/zonemaster-engine/blob/v4.7.2/lib/Zonemaster/Engine/Test.pm#L63))
- a specific Test module (using [run_module()](https://github.com/zonemaster/zonemaster-engine/blob/v4.7.2/lib/Zonemaster/Engine/Test.pm#L113))
-  a specific Test Case (using [run_one()](https://github.com/zonemaster/zonemaster-engine/blob/v4.7.2/lib/Zonemaster/Engine/Test.pm#L155)). 
 
The first two take at least one argument which is a `Zonemaster::Engine::Zone` object. However, the latter takes an array of `Zonemaster::Engine::Zone` objects (under the variable `@arguments`). I don't see why that is. From what I have seen all Test Cases only ever handle one argument (which is a `Zonemaster::Engine::Zone` object).

It has no sizable effect in practice but this part of the code should probably be harmonized.

tgreenx:
## Purpose

This PR proposes an update of function `Zonemaster::Engine::Test::run_one()`. See the **Changes** section below.

## Context

Fixes #1278 and #1284

## Changes

- Harmonize input arguments (with respect to `Zonemaster::Engine::Test::run_all_for()` and `Zonemaster::Engine::Test::run_module()`)
- Remove message tag `TEST_ARGS`
- Add conditional check for Test Case - trying to run a Test Case disabled by the profile will now return an error.
   --> **note that this was a feature only used in CLI and is now moved there: https://github.com/zonemaster/zonemaster-cli/pull/361**
- Add `basic{01,02}` to the default profile

## How to test this PR

Tests should pass.

Also, attempting to run a Test Case disabled by the current profile will now return an error instead of running the test:
```
$ perl -MZonemaster::Engine -E 'Zonemaster::Engine::Profile->effective->set( 'test_cases', [ "address01" ] ); my @res = Zonemaster::Engine->test_method("Address", "address01", "zonemaster.net"); say $_ for @res; say "nothing" unless @res'

System:Unspecified:START_TIME string=2023-12-04 16:18:16 +0100; time_t=1701703096
System:Unspecified:TEST_TARGET module=Address; testcase=address01; zone=zonemaster.net
Address:Address01:TEST_CASE_START testcase=Address01
Address:Address01:NO_IP_PRIVATE_NETWORK
Address:Address01:TEST_CASE_END testcase=Address01

$ perl -MZonemaster::Engine -E 'Zonemaster::Engine::Profile->effective->set( 'test_cases', [ "address01" ] ); my @res = Zonemaster::Engine->test_method("Address", "address02", "zonemaster.net"); say $_ for @res; say "nothing" unless @res'

nothing
```

### DIFF
--- a/share/profile.json
+++ b/share/profile.json
@@ -524,6 +524,8 @@
         "address01",
         "address02",
         "address03",
+        "basic01",
+        "basic02",
         "basic03",
         "connectivity01",
         "connectivity02",
--- a/share/profile.yaml
+++ b/share/profile.yaml
@@ -25,6 +25,8 @@ test_cases:
   - address01
   - address02
   - address03
+  - basic01
+  - basic02
   - basic03
   - connectivity01
   - connectivity02


matsduf (changes requested):
I disagree with the change.

I do not understand why it should be impossible to run a specific test if it is not listed in the current profile. What problem does that solve?

pnax:
> I disagree with the change.

To better understand your position, do you disagree with all the changes in this PR or only with the part that is described as "_Add conditional check for Test Case - trying to run a Test Case disabled by the profile will now return an error._" in the PR's description?

matsduf:
> > I disagree with the change.
> 
> To better understand your position, do you disagree with all the changes in this PR or only with the part that is described as "_Add conditional check for Test Case - trying to run a Test Case disabled by the profile will now return an error._" in the PR's description?

Yes, the change that I am against is 
> * Add conditional check for Test Case - trying to run a Test Case disabled by the profile will now return an error.

I have seen no explanation of the problem that it would solve so I do not understand why the flexibility should be removed.

The following sounds fine:
> * Harmonize input arguments (with respect to Zonemaster::Engine::Test::run_all_for() and
> Zonemaster::Engine::Test::run_module())

The following I do not know what it changes. I probably affects the output, but I do not understand how. Could be fine.
> * Remove message tag TEST_ARGS

matsduf:
> I responded in [#1284 (comment)](https://github.com/zonemaster/zonemaster-engine/issues/1284#issuecomment-1700599475). We can continue the discussion in the issue if you want. Other opinions are welcome - to me it is logical to make that change, the configuration (profile) should take precedence in all cases.

What problem do you see by allowing the user to run an unlisted test case? Why is this a change to the better?

pnax:
> What problem do you see by allowing the user to run an unlisted test case? Why is this a change to the better?

On one hand we can see it as a feature. On the other hand, if you deploy Zonemaster with a custom profile, it allows to bypass it and run a specific test.
Say you want to have Zonemaster installed that runs only one testcase and use the following configuration:
```
$ cat /etc/zonemaster/profile.zone01.json
{ "test_cases": [ "zone01" ] }

$ cat /etc/zonemaster/cli.args
--profile /etc/zonemaster/profile.zone01.json
```
Then you would expect your user to only be able to run testcase `ZONE01`. However with the current implementation you can bypass this limitation with the  `--test` option, for instance `--test zone/zone02`.
If I understand correctly, this PR would solve this and only allow to run testcases defined in the used profile. I think this proposal makes sense.

pnax (dismissed):

matsduf:
> ```
> $ cat /etc/zonemaster/profile.zone01.json
> { "test_cases": [ "zone01" ] }
> 
> $ cat /etc/zonemaster/cli.args
> --profile /etc/zonemaster/profile.zone01.json
> ```
> 
> Then you would expect your user to only be able to run testcase `ZONE01`. However with the current implementation you can bypass this limitation with the `--test` option, for instance `--test zone/zone02`. If I understand correctly, this PR would solve this and only allow to run testcases defined in the used profile. I think this proposal makes sense.

The limitation is not harder than you can create your own profile without any superuser privileges:

```
$ cat  ~/.zonemaster/profile.zone0102.json
{ "test_cases": [ "zone01", "zone02" ] }
 
$ cat ~/.zonemaster/cli.args
--profile ~/.zonemaster/profile.zone0102.json
```

I see no reason to remove the flexibility that we have today. You can test a test case not included in the profile without having to include it.

pnax:
> The limitation is not harder than you can create your own profile without any superuser privileges:

I agree. And I think this would be a good practice to do so (and not rely on a hidden _feature_).

> You can test a test case not included in the profile without having to include it.

I think that for the time being I disagree with your point of view.
(Also by default, Zonemaster uses its internal profile which has all testcases defined. So the use case we're discussing about only happens on setups with an explicitly changed profile.)

matsduf:
> > The limitation is not harder than you can create your own profile without any superuser privileges:
> 
> I agree. And I think this would be a good practice to do so (and not rely on a potential hidden _feature_).

Firstly, it means that your argument above is not valid. Secondly, it is not a hidden feature.

I still do not see what problem it would solve to introduce to limit the flexibility.

mattias-p:
This is the [current documentation](https://metacpan.org/pod/Zonemaster::Engine::Profile#test_cases):

> #### test_cases
>
> An arrayref of names of implemented test cases as listed in the [test case specifications](https://metacpan.org/pod/%20https:#/github.com/zonemaster/zonemaster/tree/master/docs/specifications/tests/ImplementedTestCases.md). Default is an arrayref listing all the test cases.
>
> Specifies which test cases to consider when a test module is asked to run of all of its test cases.
>
> Test cases not included here can still be run individually.
>
> The test cases basic00, basic01 and basic02 are always considered no matter if they're excluded from this property. This is because part of their function is to verify that the given domain name can be tested at all.

If we change the behavior we also need to update the documentation. Since this behavior is stated so explicitly I'd say it's a breaking change.

I don't think the current behavior has a lot of merit other than the fact that it's documented. I don't believe anyone consciously designed it that way. It's just a weird design. When we wrote the documentation for the profile properties our intention was just to document whatever the current behavior happened to be, and then possibly improve it later once the interface was well understood.

I think the change suggested in this PR is an improvement. It's a simple clean model that a test case is either available in a profile or not. I guess it would be helpful to have different error messages when attempting to run a completely unrecognized test case and when attempting to run a test case that's disabled in the effective profile.

While we're on the topic of changing the semantics of `test_cases` I'd like to suggest an additional change. Instead of letting Zonemaster Engine implicitly add `basic00`, `basic01` and `basic02` to `test_cases`, we should require that they are explicitly included. This raises the bar slightly when specifying a profile, but it makes things much clearer for whoever reads it.

lib/Zonemaster/Engine.pm
```diff
@@ -260,7 +260,7 @@ Runs all available tests and returns a list of L<Zonemaster::Engine::Logger::Ent
 
 Runs all available tests for the zone with the given name in the specified module.
 
-=item test_method($module, $method, @arguments)
+=item test_method($module, $method, $name)
 
 Run one particular test method in one particular module. The requested module must be in the list of active loaded modules (that is, not the Basic
```
marc-vanderwal, line 265:
```suggestion
Run one particular test method in one particular module for one particular zone. The requested module must be in the list of active loaded modules (that is, not the Basic
```

marc-vanderwal (commented):
Looks fine. Might as well update or improve the documentation a bit for the affected code.

mattias-p (approved):

pnax:
Note: once this is done, and especially the fact that the Basic test cases will now be defined in the profile, this _hack_ in the Backend could be removed:
```perl
        # BASIC methods are always run: Basic0{0..4}
        my $nbr_testcases_planned = 5;
```
https://github.com/zonemaster/zonemaster-backend/blob/b8113130f3d18512e232923903f8670e8f0e2727/lib/Zonemaster/Backend/TestAgent.pm#L98-L99

share/profile.json
```diff
@@ -500,6 +499,9 @@
         "address01",
         "address02",
         "address03",
+        "basic00",
```
mattias-p, line 502: Once #1308 is merged new references to basic00 need to removed from here. Or if this PR is merged before #1308 that one needs to be updated to remove these lines.

mattias-p (approved):
Looks good.

matsduf:
If I create a profile with Zone09 only and run that with `zonemaster-cli --profile PROFILE` then besides Zone09 also Basic01-02 will be run. If I instead run `zonemaster-cli --test Zone/Zone09` then only Zone09 is run.

I want to keep the behavior of `--test`, especially when developing test zones for test scenarios for test cases. I just want to make sure that the zone has the expected behavior for the specific test case. The test zones are only created for the specific test case.

When unit tests are run for test cases, are also Basic1-2 run?

Will this PR removed the current behavior of `--test`?

pnax (dismissed):
LGMT. Just a remark, I needed to rebase on `develop` to test with the develop branch from Zonemaster-CLI.

> If I create a profile with Zone09 only and run that with zonemaster-cli --profile PROFILE then besides Zone09 also Basic01-02 will be run. If I instead run zonemaster-cli --test Zone/Zone09 then only Zone09 is run.

Maybe we need to clarify how Basic tests are run. I'm okay with all following proposal (and I might be missing other possibilities):
* keep this PR as it is and update the documentation/code on Basic usage later
* remove `basic0{0..3}` from the profile (but then you won't be able to run them specifically)
* update this PR with either
   * clear documentation on how Basic test cases are run
   * updating the code so that is Basic test cases are not part of the profile, then they should not be run

matsduf:
If the issues in https://github.com/zonemaster/zonemaster-cli/pull/361 are fixed, and it is merged in combination with this, it will be fine.

matsduf (dismissed):

marc-vanderwal (approved):

mattias-p (approved):

pnax:
### v2023.2 release testing

Tested with the "How to test" section and works as expected.

PR #1297, opened 2023-09-28
Betterize new unit testing framework

tgreenx:
## Purpose

This PR proposes an update to the new unit testing framework that is being put in place. It aims at improving the content of each test case unit test file (`t/Test-*.t`) by simplifying them. It does so with the following:

- A sizable amount of duplicate code is moved to a new helper module.
- A hash holding test case scenarios is used (instead of a `__DATA__` token at the end of the file)
- All relevant, modifiable content is moved at the start of the file (within `###########` _banners_).

With this PR, current test case unit test files (in the new framework) are reduced by about 30%.

I believe these changes are desirable to ease the migration process of other test case unit test files.

## Context

Inspired from https://github.com/zonemaster/zonemaster-engine/pull/1287

## Changes

- Create helper module `t/TestUtil.pm`
- Create helper function `TestUtil::perform_testcase_testing()`
- Refactoring

## How to test this PR

Tests should pass.

### DIFF
--- a/t/Test-dnssec16.t
+++ b/t/Test-dnssec16.t
@@ -14,7 +14,7 @@ BEGIN {
 }
 
 ###########
-# dnssec16
+# dnssec16 - https://github.com/zonemaster/zonemaster/blob/master/docs/public/specifications/test-zones/DNSSEC-TP/dnssec16.md
 my $test_module = q{DNSSEC};
 my $test_case = 'dnssec16';
 
@@ -27,7 +27,8 @@ Zonemaster::Engine::Recursor->add_fake_addresses( '.',
 );
 
 # Test zone scenarios
-# Format: { SCENARIO_NAME => [ zone_name, [ MANDATORY_MESSAGE_TAGS ], [ FORBIDDEN_MESSAGE_TAGS ], testable ] }
+# - Documentation: L<TestUtil/perform_testcase_testing()>
+# - Format: { SCENARIO_NAME => [ zone_name, [ MANDATORY_MESSAGE_TAGS ], [ FORBIDDEN_MESSAGE_TAGS ], testable ] }
 my %subtests = (
     'CDS-INVALID-RRSIG' => [
         q(cds-invalid-rrsig.dnssec16.xa),
--- a/t/Test-zone09-1.t
+++ b/t/Test-zone09-1.t
@@ -14,7 +14,7 @@ BEGIN {
 }
 
 ###########
-# zone09
+# zone09 - https://github.com/zonemaster/zonemaster/blob/master/docs/public/specifications/test-zones/Zone-TP/zone09.md
 my $test_module = q{Zone};
 my $test_case = 'zone09';
 
@@ -27,7 +27,8 @@ Zonemaster::Engine::Recursor->add_fake_addresses( '.',
 );
 
 # Test zone scenarios
-# Format: { SCENARIO_NAME => [ zone_name, [ MANDATORY_MESSAGE_TAGS ], [ FORBIDDEN_MESSAGE_TAGS ], testable ] }
+# - Documentation: L<TestUtil/perform_testcase_testing()>
+# - Format: { SCENARIO_NAME => [ zone_name, [ MANDATORY_MESSAGE_TAGS ], [ FORBIDDEN_MESSAGE_TAGS ], testable ] }
 my %subtests = (
     'ROOT-EMAIL-DOMAIN' => [
         q(.),
--- a/t/Test-zone09.t
+++ b/t/Test-zone09.t
@@ -14,7 +14,7 @@ BEGIN {
 }
 
 ###########
-# zone09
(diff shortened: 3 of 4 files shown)


matsduf:
>     * A sizable amount of duplicate code is moved to a new helper module.

I think that is a good idea.

>     * An explicit hash (and keys) holding test case scenarios is used (instead of a `__DATA__` token at the end of the file)

I think the current solution with the __DATA__ token makes it easier to read and write the data section. With the token the data section gets more compact without the loosing readability.

What problem do you see by the token solution?

marc-vanderwal (dismissed):
I like the new data structures for describing the expected outcomes of each scenario.

Firstly, they are self-documenting: keys like `mandatory` and `forbidden` make it crystal clear to me that these are tags that must, or must not appear when running a test against a given zone.

Secondly, the syntax is flexible and there are no hidden surprises, because Perl’s syntax applies.

Thirdly, these structures can be extended by means of additional keys, to accommodate situations that require it.

I also like the idea of the `testable` key, which documents scenarios for which we lack the means of testing automatically for the time being.

matsduf:
First I want to state that I am in favor of finding a format that makes it easy to create and maintain the unit test files, and where a unit module, such as the one that you have created, contains shared code.

> Well, firstly, I simply couldn't appropriately feed the `__DATA__` token to an external function in a structured way. Besides, I see the following advantages to the proposed change:

If you set a package name in the file, the __DATA__ data will be available as PACKAGE::DATA file handle.


>     * It uses an hash structure with explicit key names, so comment blocks can be removed

I do not see the direct need of named keys. A small header or a pointer could solve the information need. A readable and compact format is more valuable.

>     * Test scenario names are added as the top-level key

The scenario name is in the zone name, so that is not needed. The zone name could be the key instead. That would make it shorter. The scenario name is not used for the unit tests.

>     * A `testable` sub-key name is used to discard (and still display) untestable scenarios

Yes, that is really a good idea.

>     * It reinforces point 3 described in this PR --> 'All relevant, modifiable content is moved at the start of the file (within ########### banners)'

Well, I do not think that is important. A standardized header first and then specific data at the end works as well.

>     * Increased flexibility and data manipulation in case of additional data fields

Yes, that could be good, but on the other hand, what we want for the test cases are really fixed format.

> The main downside that I see is that when transforming the specification into the unit test, commas need to be removed when added in the code. But personally I didn't find it to be much of a hassle when using a proper text editor. 

I still think the `__DATA__` makes it easy to create another unit test file. Creating the Perl skeleton requires more work. I just created the unit test files for Consistency05 and Consistency06 (#1303) and for that I could reuse the unit test file for DNSSEC16 and copy the table from the test zone specification and convert it in a few steps. And the section after `__DATA__` is easier to read than the `%subtests`, I think.

Then following would create a more compact format with fewer lines:

```
my %subtests = (
    'cds-invalid-rrsig.dnssec16.xa' => [
        [ qw(DS16_CDS_INVALID_RRSIG) ],
        [ qw(DS16_CDS_MATCHES_NON_SEP_DNSKEY DS16_CDS_MATCHES_NON_ZONE_DNSKEY DS16_CDS_MATCHES_NO_DNSKEY DS16_CDS_NOT_SIGNED_BY_CDS DS16_CDS_SIGNED_BY_UNK\
NOWN_DNSKEY DS16_CDS_UNSIGNED DS16_CDS_WITHOUT_DNSKEY DS16_DELETE_CDS DS16_DNSKEY_NOT_SIGNED_BY_CDS DS16_MIXED_DELETE_CD) ],
        1
    ],
    'cds-matches-no-dnskey.dnssec16.xa' => [
        [ qw(DS16_CDS_MATCHES_NO_DNSKEY) ],
        [ qw(DS16_CDS_INVALID_RRSIG DS16_CDS_MATCHES_NON_SEP_DNSKEY DS16_CDS_MATCHES_NON_ZONE_DNSKEY DS16_CDS_NOT_SIGNED_BY_CDS DS16_CDS_SIGNED_BY_UNKNOWN\
_DNSKEY DS16_CDS_UNSIGNED DS16_CDS_WITHOUT_DNSKEY DS16_DELETE_CDS DS16_DNSKEY_NOT_SIGNED_BY_CDS DS16_MIXED_DELETE_CDS) ],
        1
    ]
    );
```

matsduf (approved):
I approve this PR, but I would like to see the following improvements in documentation, which could be in a follow-up PR:

1. In `TestUtil.pm` the fields of the `%subtests` should be defined and described.  Specifically, it is unclear what the "testable" field does. What happends when true and false, respectively?
2. In the "template" for the `t` file it should be a pointer to  `TestUtil.pm` for the definition of the `%subtests` fields.
3. In the "template" for the `t` file it should be a link to the specification of the scenarios in queestions, i.e. https://github.com/zonemaster/zonemaster/blob/master/docs/public/specifications/test-zones/DNSSEC-TP/dnssec16.md and https://github.com/zonemaster/zonemaster/blob/master/docs/public/specifications/test-zones/Zone-TP/zone09.md, respectively, for the `t` files in this PR.

matsduf (approved):

PR #1305, opened 2023-11-15
CI: update Ubuntu and Perl versions

marc-vanderwal:
## Purpose

This PR updates the versions of Ubuntu and Perl to use in Github actions.

## Context

Discussion during work group meeting.

## Changes

- Use Ubuntu 22.04 instead of 20.04;
- Test against Perl 5.36, 5.30 and 5.26 instead of 5.30, 5.26 and 5.16;
- Reformat PO files according to newer version of gettext provided by Ubuntu 22.04.

## How to test this PR

The Github Action should succeed.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,11 +20,11 @@ jobs:
           - develop
           #- latest
         perl:
-          - '5.32'
+          - '5.36'
+          - '5.30'
           - '5.26'
-          - '5.16'
         runner:
-          - ubuntu-20.04
+          - ubuntu-22.04
 
     runs-on: ${{ matrix.runner }}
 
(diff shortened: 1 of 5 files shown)


marc-vanderwal:
CI fails because of a failing test in `t/po-files.t`. Upgrading Ubuntu from version 20.04 to 22.04 caused an upgrade of gettext from version 0.19.8.1-10build1 to version 0.21-4ubuntu4. The formatting applied by `msgcat` might have slightly changed.

marc-vanderwal:
Seems like I was correct. The new `msgcat` has a better line-breaking algorithm that avoids the awkward line breaks in places like `sur une requête de type "PTR"`, where previously, the line was broken between `PTR` and the closing double quote. Now the line is broken between `type` and `"PTR"`, which is much better. In order for CI to pass, I’m going to have to update the PO files to reflect `msgcat`’s new line-breaking algorithm.

matsduf (approved):

PR #1319, opened 2024-01-19
Migrate away from Moose

### LINKED ISSUE #1318: Remove dependency on Moose
We expect removing dependency on Moose to have two benefits.
1. Make processing faster.
2. Reduce the dependency foot print.

Instead of Moose we should use [Class::Accessor](https://metacpan.org/pod/Class::Accessor) which is more light weight.

mattias-p:
## Purpose

This PR replaces the last remaining uses of Moose in Engine.

## Context

Fixes #1318.

## Changes

* In most cases Moose is replaced with the more light-weight Class::Accessor. But in some cases Class::Accessor doesn't offer much in the way of replacement. In those cases the Moose constructs are replaced with plain Perl.
* The dependency on Moose is removed.

## How to test this PR

Unit tests should pass.


marc-vanderwal (commented):
Looks good to me so far. I hope this change will improve the engine’s performance somewhat.

mattias-p:
I ran the unit tests ten times each on b1d6f90c6ff8df5ce7ad175ca74d8e41396e2e25 (current develop) 4e0a57859a72a6dcea30a8edc0afb8656ef20607 (current 1318-demoose) and found that the minimum running time decreased by about 10% and the average running time decreased by about 9%.

mattias-p:
I ran the unit tests once each for the same commits and found that the memory usage decreased by about 18%. ("Maximum resident set size" as reported by GNU time)

lib/Zonemaster/Engine/Packet.pm
```diff
@@ -1,48 +1,146 @@
...
+
+sub string {
+    my ( $self ) = @_;
+    return $self->packet->string;
+}
+
+sub data {
+    my ( $self ) = @_;
+    return $self->packet->data;
+}
```
marc-vanderwal, line 52: It’s a lot of repetition, it’s true.

Out of curiosity, I tried looking for a solution that involved some metaprogramming to get rid of the repetition. The following code seems to work (well, at least, unit tests pass):

```perl
BEGIN {
    # This block uses metaprogramming techniques to define a number of almost
    # identical methods that are passthrough functions to methods in
    # Zonemaster::LDNS::Packet to be called on the `packet` member of
    # instances of this class.

    no strict 'refs';

    # Accessors taking an optional argument, defaulting to an empty list.
    for my $m (qw(timestamp querytime id opcode rcode edns_version)) {
        *{$m} = sub {
            my ( $self, $value ) = @_;
            return $self->packet->$m( $value // () );
        };
    }

    # Accessors that do not take any arguments
    for my $m (qw(type string data
                  aa ra tc do
                  question answer authority additional
                  edns_size edns_rcode edns_z edns_data has_edns)) {
        *{$m} = sub {
            my ( $self ) = @_;
            return $self->packet->$m;
        }
    }
}
```

I know that touching the symbol table like this is very akin to dark magic in Perl and may scare away newcomers, so I’m also fine with keeping the code as it is, even with the repetition.

marc-vanderwal: Well, you know the saying in Perl: TIMTOWTDI. I like the way this code is indented, and it’s the way I’ve been doing simple accessors like these in my personal projects.

So yeah, you have a point: there is little gain to be had in maintainability to try to make it DRY. That’s why I said I was fine with the non-DRY version.

mattias-p:
@matsduf, @tgreenx, @marc-vanderwal please review again.

lib/Zonemaster/Engine/Packet.pm
matsduf: I think the more compressed alternative that you @mattias-p propose above is better because it is easier to read, and I prefer that to the more magic solution that you @marc-vanderwal suggests. The explicit solution is easier to read and follow.

matsduf:
@mattias-p, you write "Unit tests should pass". Do you think there is enough unit test coverage for this change? This change is hard to review. Make sure that all unit tests work is probably the best test. Could more unit tests be added to verify the correctness of the change?

I think this should be merged soon so that we have a chance to let pass through tests that we do.

I am prepared to approve.

lib/Zonemaster/Engine/Packet.pm
mattias-p: Apparently I forgot to push the commit I made for this, but I pushed it now.

marc-vanderwal:
> @mattias-p, you write "Unit tests should pass". Do you think there is enough unit test coverage for this change? This change is hard to review. Make sure that all unit tests work is probably the best test. Could more unit tests be added to verify the correctness of the change?

I checked out the PR’s branch and ran the test suite through [Devel::Cover](https://metacpan.org/pod/Devel::Cover). According to  the [report for the Zonemaster::Engine::Packet module](https://github.com/zonemaster/zonemaster-engine/files/14914972/Coverage_Zonemaster_Engine_Packet.pdf) (originally HTML, had to print to PDF to be able to upload it to Github), there are a handful of untested methods:
- `querytime`;
- `id`;
- `opcode`;
- `data`;
- `edns_data`;
- `edns_z`.

The `querytime` and `data` methods are only exercised when the Redis cache is in use. The `edns_data` method is used in Nameserver11 and `edns_z` is used in Nameserver12, but the code paths don’t seem to be covered. The remaining methods are entirely unused.

mattias-p:
I added a unit test for the mentioned methods without test coverage. I considered removing the unused `id` and `opcode` methods instead of adding tests for them, but they are documented and I didn't want to break the API. The new test's aren't very interesting, but I did find a nicer way to integrate Test::NoWarnings and it's nice to have a file where you can just add more unit tests for the Engine Packet module.

matsduf (approved):

PR #1322, opened 2024-03-04
Make syntax04 take zone name instead of ns name

### LINKED ISSUE #1321: Bring syntax04 in line with other methods
Today all test methods but syntax04 share the same interface. Syntax04 expects its caller to figure out all the nameservers of the zone to be tested and call it repeatedly with each nameserver name while everyone else just takes the name of the zone to be tested. Moreover this exception is not documented anywhere.

Today this works for Zonemaster::Engine::test_zone() and Zonemaster::Engine::test_module() because they call Zonemaster::Engine::Test::Syntax::all() which meets the special expectation of syntax04. But Zonemaster::Engine::test_method() doesn't do anything special for syntax04, so the responsibility is pushed onto the caller. A notable caller is Zonemaster CLI which also doesn't do anything special for syntax04 so the responsibility is pushed onto the user.

The determination of nameserver names for syntax04 should be moved from Zonemaster::Engine::Test::Syntax::all() into Zonemaster::Engine::Test::Syntax::syntax04().

mattias-p:
## Purpose

Bring syntax04 in line with other methods

## Context

Fixes #1321.

## Changes

The determination of nameserver names for syntax04 is moved from Zonemaster::Engine::Test::Syntax::all() into Zonemaster::Engine::Test::Syntax::syntax04().

## How to test this PR

Find/create zones with NS records with invalid NSDNAME:
* One NSDNAME with a numeric TLD (e.g. `ns1.nic.47`).
* One NSDNAME with a double dash (e.g. `ns1.ns--nic.fr`). 

Run this command (with ZONE replaced with your test zone) and verify that you get complaints about nameserver records.
```sh
$ perl -MZonemaster::Engine -E 'say join "\n", Zonemaster::Engine->test_method("syntax", "syntax04", "ZONE")'
```

Unit tests have been updated to test the converse. I.e. that syntax04 does not complain about zones whose names themselves have these properties.

### DIFF
--- a/lib/Zonemaster/Engine/Test/Syntax.pm
+++ b/lib/Zonemaster/Engine/Test/Syntax.pm
@@ -60,12 +60,8 @@ sub all {
 
     if ( any { $_->tag eq q{ONLY_ALLOWED_CHARS} } @results ) {
 
-        foreach my $local_nsname ( uniq map { $_ } @{ Zonemaster::Engine::TestMethods->method2( $zone ) },
-            @{ Zonemaster::Engine::TestMethods->method3( $zone ) } )
-        {
-            push @results, $class->syntax04( Zonemaster::Engine->zone( $local_nsname ) )
-              if Zonemaster::Engine::Util::should_run_test( q{syntax04} );
-        }
+        push @results, $class->syntax04( Zonemaster::Engine->zone( $zone ) )
+          if Zonemaster::Engine::Util::should_run_test( q{syntax04} );
 
         push @results, $class->syntax05( $zone ) if Zonemaster::Engine::Util::should_run_test( q{syntax05} );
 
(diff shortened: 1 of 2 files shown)


marc-vanderwal (approved):

marc-vanderwal:
Well, actually it seems that this change introduces a regression.

Take `hipotel.fr`, whose NS resource record set currently includes a record whose name contains spaces.

On 2023.2.1, Zonemaster correctly reports the invalid characters in the NS’s name:
Syntax04 complains about 100\\0321\\032443\\032sipdir.online.lync.com.

On latest develop, however, this doesn’t work anymore and Syntax04 reports the domain, instead of the name server name:
Syntax04 does not show any errors anymore

matsduf:
Is this a show stopper?

Actually, I do not understand what the purpose of this PR is.

marc-vanderwal:
From what I understand, it’s a PR that refactors some code. It looked alright on the surface but the affected code doesn’t seem to be covered by unit tests.

PR #1325, opened 2024-03-14
Update test module interface documentation

mattias-p:
## Purpose

I started out making this update because I felt we need a good and up to date definition of the test module interface.

## Context

Having a clear definition of the test module interface will be useful when we design the custom test module feature and in turn when people make use of it.

## Changes

The justification for this PR is the added and updated details. While this PR contains some amount of new details most of the diff comes from formatting and editing and moving things around. Some older language and terminology is also updated to be more precise and contemporary. I may have got a little carried away but I believe the end result is good.

## How to test this PR

N/A


mattias-p:
@matsduf @marc-vanderwal @tgreenx Please review. I've rebased this on top of latest develop.

tgreenx (dismissed):

lib/Zonemaster/Engine/Localization.pod
```diff
@@ -11,41 +13,17 @@ All translation files live in the F<share> directory in the
 L<Zonemaster::Engine> source directory and all commands described here are
 executed from that directory.
 
-=head2 For translators
+=head1 FOR TRANSLATORS
```
matsduf, line 16: The file name is "Localization" but it is only about translation.

lib/Zonemaster/Engine/TestModuleInterface.pod
```diff
@@ -0,0 +1,169 @@
...
+=back
+
+=head2 Code clarity over performance
+
+Strive to make the test code as easily understandable as possible.
+Ideally, it should be possible for a person with only a basic understanding of
+Perl to read the code and verify that it correctly implements the corresponding
+Test Case specifications.
+
+=head2 Internationalization
```
matsduf, line 66: It is not about internationalization. It is related to translation, but mostly about the mechanism where a tag is replaced by human language, humanization?

lib/Zonemaster/Engine/TestModuleInterface.pod
```diff
@@ -0,0 +1,169 @@
...
+
+Make as few assumptions as possible when implementing test modules.
+When the entry method in a test module is called, it can only rely on a handful
+of things about the testing process so far:
+
+=over
+
+=item
+
+The zone object it's given has a parent.
```
matsduf, line 47: Not necessarily. The root zone has no parent. And a undelegated zone can be tested even if it is not possible to determine its parent-to-be.

lib/Zonemaster/Engine/TestModuleInterface.pod
```diff
@@ -0,0 +1,169 @@
...
+
+=item
+
+The zone object it's given has glue records.
```
matsduf, line 51: Not necessarily. But then testing will fail at Basic02.

matsduf (changes requested):
Related issues are referred to as "localization", "translation" and "internationalization". Unnecessary confusion. I suggest that translation is used in all places.

mattias-p:
@matsduf

Regarding "internationalization" and "localization", I didn't make this up. It's actually well established terminology in the industry. See https://en.wikipedia.org/wiki/Internationalization_and_localization.

Regarding the "Make few assumptions" section, I didn't make this up either. It is what our current documentation says. I simply moved it to a better place. It's good that you spotted these flaws in our documentation but I feel they're out of scope for this PR.

matsduf:
Even if "internationalization" and "localization" are established terms the documentation could be less confusing if it concretely explains what it is all about.

mattias-p:
@matsduf I believe the documentation is quite concrete already. It explains what translators and developers need to do and what they need to think about, and I believe we have a decent linking structure that directs people to the right places. I'll add a link to the Wikipedia page. Would that be enough or do you have anything else in mind?

lib/Zonemaster/Engine/TestModuleInterface.pod
marc-vanderwal: “Internationalization” is the industry-standard term for the process of modifying software so as to make it adaptable to different languages and customs (e.g. right-to-left scripts, formatting of numbers, etc.). The use of message tags instead of hard-coded messages in English is the way Zonemaster implements internationalization.

matsduf:
@mattias-p, what is the status of this PR?

PR #1327, opened 2024-03-16
Fixes a bug in Redis global cache triggered when TTL is 0

### LINKED ISSUE #1326: Fatal error in Redis cache
I followed the instructions in "[Configuration: Global cache in Zonemaster-Engine]" (develop branch) on Ubuntu 22.04 and created and updated profile where the following has been added:
```
$ head /etc/zonemaster/profile.json
{
    "cache": {
        "redis": {
            "server": "127.0.0.1:6379",
            "expire": 300
        }
    },
    "asn_db" : {
        "style" : "Cymru",
        "sources" : {
```

The Redis daemon was running and I ran zonemaster-cli on zone sunet.se with no specific errors (not using the custom profile). I ran it again with the custom profiile:
```
$ zonemaster-cli sunet.se --profile /etc/zonemaster/profile.json
Loading profile from /etc/zonemaster/profile.json.

Seconds Level    Message
======= ======== =======
(...)
   2.53 CRITICAL Fatal error in Zonemaster::Engine::Test::Nameserver: [set] ERR invalid expire time in set,  at /usr/local/share/perl/5.34.0/Zonemaster/Engine/Nameserver/Cache/RedisCache.pm line 96.
(...)
```
I turned the Redis daemon off and got a different errror
```
$ zonemaster-cli sunet.se --profile /etc/zonemaster/profile.json
Loading profile from /etc/zonemaster/profile.json.

Seconds Level    Message
======= ======== =======
Looks OK.
Could not connect to Redis server at 127.0.0.1:6379: Connection refused at /usr/local/share/perl/5.34.0/Zonemaster/Engine/Nameserver/Cache/RedisCache.pm line 50.
```

[Configuration: Global cache in Zonemaster-Engine]: https://github.com/zonemaster/zonemaster/blob/develop/docs/public/configuration/global-cache.md

matsduf:
## Purpose

This PR fixes the bug described in #1326 by setting the lowest TTL value to 1.

This should be seen as a temporary solution. There should probably be a shortest time that messages are cached in Redis to make sure that in the normal case the same query should not be sent more than once when running a test. In issue #1326 we can see that the TTL of "iis.se. CDS" is 0, and a query for that would be sent several times if not cached.

That shortest time should be configurable in the profile.

## Context

Fixes #1326.

## How to test this PR

Find a zone that returns TTL=0 for some record that Zonemaster queries for a verify that there is no fatal error of the type describe in #1326.

### DIFF
--- a/lib/Zonemaster/Engine/Nameserver/Cache/RedisCache.pm
+++ b/lib/Zonemaster/Engine/Nameserver/Cache/RedisCache.pm
@@ -93,6 +93,8 @@ sub set_key {
             }
         }
         $ttl = $ttl < $redis_expire ? $ttl : $redis_expire;
+        # Redis requires cache time to be greater than 0.
+        $ttl = 1 if $ttl == 0;
         $self->redis->set( $key, $msg, 'EX', $ttl );
     } else {
         $self->redis->set( $key, '', 'EX', $ttl );


marc-vanderwal:
Why don’t we skip caching altogether if the TTL is 0?

matsduf:
> Why don’t we skip caching altogether if the TTL is 0?

That is an alternative to work-around the bug, but that will decrease caching in some cases compared to "local cache". With the discovered example, there is more than one test case querying for the CDS record, and with "local cache" that query only goes on the wire once. The global caching should be long enough to survive a test, in the normal case.

Another answer is that I started my investigation by setting the minimum cache value to 1 to see if that solved the issue, and that works as well.

hannaeko:
> > Why don’t we skip caching altogether if the TTL is 0?
> 
> That is an alternative to work-around the bug, but that will decrease caching in some cases compared to "local cache". With the discovered example, there is more than one test case querying for the CDS record, and with "local cache" that query only goes on the wire once. The global caching should be long enough to survive a test, in the normal case.
> 
> Another answer is that I started my investigation by setting the minimum cache value to 1 to see if that solved the issue, and that works as well.

You probably should skip inserting the key in Redis if the TTL is 0, the Redis cache still has a "local cache" so the answer will still be cached for the duration of the test anyway, same as the previous cache system.

https://github.com/zonemaster/zonemaster-engine/blob/17688f7995e16dd9f5a4bec2a1437fc7804b31f0/lib/Zonemaster/Engine/Nameserver/Cache/RedisCache.pm#L74

matsduf:
> You probably should skip inserting the key in Redis if the TTL is 0, the Redis cache still has a "local cache" so the answer will still be cached for the duration of the test anyway, same as the previous cache system.

Thanks! I will create an update. I did not realize that the local cache also was used.

Issue #1337, opened 2024-04-04
Custom test modules


mattias-p:
Depends on:
 * #1336

Zonemaster::Engine::Test/run_for_all() needs to know about all the test modules
it should run and in which order to run them. Today it makes this determination
by parsing a file from disk and prepending the built-in Basic test module. To
run an additional test module you need to include it in the file.

This mechanism should be replaced like so:
 * Add a class method to Zonemaster::Engine::Profile that returns the
   Perl package that was provided to the test module registration method.
 * Add a class method to Zonemaster::Engine::Profile that returns the of test
   module names. The list should contain an entry for each registered test
   module. Test modules under 'Zonemaster::Engine::Test::' be ordered before any
   others. Within these two groups, test modules should be listed in the order
   they were registered.
 * Update Zonemaster::Engine::Test/modules() to call the new Profile
   method instead of using the old parse and prepend algorithm.
 * Update the test running methods in Zonemaster::Engine::Test to query the
   Profile module for the ordered set of test module names and their
   corresponding Perl modules.

mattias-p:
I'm not super-fond of the special handling of modules under 'Zonemaster::Engine::Test::'. The idea is that it should be impossible for 3rd party modules to get loaded before the built-in ones. Perhaps there is a better way to ensure this.

PR #1339, opened 2024-05-06
Change supported way of extracting default profile

### LINKED ISSUE #1333: Dynamically generated default profile file
Today we have references into the filesystem for the default profile file.
Instead we should tell people to run `perl -MZonemaster::Engine::Profile -E 'say
Zonemaster::Engine::Profile->default->to_json' | jq -S .`.

mattias-p:
## Purpose

This PR replaces the supported method of extracting the default profile from Zonemaster Engine.

## Context

Together with zonemaster/zonemaster#1262 this PR fixes #1333, which is a step towards #1337.

## Changes

This is a breaking change because it removes the installed profile.json file from the public interface. Instead of inspecting this file directly, users are instructed to extract it from Zonemaster::Engine.

## How to test this PR

Test the command to extract values from the default profile `lib/Zonemaster/Engine/Profile.pm`.

### DIFF
--- a/lib/Zonemaster/Engine/Profile.pm
+++ b/lib/Zonemaster/Engine/Profile.pm
@@ -528,10 +528,7 @@ Update it to change the configuration.
 The effective profile is initialized with the default values declared
 in the L</PROFILE PROPERTIES> section.
 
-For the effective profile, all properties are always set (to valid
-values).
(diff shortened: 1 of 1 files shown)


matsduf (approved):

matsduf:

Works as documented.

PR #1340, opened 2024-05-06
Updates TestUtil.pm for unit tests

matsduf:
## Purpose

This PR introduces two changes to function in TestUtil.pm:
*  Support of undelegated data (both NS and DS) for unit tests.
* Makes the function more flexible to avoid long lists of tags when not needed.
* Do not use diag/fail when there is an error in the `t` file. Use croak() instead.
* More verifications of data.

The unit test files using TestUtil.pm have been updated to match the new format.

Five new scenarios have been added to Consistency05 and two to Consistency06. Both updates are to unit tests that uses the new function in TestUtil.pm that supports undelegated data.

The following new scenario fail:
* MULT-SOA-MNAMES-NO-DEL-UNDEL-2 (Consistency06)

## Context

The update is needed for https://github.com/zonemaster/zonemaster/pull/1255.

## How to test this PR

Review TestUtil.pm. If the unit test files using TestUtil.pm then they should be fine.


t/TestUtil.pm
```diff
@@ -63,37 +91,92 @@ sub perform_testcase_testing {
...
+             $forbidden_message_tags,
+             $undelegated_ns,
+             $undelegated_ds
+            ) = @{ $subtests{$scenario} };
+
+        if ( ref( $testable ) ne '' ) {
+            diag("Scenario $scenario: Type of testable must not be a reference");
+            fail("Testable is of the correct type");
+            next;
+        }
```
mattias-p, line 123: This is problematic for a couple of reasons, but the biggest one is that the diag() and fail() methods are uses here for complaining about the test script. They should only be used for complaining about the object under test. When complaining about the test script it's better to use croak().

There are more of these below.

I do realize you followed the established pattern, but it would be a really good to break this pattern.

t/TestUtil.pm
```diff
@@ -63,37 +91,92 @@ sub perform_testcase_testing {
...
 
         if ( ref( $zone_name ) ne '' ) {
             diag("Scenario $scenario: Type of zone name must not be a reference");
             fail("Zone name is of the correct type");
             next;
         }
 
-        if ( ref( $mandatory_message_tags ) ne 'ARRAY' ) {
+        if ( ref( $all_test_case_tags ) ne 'ARRAY' ) {
```
mattias-p, line 131: If we care to verify that it's an array, shouldn't we also check the type of the array elements?

There are more of these below.

t/TestUtil.pm
```diff
@@ -44,8 +44,36 @@ and if the test scenario is testable, it runs the specified test case and checks
...
+a string (zone name)
+
+=item *
+an array of strings (all test case message tags)
+
+=item *
+an array of strings (mandatory message tags)
+
+=item *
+an array of strings (forbidden message tags)
```
mattias-p, line 64: Maybe spell out that `undef` is a valid value for the mandatory and forbidden fields in some circumstances?

t/TestUtil.pm
```diff
@@ -103,6 +186,27 @@ sub perform_testcase_testing {
         }
 
         subtest $scenario => sub {
+
+            if ( @$undelegated_ns ) {
+                my %hash;
+                foreach my $nsexp ( @$undelegated_ns ) {
+                    my ($ns, $ip) = split m(/), $nsexp;
+                    $hash{$ns} = [] unless exists $hash{$ns};
```
mattias-p, line 194: There's a nice way to simplify this line:
```suggestion
                    $hash{$ns} //= [];
```

t/Test-nameserver15.t
```diff
@@ -17,6 +17,7 @@ BEGIN {
 # nameserver15 - https://github.com/zonemaster/zonemaster/blob/master/docs/public/specifications/test-zones/Nameserver-TP/nameserver15.md
 my $test_module = 'Nameserver';
 my $test_case = 'nameserver15';
+my @all_tags = qw(N15_NO_VERSION_REVEALED N15_ERROR_ON_VERSION_QUERY N15_SOFTWARE_VERSION N15_WRONG_CLASS);
```
mattias-p, line 20: Maybe put each list element on its own line? That should improve readability a lot.

t/TestUtil.pm
```diff
@@ -44,8 +44,36 @@ and if the test scenario is testable, it runs the specified test case and checks
...
+=item *
+an array of strings (forbidden message tags)
+
+=item *
+an array of name server expressions for undelegated name servers
+
+=item *
+an array of DS expressions for "undelegated" DS
+
+=back
```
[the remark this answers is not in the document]
matsduf: I think that I understand what you mean, but that will make it more complex to create e.g. `t/Test-nameserver15.t` and harder to read it. Please note that "all test case message tags" will just be a variable, and that the two last arguments will not be used in most cases.

t/Test-nameserver15.t
matsduf: Well, I do agree. I will update.

t/TestUtil.pm
matsduf: This is not new code. It is the same code as in the current version, but moved up. Could you suggest the replacement code?

t/TestUtil.pm
matsduf: This is not new code. It is the same code as in the current version, but maybe moved. What you suggest is that all elements in the array should be matching the tag format? I guess that could be done. Should that also use croak() instead of diag/fail? If so, what is the benefit of that?

t/TestUtil.pm
matsduf: Good point. I thought I had included that. Updated now.

mattias-p: I see that you added descriptions for the semantics undef in these fields. That's great. But this bullet list describing the syntax also needs to be updated to make it clear that undef is an accepted value.

t/TestUtil.pm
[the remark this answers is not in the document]
mattias-p: I'm not sure how you're measuring complexity in the call site. The calling code would be easier to read and less error prone because you wouldn't run the risk of mixing up the arguments. There *would* be more code inside perform_testcase_testing() but this is a quite good investment.

t/TestUtil.pm
mattias-p: Just replace the `diag`, `fail` and `next` with a `croak`.

matsduf:
@mattias-p, please re-review.

t/TestUtil.pm
```diff
@@ -63,37 +96,106 @@ sub perform_testcase_testing {
...
+
+        if ( ref( $undelegated_ns ) ne 'ARRAY' ) {
+            diag("Scenario $scenario: Incorrect reference type of undelegated name servers expressions. Expected: ARRAY");
+            fail("Undelegated name server expressions are of the correct type");
+            next;
+        }
+
+        if ( ref( $undelegated_ds ) ne 'ARRAY' ) {
+            diag("Scenario $scenario: Incorrect reference type of undelegated name servers expressions. Expected: ARRAY");
+            fail("Undelegated name server expressions are of the correct type");
```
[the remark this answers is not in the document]
matsduf: I'll fix this.

t/TestUtil.pm
```diff
@@ -103,6 +205,27 @@ sub perform_testcase_testing {
         }
 
         subtest $scenario => sub {
+
+            if ( @$undelegated_ns ) {
+                my %hash;
```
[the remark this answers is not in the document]
matsduf: I'll do.

t/Test-zone09.t
```diff
@@ -28,85 +41,122 @@ Zonemaster::Engine::Recursor->add_fake_addresses( '.',
...
+#   ] }
+#
+# - One of MANDATORY_MESSAGE_TAGS and FORBIDDEN_MESSAGE_TAGS may be undefined.
+#   See documentation for the meaning of that.
+
 my %subtests = (
     'NO-RESPONSE-MX-QUERY' => [
-        q(no-response-mx-query.zone09.xa),
+        1,
+        q(no-response-mx-query.zone09.xa),#
```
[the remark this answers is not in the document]
matsduf: I tagged scenarios where MANDATORY and FORBIDDEN did not cover all tags. I'll make it explicit.

t/TestUtil.pm
matsduf: Fixed.

matsduf:
@tgreenx and @mattias-p, please re-review.

marc-vanderwal:
I’ve had a look. I see no issues that haven’t already been brought up by @tgreenx.

t/TestUtil.pm
```diff
@@ -37,64 +39,158 @@ unknown to the include path @INC, it can be including using the following code:
...
+an array of strings (forbidden message tags) or undef
+
+=item *
+an array of name server expressions for undelegated name servers
+
+=item *
+an array of DS expressions for "undelegated" DS
+
+=back
+
```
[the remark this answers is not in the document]
matsduf: @tgreenx, what do you mean by "If any array is empty, then it will be ignored"? Technically both could be empty, but both cannot be undef.

t/Test-zone11.t
```diff
@@ -15,146 +15,208 @@ BEGIN {
...
+#   ] }
+#
+# - One of MANDATORY_MESSAGE_TAGS and FORBIDDEN_MESSAGE_TAGS may be undefined.
+#   See documentation for the meaning of that.
+
 my %subtests = (
     'NO-TXT' => [
-        q(no-txt.zone11.xa),
+        1,
+        q(no-txt.zone11.xa),# All tags in "@all_tags" are not defined for this scenario
```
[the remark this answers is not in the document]
matsduf: Done.

matsduf:
@marc-vanderwal  and @tgreenx, please re-review after my latest updates.

t/TestUtil.pm
marc-vanderwal: I don’t think it’s useful to check the type of the array elements: because the array is usually prepared using `qw()`, it is reasonable to assume that each element can be treated as a string. I’m not saying this kind of thorough checking shouldn’t be done; only that I don’t mind if it isn’t done.

However, I do agree that we should use `croak` instead of `diag` and `fail` here. We don’t want to raise an issue with the domain under test here, but with the _code_ that tests the code. In that case, aborting early gives a stronger signal that the problem is with the testing code.

t/TestUtil.pm
```diff
@@ -63,37 +91,92 @@ sub perform_testcase_testing {
...
 
-        if ( ref( $mandatory_message_tags ) ne 'ARRAY' ) {
+        if ( ref( $all_test_case_tags ) ne 'ARRAY' ) {
+            diag("Scenario $scenario: Incorrect reference type of all test case tags. Expected: ARRAY");
+            fail("Mandatory message tags are of the correct type");
+            next;
+        }
+
+        if ( ! defined( $mandatory_message_tags ) and !defined( $forbidden_message_tags ) ) {
+            diag("Scenario $scenario: Not both array of mandatory tags and array of forbidden tags can be undefined");
```
marc-vanderwal:
```suggestion
            diag("Scenario $scenario: Arrays of mandatory tags and forbidden tags cannot be both undefined");
```

t/TestUtil.pm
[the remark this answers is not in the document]
marc-vanderwal: I understand @tgreenx’s suggestion but do think it could be better worded. How about this:
```
If the array of mandatory message tags is C<undef>, it means “any message tag
not explicitly forbidden must be emitted”.

If the array of forbidden message tags is C<undef>, it means “any message tag
not explicitly allowed must not be emitted”.

Both of the above arrayrefs cannot be simultaneously C<undef>.
```

marc-vanderwal (commented):
Looks fine. I do think that `croak` is better than `diag`/`fail` when there is a problem with the data in the test script. It’s better to have a specific failure mode for problems with the testing code, that can’t be confused with the outcome of a failing unit test suite for reasons unrelated to the testing code itself.

t/TestUtil.pm
matsduf: Your proposal must be based on old code. The current code uses "croak" for such tests.

t/TestUtil.pm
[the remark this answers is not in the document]
matsduf: Fixed.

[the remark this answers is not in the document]
matsduf: Maybe it should be forbidden to have both arrays empty.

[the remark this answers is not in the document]
marc-vanderwal: Yes. Having both arrays empty means that we don’t want any message tag to be output, while simultaneously not caring whether a message tag was output. That’s a self-contradicting statement.

The use case where we don’t want any message to be output is covered by having `$mandatory_message_tags = []` and `$forbidden_message_tags = undef`. And the use case where we don’t care at all whether any message tag was output (though in my opinion extremely unlikely, because why test if we don’t examine the outcome?) is covered by not testing the scenario, or at least not with this framework.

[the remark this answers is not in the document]
matsduf: The change in commit ed80af88605fb2165d834a47014eaea8b04e57ee was not correct. Since empty-empty is meaningless it should not be in the `t` file. It could be a mistake. I'll make it not permitted instead.

matsduf:
I have commited the proposal from @tgreenx and I have updated the documentation so that it states that empty-empty is not permitted. @marc-vanderwal and @tgreenx, please re-review.

t/TestUtil.pm
```diff
@@ -75,6 +75,9 @@ undefined. If the mandatory message tag array is undefined, then it will be gene
 all message tags not included in the forbidden message tag array. The same mechanism is used if the
 forbidden message tag array is undefined.
 
+The arrays of mandatory message tags and forbidden message tags, respectively, can be empty, but not
+both. At least on of the arrays must be non-empty.
```
marc-vanderwal, line 79:
```suggestion
both. At least one of the arrays must be non-empty.
```

matsduf:
@marc-vanderwal and @tgreenx, please re-review.

marc-vanderwal (approved):

Issue #1341, opened 2024-05-06
"my" variable $p masks earlier declaration in same scope


matsduf:
```
"my" variable $p masks earlier declaration in same scope at /usr/local/share/perl/5.34.0/Zonemaster/Engine/Nameserver.pm line 363.
```

PR #1342, opened 2024-05-07
Update minimum Perl version to 5.16.0 and add missing 'warnings'

tgreenx:
## Purpose

This PR updates Zonemaster-Engine modules which were missing either a minimum Perl version or the `warnings` Perl library.

Moreover, the minimum Perl version is raised from `5.14.2` to `5.16.0` to match [our current documentation](https://github.com/zonemaster/zonemaster/blob/v2023.2.1/docs/public/installation/prerequisites.md#supported-perl-versions).

As a drive-by change, this also fixes a variable declaration error from #1341.

## Context

Spotted while fixing #1341 

## How to test this PR

Check that no Zonemaster-Engine Perl modules are missing these libraries:

```
cd zonemaster-engine/
find lib -type f -name '*.pm' \( -exec grep -q '^use warnings;$' '{}' \; -or -print \)
find lib -type f -name '*.pm' \( -exec grep -q '^use v5.16.0;$' '{}' \; -or -print \)
```

Note that with this change, other errors have started to show and will be fixed in another PR: #1343

### DIFF
--- a/lib/Zonemaster/Engine.pm
+++ b/lib/Zonemaster/Engine.pm
@@ -1,8 +1,6 @@
 package Zonemaster::Engine;
 
-use 5.014002;
-
-use strict;
+use v5.16.0;
 use warnings;
 
 use version; our $VERSION = version->declare("v5.0.0");
@@ -19,6 +17,7 @@ BEGIN {
 
 use Class::Accessor "antlers";
 use Carp;
+
 use Zonemaster::Engine::Nameserver;
 use Zonemaster::Engine::Logger;
 use Zonemaster::Engine::Profile;


matsduf (dismissed):

lib/Zonemaster/Engine.pm
```diff
@@ -1,9 +1,12 @@
 package Zonemaster::Engine;
 
-use version; our $VERSION = version->declare("v5.0.0");
-
 use 5.014002;
 
+use strict;
```
mattias-p, line 5: When we have `use VERSION` >= 5.12 we don't need to specify `use strict` because it's implicitly enabled. Once we reach 5.30 `use warnings` will be implicit too.

https://perldoc.perl.org/functions/use#use-VERSION

mattias-p (approved):

marc-vanderwal:
I suggest an alternative test procedure:

In the root of Zonemaster::Engine’s code repository, run the following command:
```
find lib -type f -name '*.pm' \( -exec grep -q '^use warnings;$' '{}' \; -or -print \)
```
Expect no output. Files that appear are missing a `use warnings;` line.

marc-vanderwal (approved):

marc-vanderwal:
I see that a test in `t/Test.t` is now failing, but this definitely seems to be a bug uncovered by the use of `use warnings;`. Not a regression, but simply a latent problem waiting to be fixed.

mattias-p (approved):
I checked https://perldoc.perl.org/feature#FEATURE-BUNDLES and AFAICT the bump from 5.14 to 5.16 shouldn't break anything.

matsduf:
@tgreenx, can you check this? Will rebasing solve the test issue?

matsduf (approved):

matsduf:

I have verified in the code, but I have no installation of Perl v5.15 or lower to test with, but I do not think it is needed.

PR #1347, opened 2024-05-15
Fix Zonemaster::Engine::Translator’s instance() method

marc-vanderwal:
## Purpose

This PR fixes a bug in Zonemaster::Engine::Translator that causes problems in Zonemaster::Backend::Translator (in Backend, not Engine), which inherits from Zonemaster::Engine::Translator.

## Context

While adding unit tests in Zonemaster::Backend for its Translator module, I noticed that the `instance()` method in Zonemaster::Engine::Translator always returned `undef` instead of a Translator object.

## Changes

- Fix a bug in the `instance()` method which prevented it from working as documented.
- Fix small mistakes in the module’s documentation.
- Change unit test so that it calls `instance()` instead of `new()`, to exercise the code.

## How to test this PR

Run the test suite in Zonemaster::Engine. Expect all unit tests to pass.

### DIFF
--- a/lib/Zonemaster/Engine/Translator.pm
+++ b/lib/Zonemaster/Engine/Translator.pm
@@ -107,7 +107,7 @@ sub instance {
     my ( $class ) = @_;
 
     if ( !defined $instance ) {
-        $instance = $class->initialize();
+        $class->initialize();
     }
 
     return $instance;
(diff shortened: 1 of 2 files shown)


marc-vanderwal:
I’m moving one commit from this PR into its neighbor #1346, because it’s more relevant there than here. The idea is to make #1346 address the most urgent issues in Zonemaster::Engine::Translator and this one the more latent problems.

marc-vanderwal:
I’m keeping this PR as a draft for now, because I’m considering also deprecating the `new()` method altogether (as is stated in the documentation). A proper singleton class ought to have a *private* constructor and only expose the `instance()` function. If so, I’ll need to remember to change the `V-Patch` tag to `V-Major`.

marc-vanderwal:
> I’m keeping this PR as a draft for now, because I’m considering also deprecating the `new()` method altogether (as is stated in the documentation). A proper singleton class ought to have a _private_ constructor and only expose the `instance()` function. If so, I’ll need to remember to change the `V-Patch` tag to `V-Major`.

That’s better done in another PR.

PR #1352, opened 2024-05-31
sort generated data files

### LINKED ISSUE #1265: Unit tests: improve idempotence of generating data files
Currently, the way that data files are generated with `ZONEMASTER_RECORD=1 make test` or `Zonemaster::Engine::Nameserver::save()` has a few shortcomings that make it difficult to:
 * actually see what changed: `git diff` shows a huge mess of deletions and insertions that completely hide any meaningful changes;
 * sort out merge conflicts when several people made changes to said files.

This is due to two main reasons:

Firstly, the cache is implemented as a hash of hashes of hashes and `Zonemaster::Engine::Nameserver::save()` just iterates over the keys of the level 1 and 2 hashes without sorting them beforehand. This means that the cache entries are written out in random order, essentially rewriting the whole file. I propose that these hashes be walked through using a sorted list of keys (i.e. with `for my $k (sort keys %h)` instead of `for my $k (keys %h)`).

Secondly, packets are written as a binary representation of their wire formats, including the ID field. Because that field is randomized, two semantically identical packets can still differ only by their IDs. As a result, we are saving a different packet whenever we call `Zonemaster::Engine::Nameserver::save()`, even if nothing else changed. Besides, the packet’s transaction ID doesn’t matter anymore at the level of abstraction of `Zonemaster::Engine::Nameserver`. I therefore propose that the ID field of the packet’s header be set to zero when caching a response (similar to what is done in DNS over HTTPS to improve cache-friendliness; see [RFC 8484, § 4.1](https://www.rfc-editor.org/rfc/rfc8484.html#section-4.1)).

The two changes I propose are non-breaking and don’t even require regeneration of the `.data` files. Existing files can be sorted using Unix’s `sort` command.

Implementing both of these changes shouldn’t break anything at all, yet will ensure that running `ZONEMASTER_RECORD=1 make test` twice in a row will result in semantically identical `.data` files where changes are easier to spot with `git diff`.

MichaelTimbert:
## Purpose

This PR improve generating data files by sorting entries in *.data files and ensuring JSON dump are also sorted.
The goal is to reduce the mess of deletions and insertions from `git diff` when `t/*.data` files are modified.

## Context

Fixes #1265 

## How to test this PR

run ```ZONEMASTER_RECORD=1 make test``` to save new formated *.data file.

- "t/*.data" file must be sorted
- each "Zonemaster::LDNS::Packet" JSON dump must also be sorted

## Notes

To improve visibility of change we can also activate indentation of JSON dump with `$json->indent( 1 );`.
This makes it easier to read the differences between JSON dump with `vimdiff` or `git diff` but makes `*.data` files less readable/compact.

### DIFF
--- a/lib/Zonemaster/Engine/Nameserver.pm
+++ b/lib/Zonemaster/Engine/Nameserver.pm
@@ -523,6 +523,8 @@ sub save {
 
     my $old = POSIX::setlocale( POSIX::LC_ALL, 'C' );
     my $json = JSON::PP->new->allow_blessed->convert_blessed;
+    $json = $json->canonical(1);
+
     open my $fh, '>', $filename or die "Cache save failed: $!";
     foreach my $name ( sort keys %object_cache ) {
         foreach my $addr ( sort keys %{ $object_cache{$name} } ) {
--- a/lib/Zonemaster/Engine/Nameserver.pm
+++ b/lib/Zonemaster/Engine/Nameserver.pm
@@ -528,6 +528,13 @@ sub save {
     open my $fh, '>', $filename or die "Cache save failed: $!";
     foreach my $name ( sort keys %object_cache ) {
         foreach my $addr ( sort keys %{ $object_cache{$name} } ) {
+
+            # set ID field of packets to 0
+            my %data = %{$object_cache{$name}{$addr}->cache->data};
+            foreach my $key ( keys %data){
+                $data{$key}->packet->id(0) if defined($data{$key});
+            }
+
             say $fh "$name $addr " . $json->encode( $object_cache{$name}{$addr}->cache->data );
         }
     }
--- a/lib/Zonemaster/Engine/Nameserver.pm
+++ b/lib/Zonemaster/Engine/Nameserver.pm
@@ -524,8 +524,8 @@ sub save {
     my $old = POSIX::setlocale( POSIX::LC_ALL, 'C' );
     my $json = JSON::PP->new->allow_blessed->convert_blessed;
     open my $fh, '>', $filename or die "Cache save failed: $!";
-    foreach my $name ( keys %object_cache ) {
-        foreach my $addr ( keys %{ $object_cache{$name} } ) {
+    foreach my $name ( sort keys %object_cache ) {
+        foreach my $addr ( sort keys %{ $object_cache{$name} } ) {
             say $fh "$name $addr " . $json->encode( $object_cache{$name}{$addr}->cache->data );
         }
     }
--- a/lib/Zonemaster/Engine/Nameserver.pm
+++ b/lib/Zonemaster/Engine/Nameserver.pm
@@ -528,13 +528,6 @@ sub save {
     open my $fh, '>', $filename or die "Cache save failed: $!";
     foreach my $name ( sort keys %object_cache ) {
         foreach my $addr ( sort keys %{ $object_cache{$name} } ) {
-
-            # set ID field of packets to 0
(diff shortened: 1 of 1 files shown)


matsduf:
>     * each field "data" of ""Zonemaster::LDNS::Packet" must start with 'AA' to reflect the setting of ID field to 0.

What benefit does that give?

> (...) but makes `*.data` files less readable/compact.

What are the backsides of that?

MichaelTimbert:
> > (...) but makes `*.data` files less readable/compact.
> 
> What are the backsides of that?

`*.data` files will no longer resemble log files where there is only one entry per line.
The main drawback is that we have to rewrite the *.data loading code.
After reflection, I think this is not a good idea. If we need to compare different versions of *.data files we can always make a small script for that.

MichaelTimbert:
> > ```
> > * each field "data" of ""Zonemaster::LDNS::Packet" must start with 'AA' to reflect the setting of ID field to 0, QR bit is set to 1 and MSB of OPCODE is 0.
> > ```
> 
> 
> What benefit does that give?

To ensure the reproducibility of *.data files when doing a `ZONEMASTER_RECORD=1 make test`.
Otherwise, each time a new identifier will be used and the packet will be considered to have changed even if the content of the request is identical.

MichaelTimbert:
> > > > (...) but makes `*.data` files less readable/compact.
> > > 
> > > 
> > > What are the backsides of that?
> > 
> > 
> > `*.data` files will no longer resemble log files where there is only one entry per line. The main drawback is that we have to rewrite the *.data loading code. After reflection, I think this is not a good idea. If we need to compare different versions of *.data files we can always make a small script for that.
> 
> I cannot remember that I have ever compare data files. Unless someone has pointed out that he/she really wants it I think it is better not to change. I assume you will remove that change.

This is something i proposed, it is not implemented in this PR.

mattias-p:
With this approach we gain idempotence in the sense that the saved file is the same every time (provided that no code or configuration has changed). But we also lose idempotence in the sense that the original analysis sees the real IDs, but when we analyze a restored cache the analysis sees zeroed out IDs. In this light wouldn't it be more proper to clear the ID at response reception? I.e. before analysis and insertion into the cache.

I don't think we're ever looking at the ID fields in our built-in test modules, but since we want to support custom test modules, I believe it could make sense to have idempotence in both these ways.

matsduf:
> I don't think we're ever looking at the ID fields in our built-in test modules, but since we want to support custom test modules, I believe it could make sense to have idempotence in both these ways.

The thing about Message ID is that if the response does not have the same Message ID as the query it is not a response to that query and should be ignored or match with another query.

mattias-p:
> > I don't think we're ever looking at the ID fields in our built-in test modules, but since we want to support custom test modules, I believe it could make sense to have idempotence in both these ways.
> 
> The thing about Message ID is that if the response does not have the same Message ID as the query it is not a response to that query and should be ignored or match with another query.

Yes, it wouldn't make much sense to clear the response's Message ID before it's matched to the request's Message ID. The ldns library takes care of this matching for us and just returns the correct response matching our request.

What I'm suggesting is that we clear the ID field in Zonemaster Engine or perhaps in Zonemaster LDNS after ldns returns it but before we add it to the cache.

matsduf:
I suggest that we discuss this at the F2F.

matsduf (approved):

mattias-p (approved):

matsduf:
@MichaelTimbert, please merge.

PR #1343, opened 2024-05-07
Remove planned deprecated 'resolver.source' profile property

tgreenx:
## Purpose

This PR removes the planned deprecated `resolver.source` profile property. It also adds a proper IP validation module (and functions) in `Zonemaster::Engine::Validation`.

Other related changes to the `resolver.source{4,6}` profile properties and `Zonemaster::Engine::Nameserver source_address{4,6}` attributes are made. See the Changes section.

## Context

Fixes #1360 
Also fixes Perl warnings from #1342 

## Changes

This PR removes the `source_address{4,6}` and `source_address` attributes of `Zonemaster::Engine::Nameserver` objects.

Other related changes:
- Changes `resolver.source_address{4,6}` profile properties to not accept illegal values.
- Removes now unused constant `$RESOLVER_SOURCE_OS_DEFAULT`
- Removes now unused function `Zonemaster::Engine::Profile::check_validity()`
- ~~Rename `Zonemaster::Engine::Nameserver::source_address()` to `Zonemaster::Engine::Nameserver::_build_source_address()`~~
- Add IP validation module and functions `Zonemaster::Engine::Validation::validate_ipv{4,6}()`
- Refactoring
- Add and updates documentation
- Add and updates unit tests

## How to test this PR

Unit tests are updated and should pass.

PR #1363, opened 2024-06-17
Fix undef eq comparison

tgreenx:
## Purpose

This PR fixes an oversight from #1343.

## Changes

- Set default value of properties 'resolver.source{4,6}' in Profile.pm

## How to test this PR

Checks should now pass.


marc-vanderwal (dismissed):

mattias-p:
I disagree with the premise of this PR. [For the effective profile, all properties are always set](https://metacpan.org/pod/Zonemaster::Engine::Profile#effective), i.e. they have defined values. If $src_address is ever undef here then we have a bug in the profile module and we should fix it there instead.

marc-vanderwal (approved):

mattias-p:
I've verified that the default profile contains the default values for "resolver.source4" and "resolver.source6".

PR #1366, opened 2024-06-26
Spelling

jsoref:
## Purpose

Fixes misspellings identified by the [check-spelling action](https://github.com/marketplace/actions/check-spelling).

## Context

The misspellings have been reported at https://github.com/jsoref/zonemaster-engine/actions/runs/9671897619#summary-26683311397

## Changes

...

## How to test this PR

The action will report that the changes in this PR would make it happy: https://github.com/jsoref/zonemaster-engine/actions/runs/9671897732#summary-26683311682


matsduf (commented):
Changes must be done in the develop branch.

jsoref:
Do you want everything sent to develop, or should I just drop the commits that are problematic?

matsduf:
> Do you want everything sent to develop, or should I just drop the commits that are problematic?

The domain name change in the `t` file will make CI fail since then i will match the recorded data, and re-recording will fail since the names will not match the implemented zones.

`qr/^.* can not be undef/` is a matching expression and that will also fail if changed, as far as I can see.

I'll turn CI on.

jsoref:
ok, I think I've dropped the changes that are likely to cause problems for CI. We can revisit them later.

matsduf:
Please create the PR against the develop branch. If you click "edit" in the top you can change or I can do it.

lib/Zonemaster/Engine/Profile.pm
```diff
@@ -331,7 +331,7 @@ sub _set {
                             Data::Dumper::Dumper($value);
     # $value is a Scalar
     if ( ! $value_type  or $value_type eq q{SCALAR} ) {
-        die "Property $property_name can not be undef" if not defined $value;
+        die "Property $property_name cannot be undef" if not defined $value;
```
matsduf, line 334: What is stated here must match what `t/profiles.t` expects, or maybe the other way around.

Changing here without changing in `t/profiles.t` too makes the unit tests break. If you want to change here you must also do the update that you had before.

jsoref: Right. I'll drop it for now. It's too many things to reason over and it can come along in a later round.

matsduf (approved):

marc-vanderwal (approved):

PR #1379, opened 2024-08-08
$p masking in Zonemaster::Engine::Nameserver

emollier:
Greetings,

## Purpose

This change fixes the following warning caught by Debian Perl Team's autopkgtest starting with Zonemaster::Engine 5.0.0:

	"my" variable $p masks earlier declaration in same scope at /usr/share/perl5/Zonemaster/Engine/Nameserver.pm line 363.

## Context

This warning occurs because the variable is declared multiple times behind "if" statements, and another time unconditionally later in the subroutine.

## Changes

This patch proceeds to the $p variable declaration earlier in the subroutine unconditionnally, so they are always guaranteed to be available and unmasked.  The only drawback is that the $in_cache variable now needs to be declared in a separate statement.

## How to test this PR

Run the test suite, or run syntax checks, and ensure that the warning does not occur anymore in the logs.

Have a nice day,  :)
Étienne

### DIFF
--- a/lib/Zonemaster/Engine/Nameserver.pm
+++ b/lib/Zonemaster/Engine/Nameserver.pm
@@ -221,6 +221,7 @@ sub query {
         }
     );
 
+    my $p;
     my $class     = $href->{class}     // 'IN';
     my $dnssec    = $href->{dnssec}    // 0;
     my $usevc     = $href->{usevc}     // Zonemaster::Engine::Profile->effective->get( q{resolver.defaults.usevc} );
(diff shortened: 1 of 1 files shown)


emollier:
Thanks Thomas for your reviews, and the two other merges!

> Note that I have changed the base branch from `master` to
> `develop`, which is the branch that should be used for this
> project

I adjusted my local copies of Zonemaster::Engine and CLI
repositories to track the "develop" branch by default (and
dropped the "master" branch altogether as I'm tracking CPAN for
releases fit for publication).  This should avoid this issue in
the future.

Have a nice day,  :)
Étienne.

PR #1393, opened 2024-10-22
Update Connectivity04 implementation

tgreenx:
## Purpose

This PR proposes an update to the implementation of Connectivity04 following a [specification update](https://github.com/zonemaster/zonemaster/pull/1298). As a drive by change, this also updates the `t/TestUtil.pm` module to add support for dynamic scenario testing for test cases.

Note that this is the first implementation to make use of [MethodsV2](https://github.com/zonemaster/zonemaster-engine/blob/v6.0.0/lib/Zonemaster/Engine/TestMethodsV2.pm).

## Context

Test case specification update: https://github.com/zonemaster/zonemaster/pull/1298

Test zones specification: https://github.com/zonemaster/zonemaster/pull/1299

## Changes

`lib/Zonemaster/Engine/ASNLookup.pm`
`lib/Zonemaster/Engine/Test/Connectivity.pm`
`t/Test-connectivity04.t`
`t/Test-connectivity04.data`
`t/Test-connectivity04-A.t`
`t/Test-connectivity04-A.data`
`t/TestUtil.pm`

## How to test this PR

Unit tests are updated (based on https://github.com/zonemaster/zonemaster/pull/1299) and should pass.

### DIFF
--- a/t/Test-connectivity04.t
+++ b/t/Test-connectivity04.t
@@ -25,7 +25,7 @@ my @all_tags = qw(CN04_EMPTY_PREFIX_SET
                   CN04_IPV6_SAME_PREFIX
                   CN04_IPV6_SINGLE_PREFIX);
 
-# Specific hint file (test-zone-data/Connectivity-TP/hintfile.zone)
+# Specific hint file (https://github.com/zonemaster/zonemaster/blob/master/test-zone-data/Connectivity-TP/connectivity04/hintfile.zone)
 Zonemaster::Engine::Recursor->remove_fake_addresses( '.' );
 Zonemaster::Engine::Recursor->add_fake_addresses( '.',
     { 'root-ns1.xa' => [ '127.13.4.23', 'fda1:b2:c3::127:13:4:23' ],


t/Test-connectivity04.t
```diff
@@ -1,70 +1,239 @@
...
-zone_gives( $testcase, $zone, [qw{CN04_IPV4_SAME_PREFIX}] );
-zone_gives_not( $testcase, $zone, [qw{CN04_IPV4_DIFFERENT_PREFIX CN04_IPV6_DIFFERENT_PREFIX CN04_IPV6_SAME_PREFIX CN04_EMPTY_PREFIX_SET CN04_ERROR_PREFIX_DATABASE}] );
-
-TODO: {
-    my @missing = qw( CN04_EMPTY_PREFIX_SET CN04_ERROR_PREFIX_DATABASE CN04_IPV6_SAME_PREFIX );
-    local $TODO = "Need to find/create zones with those errors: ";
-    warn $TODO, "\n\t", join("\n\t", @missing), "\n";
-}
-###
+perform_testcase_testing( $test_case, $test_module, \@all_tags, %subtests );
```
matsduf, line 234: Please include support for `ZONEMASTER_SINGLE_SCENARIO`. See https://github.com/zonemaster/zonemaster-engine/blob/c2e37a09e54a6b0fa276078a398df10c2bd0eb3c/t/methodsv2.t#L807-L830

matsduf (commented):
When https://github.com/zonemaster/zonemaster/pull/1298 has been fixed I will update https://github.com/zonemaster/zonemaster/pull/1299 and test this implementation.

matsduf:
> Note that scenario ERROR-PREFIX-DATABASE-3 is currently disabled because since its query ends up blacklisted (no response), it causes subsequent queries to be blacklisted as well. I'll look for a workaround.

If that scenario is run by a separate `t` file it will not affect other scenarios. Isn't that the simple workaround? I do not think that the code should handle ASN lookup queries special in that sense.

matsduf:
Should there be double space between tag and argument?
```                                                                                                                                                                                                            
$ zonemaster-cli --hint=hintfile.zone --test=connectivity04 --level=info --show-testcase --raw GOOD-1.connectivity04.xa                                                                                        
   0.00 INFO     Unspecified    GLOBAL_VERSION  version=v6.0.0                                                                                                                                                 
   0.23 INFO     Connectivity04 CN04_IPV4_DIFFERENT_PREFIX  ns_list=dns0.connectivity04.xa/127.100.100.1;dns1.connectivity04.xa/127.100.101.1                                                                  
   0.23 INFO     Connectivity04 CN04_IPV6_DIFFERENT_PREFIX  ns_list=dns0.connectivity04.xa/fda1:b2:c3:0:127:100:100:1;dns1.connectivity04.xa/fda1:b2:c3:0:127:100:101:1                                        
```

matsduf:
I tested a domain that does not exist and got errors from Perl. It is not a big problem, but better if we can avoid it.
```
$ zonemaster-cli --hint=hintfile.zone --test=connectivity04 --level=info --show-testcase --raw ERROR-PREFIX-DATABASE-4.connectivity04.xa
   0.00 INFO     Unspecified    GLOBAL_VERSION  version=v6.0.0
   0.04 CRITICAL Unspecified    MODULE_ERROR  module=Connectivity; msg=Can't use an undefined value as an ARRAY reference at /usr/local/share/perl/5.34.0/Zonemaster/Engine/Test/Connectivity.pm line 807.
```

matsduf:
Tag without argument, but with two trailing space characters, `CN04_IPV4_SINGLE_PREFIX  `.

matsduf (changes requested):
Scenario name            | Mandatory message tag                                                             | Forbidden message tags
:------------------------|:----------------------------------------------------------------------------------|:--------------------
ERROR-PREFIX-DATABASE-6  | CN04_IPV4_DIFFERENT_PREFIX, CN04_IPV6_DIFFERENT_PREFIX, CN04_ERROR_PREFIX_DATABASE| 2)

```                                                                                                                                                                                                            
$ zonemaster-cli --hint=hintfile.zone --test=connectivity04 --level=info --show-testcase --raw ERROR-PREFIX-DATABASE-6.connectivity04.xa                                                                       
   0.00 INFO     Unspecified    GLOBAL_VERSION  version=v6.0.0                                                                                                                                                 
   0.15 INFO     Connectivity04 CN04_IPV4_DIFFERENT_PREFIX  ns_list=dns1.connectivity04.xa/127.100.101.1;dns25.connectivity04.xa/127.100.125.1                                                                 
   0.15 INFO     Connectivity04 CN04_IPV6_DIFFERENT_PREFIX  ns_list=dns1.connectivity04.xa/fda1:b2:c3:0:127:100:101:1;dns25.connectivity04.xa/fda1:b2:c3:0:127:100:125:1                                       
```                                                                                                                                                                                                            
--> Unexpected

matsduf:
> Unless I misunderstand, this is unrelated to this PR. This is also the case for other Test cases.

Not an issue for this PR.

> I don't see the issue. Could you provide an example? I get:

Just trailing spaces I observed. Nothing important.

> > --> Unexpected
> > Two TXT are returned and one of them does not have a prefix matching the queried IP address.
> > ```
> > ;; ANSWER SECTION:
> > 1.125.100.127.origin.asnlookup.zonemaster.net. 3600 IN TXT "64512 | 127.100.125.0/24 | NA | NA | NA"
> > 1.125.100.127.origin.asnlookup.zonemaster.net. 3600 IN TXT "64512 | 127.100.255.0/24 | NA | NA | NA"
> > ```
> 
> I believe this one is a bit tricky from a specification standpoint. We have both a valid and invalid record. Shouldn't we care more about returning the correct one? Moreover the network mask is the same in both cases, which makes it even more of an edge case. Finally, for this scenario this will lead to output of message tags CN04_IPV{4,6}_SINGLE_PREFIX. See https://github.com/zonemaster/zonemaster/blob/b38a3bd9acf37c55211117ff387c2220fec58fd0/docs/public/specifications/tests/Connectivity-TP/connectivity04.md?plain=1#L173-L187.

If more than one ASN TXT record is returned then I think it is reasonable to see that as a set. If one is more specific than the other, and both are compatible with the address, that is fine. If one record is not compatible with address, but the other is, then I think it is reasonable to say that the set is broken. That is what I think the specification says, and is meant to say:

>   4. If *Input IP* does not match the extracted subnet, output
>      *[CN04_ERROR_PREFIX_DATABASE]*, break the processing of TXT records and
>      exit this loop without returning any prefix.

matsduf:
> Moreover the network mask is the same in both cases, which makes it even more of an edge case.

I do not understand why the size of the mask matters. One prefix covers the address, the other does not.

t/Test-connectivity04.t
```diff
@@ -1,70 +1,234 @@
...
+        [],
+        [],
+    ],
+    #
+    # 'ERROR-PREFIX-DATABASE-3' defined in Test-connectivity04_2.t
+    #
+    'ERROR-PREFIX-DATABASE-6' => [
+        1,
+        q(error-prefix-database-6.connectivity04.xa),
+        [ qw(CN04_IPV4_DIFFERENT_PREFIX CN04_IPV6_DIFFERENT_PREFIX) ],
```
matsduf, line 113: This does not match the scenario specification:


Scenario name            | Mandatory message tag                                                             | Forbidden message tags
:------------------------|:----------------------------------------------------------------------------------|:--------------------
ERROR-PREFIX-DATABASE-6  | CN04_IPV4_DIFFERENT_PREFIX, CN04_IPV6_DIFFERENT_PREFIX, CN04_ERROR_PREFIX_DATABASE| 2)

matsduf: See https://github.com/zonemaster/zonemaster/pull/1299/files#r1822194134

matsduf (approved):

matsduf:
@marc-vanderwal, @MichaelTimbert, @mattias-p, please review.

t/Test-connectivity04.t
```diff
@@ -1,70 +1,234 @@
...
+                  CN04_IPV6_SAME_PREFIX
+                  CN04_IPV6_SINGLE_PREFIX);
 
-sub zone_gives {
-    my ( $test, $zone, $gives_ref ) = @_;
-    Zonemaster::Engine->logger->clear_history();
-    my @res = grep { $_->tag !~ /^TEST_CASE_(END|START)$/ } Zonemaster::Engine->test_method( $checking_module, $test, $zone );
-    foreach my $gives ( @{$gives_ref} ) {
-        ok( ( grep { $_->tag eq $gives } @res ), $zone->name->string . " gives $gives" );
+# Specific hint file (test-zone-data/Connectivity-TP/hintfile.zone)
```
matsduf, line 28: It is `test-zone-data/Connectivity-TP/connectivity04/hintfile.zone` but it should maybe be stated what repository it is located in since it is not in the same repository as this file, just as the link to the specification above (after release): `https://github.com/zonemaster/zonemaster/blob/master/test-zone-data/Connectivity-TP/connectivity04/hintfile.zone`

matsduf (approved):

marc-vanderwal (approved):
Looks good to me.

matsduf:
Release testing for 2024.2: No issues.

PR #1394, opened 2024-10-30
Updates global cache

matsduf:
## Purpose

This PR does three things about global (Redis) cache:
1. Updates how and what caching time an item gets (more below).
2. Changes the logic when TTL==0 by not saving to cache instead of setting TTL to 1 and chache that.
3. Removes the experimental status of global cache.

### Caching time and default for that

A DNS response can be of three type (when it comes to caching):
1. Normal response with data in answer section, NXDOMAIN or NODATA. In these cases a TTL can be  extracted from the DNS packet.
2. Another response such as REFUSED or SERVFAIL where no TTL can be extracted from the packet.
3. No response at all.

For response type 1 the extracted TTL value was used, but limited to cache.redis.expire (default 300 seconds, but much higher is needed to get full effect). For type 2 and 3 cache.redis.expire was used.

With this PR the logic for response type 1 is kept as it. For responses of type 2 and 3 a fixed value of 1200 seconds is used. The plan is to create a new profile element for type 2 and 3 to make that caching time settable.

## Context

Global cache was introduced in release v2023.2 and has shown no issues. With this PR together with https://github.com/zonemaster/zonemaster/pull/1303 it will become stable.

## How to test this PR

1. Review.
2. Install Redis and activate.
3. Update a profile with Redis and set expire not equal to 1200.
4. Run an undelegated test with `--ns` that points at something outside that does not respond on port 53.
5. Use `redis-cli KEYS '*'`, `redis-cli TTL KEY` and `redis-cli flushall`
6. `for key in $( redis-cli KEYS '*' | cut -f2 ) ; do redis-cli TTL $key ; done`

### DIFF
--- a/lib/Zonemaster/Engine/Nameserver/Cache/RedisCache.pm
+++ b/lib/Zonemaster/Engine/Nameserver/Cache/RedisCache.pm
@@ -91,8 +91,8 @@ sub set_key {
             }
         }
         $ttl = $ttl < $redis_expire ? $ttl : $redis_expire;
-        # Redis requires cache time to be greater than 0.
-        $ttl = 1 if $ttl == 0;
+        # Redis requires cache time to be greater than 0 to be stored.
+        return if $ttl == 0;
         $self->redis->set( $key, $msg, 'EX', $ttl );
     } else {
         $self->redis->set( $key, '', 'EX', $ttl );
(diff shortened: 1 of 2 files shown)


matsduf:
@tgreenx, @marc-vanderwal, @mattias-p and @MichaelTimbert: now the global chache return without saving to cache when TTL==0.

marc-vanderwal (approved):

matsduf:
@marc-vanderwal and @tgreenx, please review.

matsduf:
@marc-vanderwal and @tgreenx, please re-review.

marc-vanderwal (approved):

mattias-p (approved):

PR #1397, opened 2024-11-19
Fix default settings of queries

tgreenx:
## Purpose
This PR proposes a fix to the default settings of queries which, due to an oversight regarding the `edns_size` flag from #1147, made all non-DNSSEC queries to be EDNS queries instead of regular non-EDNS queries. Default EDNS0 packet size values will now be properly used when appropriate, and a new, missing one has been created for DNSSEC to align with [current specification](https://github.com/zonemaster/zonemaster/blob/master/docs/public/specifications/tests/DNSQueryAndResponseDefaults.md#default-setting-in-dns-query). The caching logic of queries was also impacted.

The basic usage for creating DNSSEC and EDNS queries remains unchanged.

## Context

Fixes https://github.com/zonemaster/zonemaster/issues/1308

## Changes

- Fix logic related to flags `dnssec`and `edns_size` for when to use default values, and also when combined with `edns_details`
- Fix caching logic when using `dnssec` and `edns_size` parameters
- Add upper and lower bound checks for `edns_size`
- Make combined usage of `dnssec` with `edns_details` but without `edns_details{do}` to correctly set the `dnssec` (DO) flag
- Removed uneeded code related to resetting flags between queries (- it was already done earlier in the same method)
- Add constant `UDP_DNSSEC_QUERY_DEFAULT` (set to 1232)
- Rename constant `UDP_COMMON_EDNS_LIMIT` to `UDP_EDNS_COMMON_LIMIT`
- Add and update documentation
- Add unit tests
- Update unit tests data

## How to test this PR

Units tests have been added or updated and should pass. Note that most of the newly added unit tests cannot be properly tested when replaying unit test data files (i.e without `ZONEMASTER_RECORD=1`), as the actual tested object is the `dns` attribute of a `Zonemaster::Engine::Nameserver` object and its properties are not changed when getting query responses directly from the cache.

Manual testing can be done using the commands below:

_Default DNS query_
```
$ perl -MZonemaster::Engine -E 'my $ns = Zonemaster::Engine->ns("d.nic.fr", "194.0.9.1"); my $p = $ns->query("fr", "A"); say "dnssec (DO) flag in query: ", $ns->dns->dnssec ? "true" : "false"; say "edns_size in query: ", $ns->dns->edns_size; say "DO flag in response: ", $p->do ? "true": "false"; say "has_edns in response: ", $p->has_edns ? "true": "false"';
dnssec (DO) flag in query: false
edns_size in query: 0
DO flag in response: false
has_edns in response: false
```

An optional hash can be passed as parameter to the `query` method, with keys such as `dnssec`, `edns_size` and `edns_details`. They should be changed and/or removed with various combinaisons to see the different behaviors.

_Default DNSSEC query_
```
$ perl -MZonemaster::Engine -E 'my $ns = Zonemaster::Engine->ns("d.nic.fr", "194.0.9.1"); my $p = $ns->query("fr", "A", { "dnssec" => 1 }); say "dnssec in query: ", $ns->dns->dnssec ? "true" : "false"; say "edns_size in query: ", $ns->dns->edns_size; say "DO flag in response: ", $p->do ? "true": "false"; say "has_edns in response: ", $p->has_edns ? "true": "false"';
dnssec in query: true
edns_size in query: 1232
DO flag in response: true
has_edns in response: true
```
_Default EDNS query_
```
$ perl -MZonemaster::Engine -E 'my $ns = Zonemaster::Engine->ns("d.nic.fr", "194.0.9.1"); my $p = $ns->query("fr", "A", { "edns_details" => {} }); say "dnssec in query: ", $ns->dns->dnssec ? "true" : "false"; say "edns_size in query: ", $ns->dns->edns_size; say "DO flag in response: ", $p->do ? "true": "false"; say "has_edns in response: ", $p->has_edns ? "true": "false"';
dnssec in query: false
edns_size in query: 512
DO flag in response: false
has_edns in response: true
```


lib/Zonemaster/Engine/Nameserver.pm
```diff
@@ -406,7 +406,13 @@ sub _query {
...
     $flags{q{recurse}}   = $href->{q{recurse}}   // Zonemaster::Engine::Profile->effective->get( q{resolver.defaults.recurse} );
     $flags{q{timeout}}   = $href->{q{timeout}}   // Zonemaster::Engine::Profile->effective->get( q{resolver.defaults.timeout} );
-    $flags{q{edns_size}} = $href->{q{edns_size}} // $UDP_EDNS_QUERY_DEFAULT;
+
+    if ( $flags{q{dnssec}} ) {
+        $flags{q{edns_size}} = $href->{q{edns_size}} // $UDP_EDNS_QUERY_DEFAULT;
+    }
+    else {
+        $flags{q{edns_size}} = $href->{q{edns_size}} // 0;
+    }
```
matsduf, line 415: The value must no be larger than 64 kB. If the value is smaller than 512 it must be treated as 512. Maybe we will create a test case where it is set smaller.

Is it possible by this code to create a non-DNSSEC EDNS query without specifying the size and get the default?

marc-vanderwal: I agree that it doesn’t make much sense to set a value larger than 65 535. I suggest to `croak` in this situation.

However, if the argument is 512 or lower, we have two options:

 - be a good citizen, not expose bugs in other people’s nameservers, log a warning on our side and force the value to be 512;
 - or, on the contrary, allow ourselves to send EDNS queries with silly buffer size values and see how other servers respond to that.

How should we proceed?

marc-vanderwal (commented):
Looking pretty good already. Just some minor issues with the updated POD.

matsduf:
The EDNS UDP buffer size is a 16-bit value. If the code tries to set something larger than that I think the code should "crash". This would only come from bad (Zonemaster) code.

If we ever want to set something lower than 512 we must permit that. We do not have such use case today, so I suggest that the code should "crash" so that we do not try to do the wrong thing.

lib/Zonemaster/Engine/Nameserver.pm
```diff
@@ -824,8 +824,8 @@ If set to true, prevents a server to be black-listed on a query in case there is
 Set the EDNS0 UDP maximum size. Defaults to 0, or 512 if the query is an non-DNSSEC EDNS query,
 or 1232 if the query is a DNSSEC EDNS query.
 
-Setting a value other than 0 will also implicitly enable the query to be an EDNS query.
-Value overridden by C<edns_details{size}> (if also given). More details in L<edns_details> below.
+Setting a value other than 0 will also implicitly enable EDNS for the query.
+Value overridden by C<edns_details->{size}> (if also given). More details in L<edns_details> below.
```
marc-vanderwal, line 828: I’ve made a mistake and I apologize. My previous suggestion should have been as follows, because the `>` needs to be escaped:
```suggestion
Value overridden by C<edns_details-E<gt>{size}> (if also given). More details in L<edns_details> below.
```

lib/Zonemaster/Engine/Nameserver.pm
```diff
@@ -834,10 +834,10 @@ A hash. An empty hash or a hash with any keys below will enable the query to be
 The currently supported keys are 'version', 'z', 'do', 'rcode', 'size' and 'data'.
 See L<Zonemaster::LDNS::Packet> for more details (key names prefixed with 'edns_').
 
-Note that flag L<edns_size> also exists (see above) and has the same effect as C<edns_details{size}>, although the value of the
+Note that flag L<edns_size> also exists (see above) and has the same effect as C<edns_details->{size}>, although the value of the
```
marc-vanderwal, line 837:
```suggestion
Note that flag L<edns_size> also exists (see above) and has the same effect as C<edns_details-E<gt>{size}>, although the value of the
```

lib/Zonemaster/Engine/Nameserver.pm
```diff
@@ -834,10 +834,10 @@ A hash. An empty hash or a hash with any keys below will enable the query to be
...
 latter will take precedence if both are given.
 
-Similarly, note that flag L<dnssec> also exists (see above) and has the same effect as C<edns_details{do}>, although the value of the
+Similarly, note that flag L<dnssec> also exists (see above) and has the same effect as C<edns_details->{do}>, although the value of the
```
marc-vanderwal, line 840:
```suggestion
Similarly, note that flag L<dnssec> also exists (see above) and has the same effect as C<edns_details-E<gt>{do}>, although the value of the
```

lib/Zonemaster/Engine/Nameserver.pm
```diff
@@ -826,22 +823,23 @@ If set to true, prevents a server to be black-listed on a query in case there is
 
 =item edns_size
 
-Set the EDNS0 UDP maximum size. Defaults to 512.
+Set the EDNS0 UDP maximum size. The value must be comprised between 0 and 65535.
+Defaults to 0, or 512 if the query is a non-DNSSEC EDNS query, or 1232 if the query is a DNSSEC query.
```
matsduf, line 827:
```suggestion
Set the EDNS0 UDP requestor's maximum size. The value must be 0 or comprised between 512 and 65535 (inclusive).
Defaults to 0 (non-EDNS query), or 512 (non-DNSSEC EDNS query), or 1232 (DNSSEC query).
```

lib/Zonemaster/Engine/Nameserver.pm
```diff
@@ -301,12 +299,13 @@ sub query {
...
         $md5->add( q{EDNS_EXTENDED_RCODE} , $href->{edns_details}{rcode} // 0 );
         $md5->add( q{EDNS_DATA}           , $href->{edns_details}{data} // q{} );
-        $md5->add( q{EDNS_UDP_SIZE}       , $href->{edns_details}{size} // $edns_size );
-    }
-    else {
-        $md5->add( q{EDNS_UDP_SIZE} , 0);
+        $edns_size = $href->{edns_details}{size} // ( $href->{edns_size} // ( $dnssec ? $UDP_DNSSEC_QUERY_DEFAULT : $UDP_EDNS_QUERY_DEFAULT ) );
     }
 
+    croak "edns_size (or edns_details->size) parameter must be a value between 0 and 65535" if $edns_size > 65535 or $edns_size < 0;
```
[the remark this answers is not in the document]
matsduf: Today we have no use cases for values 1-511. Any such value would today be a mistake and it would be better to catch those. For EDNS 0-511 is defined to mean the same thing as 512. If we ever want to check for servers handling small values we could always change the logic.

One could also argue that high values would never be used. We have today no use cases for high values. From that perspective it could be argued that the code should set a limit to catch mistakes.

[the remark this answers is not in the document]
matsduf: OK.

lib/Zonemaster/Engine/Nameserver.pm
```diff
@@ -301,12 +299,13 @@ sub query {
         $md5->add( q{EDNS_Z}              , $href->{edns_details}{z} // 0 );
         $md5->add( q{EDNS_EXTENDED_RCODE} , $href->{edns_details}{rcode} // 0 );
         $md5->add( q{EDNS_DATA}           , $href->{edns_details}{data} // q{} );
-        $md5->add( q{EDNS_UDP_SIZE}       , $href->{edns_details}{size} // $edns_size );
-    }
-    else {
-        $md5->add( q{EDNS_UDP_SIZE} , 0);
+        $edns_size = $href->{edns_details}{size} // ( $href->{edns_size} // ( $dnssec ? $UDP_DNSSEC_QUERY_DEFAULT : $UDP_EDNS_QUERY_DEFAULT ) );
```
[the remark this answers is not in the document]
matsduf: Ok.

lib/Zonemaster/Engine/Nameserver.pm
```diff
@@ -406,11 +405,13 @@ sub _query {
     $flags{q{fallback}}  = $href->{q{fallback}}  // Zonemaster::Engine::Profile->effective->get( q{resolver.defaults.fallback} );
     $flags{q{recurse}}   = $href->{q{recurse}}   // Zonemaster::Engine::Profile->effective->get( q{resolver.defaults.recurse} );
     $flags{q{timeout}}   = $href->{q{timeout}}   // Zonemaster::Engine::Profile->effective->get( q{resolver.defaults.timeout} );
-    $flags{q{edns_size}} = $href->{q{edns_size}} // $UDP_EDNS_QUERY_DEFAULT;
 
     if ( exists $href->{edns_details} ) {
         $flags{q{dnssec}}    = $href->{edns_details}{do} // $flags{q{dnssec}};
-        $flags{q{edns_size}} = $href->{edns_details}{size} // $flags{q{edns_size}};
+        $flags{q{edns_size}} = $href->{edns_details}{size} // ( $href->{q{edns_size}} // ( $flags{q{dnssec}} ? $UDP_DNSSEC_QUERY_DEFAULT : $UDP_EDNS_QUERY_DEFAULT ) );
```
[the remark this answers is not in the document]
matsduf: Ok.

matsduf (approved):

matsduf (dismissed):

matsduf (approved):

matsduf (approved):

PR #1422, opened 2025-01-28
Fix infinite recursion bug when NS record points to CNAME

### LINKED ISSUE #1421: NS record pointing to alias (CNAME) can cause infinite recursion
The following command line on a zone where one of its name server names points to an alias (CNAME) causes infinite looping:

```
$ zonemaster-cli --test basic02 --level info chi-eureseine.fr

Seconds Level    Message
======= ======== =======
   0.00 INFO     Using version v7.0.0 of the Zonemaster engine.
Deep recursion on subroutine "Zonemaster::Engine::Recursor::_recurse" at /home/vanderwal/.perl5/lib/perl5/Zonemaster/Engine/Recursor.pm line 442.
Deep recursion on subroutine "Zonemaster::Engine::Recursor::_do_query" at /home/vanderwal/.perl5/lib/perl5/Zonemaster/Engine/Recursor.pm line 297.
Deep recursion on subroutine "Zonemaster::Engine::Recursor::_do_query" at /home/vanderwal/.perl5/lib/perl5/Zonemaster/Engine/Recursor.pm line 297.
Deep recursion on subroutine "Zonemaster::Engine::Recursor::_do_query" at /home/vanderwal/.perl5/lib/perl5/Zonemaster/Engine/Recursor.pm line 297.
Deep recursion on subroutine "Zonemaster::Engine::Recursor::_do_query" at /home/vanderwal/.perl5/lib/perl5/Zonemaster/Engine/Recursor.pm line 297.
Deep recursion on subroutine "Zonemaster::Engine::Recursor::_do_query" at /home/vanderwal/.perl5/lib/perl5/Zonemaster/Engine/Recursor.pm line 297.
Deep recursion on subroutine "Zonemaster::Engine::Recursor::_do_query" at /home/vanderwal/.perl5/lib/perl5/Zonemaster/Engine/Recursor.pm line 297.
Deep recursion on subroutine "Zonemaster::Engine::Recursor::_do_query" at /home/vanderwal/.perl5/lib/perl5/Zonemaster/Engine/Recursor.pm line 297.
Deep recursion on subroutine "Zonemaster::Engine::Recursor::_do_query" at /home/vanderwal/.perl5/lib/perl5/Zonemaster/Engine/Recursor.pm line 297.
Deep recursion on subroutine "Zonemaster::Engine::Recursor::_do_query" at /home/vanderwal/.perl5/lib/perl5/Zonemaster/Engine/Recursor.pm line 297.
Deep recursion on subroutine "Zonemaster::Engine::Recursor::_do_query" at /home/vanderwal/.perl5/lib/perl5/Zonemaster/Engine/Recursor.pm line 297.
Deep recursion on subroutine "Zonemaster::Engine::Recursor::_do_query" at /home/vanderwal/.perl5/lib/perl5/Zonemaster/Engine/Recursor.pm line 297.
Deep recursion on subroutine "Zonemaster::Engine::Recursor::_do_query" at /home/vanderwal/.perl5/lib/perl5/Zonemaster/Engine/Recursor.pm line 297.
Deep recursion on subroutine "Zonemaster::Engine::Recursor::_do_query" at /home/vanderwal/.perl5/lib/perl5/Zonemaster/Engine/Recursor.pm line 297.
Deep recursion on subroutine "Zonemaster::Engine::Recursor::_do_query" at /home/vanderwal/.perl5/lib/perl5/Zonemaster/Engine/Recursor.pm line 297.
Deep recursion on subroutine "Zonemaster::Engine::Recursor::_do_query" at /home/vanderwal/.perl5/lib/perl5/Zonemaster/Engine/Recursor.pm line 297.
Deep recursion on subroutine "Zonemaster::Engine::Recursor::_do_query" at /home/vanderwal/.perl5/lib/perl5/Zonemaster/Engine/Recursor.pm line 297.
Deep recursion on subroutine "Zonemaster::Engine::Recursor::_do_query" at /home/vanderwal/.perl5/lib/perl5/Zonemaster/Engine/Recursor.pm line 297.
Deep recursion on subroutine "Zonemaster::Engine::Recursor::get_addresses_for" at /home/vanderwal/.perl5/lib/perl5/Zonemaster/Engine/Recursor.pm line 385.
Deep recursion on subroutine "Zonemaster::Engine::Recursor::_resolve_cname" at /home/vanderwal/.perl5/lib/perl5/Zonemaster/Engine/Recursor.pm line 317.
Deep recursion on subroutine "Zonemaster::Engine::Recursor::_resolve_cname" at /home/vanderwal/.perl5/lib/perl5/Zonemaster/Engine/Recursor.pm line 317.
```

## Explanation

Why does it happen on `chi-eureseine.fr`? And why does it happen with Basic02?

In the process of executing the test case, a SOA query for `chi-eureseine.fr` is done. Zonemaster eventually gets a referral to `ns.crihan.fr` and `ns.ch-eureseine.fr`, with glue only for `ns.crihan.fr`; both are out-of-bailiwick (yes, the second one as well; there is no typo).

Zonemaster then attempts to get A records for `ns.ch-eureseine.fr`. The `ch-eureseine.fr` zone is delegated to `ns.crihan.fr` (with glue) and `ns.chi-eureseine.fr`.

It eventually gets a response from `ns.crihan.fr`: `ns.ch-eureseine.fr` is an alias for `lxnsext.ch-eureseine.fr`. There is an A record in the response, so Zonemaster stops there.

But then, Zonemaster attempts to get AAAA records for `ns.ch-eureseine.fr`. This time, it gets the same CNAME resource record from `ns.crihan.fr` without any AAAA records in the same response, because there are none.

Because of that, Zonemaster restarts a recursive query for `lxnsext.ch-eureseine.fr/AAAA`. In the process, it forgets that it was actually trying to resolve AAAA records for `ns.ch-eureseine.fr` to get the SOA record for `chi-eureseine.fr`. Zonemaster sees that it needs to resolve AAAA records for `ns.ch-eureseine.fr` in the process of resolving `lxnsext.ch-eureseine.fr/AAAA`, gets the lone CNAME again, and retries the recursive query for `lxnsext.ch-eureseine.fr/AAAA` once again, forgetting all context again. This goes on infinitely, until `zonemaster-engine` is killed.

marc-vanderwal:
## Purpose

This PR fixes an infinite recursion bug that could be triggered when, somewhere in the delegation chain, an NS record points to an alias.

### Problem description

In a nutshell: suppose that `a.example` is a zone whose authoritative nameserver is `ns.a.example`. But `ns.a.example` is an alias for `real-ns.a.example`. When the recursor tries to resolve `ns.a.example` to IP addresses, it chases the alias: this causes it to restart a query for `real-ns.a.example`. In the process, it needs to resolve `ns.a.example` to an IP address so it can query that name server for `real-ns.a.example`. However, the recursor threw away the context that led it to chase `real-ns.a.example` in the first place, that is: resolving `ns.a.example`.

Because of that, Zonemaster tries to resolve `ns.a.example`, which is aliased to `real-ns.a.example`, but to resolve `real-ns.a.example`, it needs to resolve `ns.a.example`, which is aliased to `real-ns.a.example`, and to resolve that name, it needs to resolve `ns.a.example`, which is aliased to `real-ns.a.example`, and so on. It’s an infinite loop that isn’t technically a CNAME chain nor an NS chain, but some hybrid of both.

### Solution

All we need to do is to keep the original query’s context around when performing a subordinate query caused by an alias (CNAME record). That way, when chasing `real-ns.a.example`, the recursor notices that it needs to resolve `ns.a.example`, which is a resolution that it is already attempting to perform, and therefore bails out instead of looping indefinitely.

## Context

Fixes #1421.

## Changes

* When initializing the state structure for a subordinate query caused by chasing an alias, copy the `in_progress` table from the original query’s state to the subordinate query’s.
* The first change prevents `CNAME_LOOP_OUTER` messages to be generated when detecting a CNAME loop across zones. We need to change the condition for generating those messages so that it uses the `in_progress` table.

## How to test this PR

Unit tests for MethodsV2 have been updated and should complete without error.

Another option is to run `zonemaster-cli --test basic02 chi-eureseine.fr`. The program should complete with `Looks OK.`, instead of printing dozens of `Deep recursion` warnings.

### DIFF
--- a/lib/Zonemaster/Engine/Recursor.pm
+++ b/lib/Zonemaster/Engine/Recursor.pm
@@ -245,7 +245,7 @@ sub _resolve_cname {
     }
 
     # CNAME target has already been followed (outer loop); no need to recurse
-    if ( $state->{tseen}{lc( $target )}  ) {
+    if ( exists $state->{in_progress}{lc( $target )}{$type} ) {
         Zonemaster::Engine->logger->add( CNAME_LOOP_OUTER => { name => $name, target => $target, targets_seen => join( ';', keys %{ $state->{tseen} } ) } );
         return ( undef, $state );
     }
@@ -263,7 +263,7 @@ sub _resolve_cname {
     unless ( $name->is_in_bailiwick( $target ) ) {
         Zonemaster::Engine->logger->add( CNAME_FOLLOWED_OUT_OF_ZONE => { name => $name, target => $target } );
         ( $p, $state ) = $class->_recurse( $target, $type, $dns_class,
-            { ns => [ root_servers() ], count => 0, common => 0, seen => {}, tseen => $state->{tseen}, tcount => $state->{tcount}, glue => {} });
+            { ns => [ root_servers() ], count => 0, common => 0, seen => {}, tseen => $state->{tseen}, tcount => $state->{tcount}, glue => {}, in_progress => $state->{in_progress} });
     }
     else {
         # What do do here?
--- a/t/methodsv2.t
+++ b/t/methodsv2.t
@@ -757,6 +757,18 @@ my %subtests = (
               ns4-cname.child-ns-cname-3.methodsv2.xa/fda1:b2:c3:0:127:40:1:34 ) ],
         [ ],
     ],
+    'CHILD-NS-CNAME-4' => [
+        1,
+        q(child.parent.child-ns-cname-4.methodsv2.xa),
+        [ qw( 127.40.1.41
+              fda1:b2:c3:0:127:40:1:41
+              127.40.1.42
+              fda1:b2:c3:0:127:40:1:42 ) ],
+        [ qw( ns1-cname.child.parent.child-ns-cname-4.methodsv2.xa/127.40.1.51 ) ],
+        [ qw( ns1-cname.child.parent.child-ns-cname-4.methodsv2.xa/127.40.1.51
+              ns2-cname.child.parent.child-ns-cname-4.methodsv2.xa/127.40.1.52 ) ],
+        [ ],
+    ],
     'PARENT-NS-CNAME-1' => [
         1,
         q(child.parent.parent-ns-cname-1.methodsv2.xa),


matsduf:
For MethodsV2 there are five test scenarios where the name server name is an alias via CNAME:
* CHILD-NS-CNAME-1
* CHILD-NS-CNAME-2
* CHILD-NS-CNAME-3
* PARENT-NS-CNAME-1
* PARENT-NS-CNAME-2

We have not seen any issues with those. @marc-vanderwal, can you see what is missing to trigger the bug fixed in this PR?

Test scenario specification: https://github.com/zonemaster/zonemaster/blob/master/docs/public/specifications/test-zones/MethodsV2/methodsv2.md

matsduf (approved):

marc-vanderwal:
Thanks for looking into that!

The scenario that triggers the bug is:
 * the zone being tested must have at least one in-bailiwick name server name pointing to an in-zone alias, similar to CHILD-NS-CNAME-1;
 * however, the target of the alias must have either A records only, or AAAA records only. This is contrary to CHILD-NS-CNAME-1, where the alias targets have both A and AAAA records.

I’ll see if I can make a new scenario for that and reproduce the bug. For the time being, given the severity of the bug that this PR fixes, I propose that this PR is merged as is; I’ll create an issue for the missing unit test and see if I can work on that.

marc-vanderwal:
I managed to reproduce the bug in a new scenario that I named `CHILD-NS-CNAME-4`. And I can verify that the problem disappears when I apply the fix in this PR. Shall I merge it?

marc-vanderwal:
@matsduf I’ve added unit tests for this fix based on the scenario that I’d like to add (see zonemaster/zonemaster#1348). Can you review again, please?

matsduf (approved):

marc-vanderwal:
Okay, here we go then!

PR #1427, opened 2025-02-04
Check for object type in returned values from some TestMethodsV2 methods used in Test Cases

### LINKED ISSUE #1426: Crash in Test Cases using TestMethodsV2 for zones with non-existing name servers
When NS records (either in the delegation or in the zone itself) of a domain name point to at least one non-existing name server name (i.e. can't be resolved to an IP address), DNSSEC10 and Connectivity04 will crash due to an oversight in the Test Cases which makes use of TestMethodsV2.

Consider the following:
```
$ dig +nocookie +nocmd NS robindesboispaysage.fr @d.nic.fr
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 58079
;; flags: qr rd; QUERY: 1, ANSWER: 0, AUTHORITY: 2, ADDITIONAL: 1
;; WARNING: recursion requested but not available

;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags:; udp: 1232
;; QUESTION SECTION:
;robindesboispaysage.fr.                IN      NS

;; AUTHORITY SECTION:
robindesboispaysage.fr. 3600    IN      NS      ns0.wixdns.ne.
robindesboispaysage.fr. 3600    IN      NS      ns1.wixdns.net.

;; Query time: 19 msec
;; SERVER: 194.0.9.1#53(d.nic.fr) (UDP)
;; WHEN: Tue Feb 04 18:09:15 CET 2025
;; MSG SIZE  rcvd: 106

$ dig +nocookie +nocmd NS robindesboispaysage.fr @ns1.wixdns.net
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 34691
;; flags: qr aa rd; QUERY: 1, ANSWER: 2, AUTHORITY: 0, ADDITIONAL: 1
;; WARNING: recursion requested but not available

;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags:; udp: 512
;; QUESTION SECTION:
;robindesboispaysage.fr.                IN      NS

;; ANSWER SECTION:
robindesboispaysage.fr. 86400   IN      NS      ns0.wixdns.net.
robindesboispaysage.fr. 86400   IN      NS      ns1.wixdns.net.

;; Query time: 19 msec
;; SERVER: 216.239.34.101#53(ns1.wixdns.net) (UDP)
;; WHEN: Tue Feb 04 18:09:43 CET 2025
;; MSG SIZE  rcvd: 97
```

As there is a typo in the delegation NS records (`ns0.wixdns.ne.` instead of `ns0.wixdns.net.`) to an unresolvable name, this currently causes Zonemaster to crash because some methods in `TestMethodsV2.pm` can return a mix of different types of objects (i.e. `Zonemaster::Engine::Nameserver` when a name can be resolved, or `Zonemaster::Engine::DNSName` otherwise) :
```
$ zonemaster-cli --level=info --no-ipv6 --show-testcase --test dnssec10 --test connectivity04 robindesboispaysage.fr

Seconds Level    Testcase       Message
======= ======== ============== =======
   0.00 INFO     Unspecified    Using version v7.0.0 of the Zonemaster engine.
   9.15 CRITICAL Unspecified    Fatal error in DNSSEC: Can't locate object method "address" via package "Zonemaster::Engine::DNSName" at /home/tgreen/projects/zonemaster/zonemaster-engine/lib/Zonemaster/Engine/Test/DNSSEC.pm line 3158.

   0.00 INFO     Unspecified    Using version v7.0.0 of the Zonemaster engine.
   0.17 CRITICAL Unspecified    Fatal error in Connectivity: Can't locate object method "address" via package "Zonemaster::Engine::DNSName" at /home/tgreen/projects/zonemaster/zonemaster-engine/lib/Zonemaster/Engine/Test/Connectivity.pm line 814.
```

tgreenx:
## Purpose

Methods `Zonemaster::Engine::TestMethodsV2::get_{zone,del}_ns_names_and_ips()` can return a mix of `Zonemaster::Engine::Nameserver` or `Zonemaster::Engine::DNSName` objects, depending on whether the name could be resolved into an IP address. Due to a previous oversight, this can cause issues in Test Cases that use them. We can filter `Zonemaster::Engine::DNSName` objects out since they can't be queried.

Moreover it also fixes a buggy ternary expression in the same place (see https://github.com/zonemaster/zonemaster-engine/pull/1427#discussion_r1942398968 below). 

## Context

Fixes #1426 

## Changes

- Filter non-`Zonemaster::Engine::Nameserver` objects out of the list of to-be-queried name servers in Test Cases
- Rework buggy ternary expression to a working alternative

## How to test this PR

Unit tests could be created but they are kind of out of scope for either the TestMethodsV2 or the Test Cases themselves.
Manual testing (from #1426):
```
$ zonemaster-cli --level=info --no-ipv6 --show-testcase --test dnssec10 --test connectivity04 robindesboispaysage.fr

Seconds Level    Testcase       Message
======= ======== ============== =======
   0.00 INFO     Unspecified    Using version v7.0.0 of the Zonemaster engine.
  18.30 NOTICE   DNSSEC10       The zone is not DNSSEC signed or not properly DNSSEC signed. Testing for NSEC and NSEC3 has been skipped. Fetched from name servers "ns0.wixdns.net/216.239.32.101;ns1.wixdns.net/216.239.34.101".
   0.00 INFO     Unspecified    Using version v7.0.0 of the Zonemaster engine.
   2.01 INFO     Connectivity04 The following name server(s) are announced in unique IPv4 prefix(es): "ns0.wixdns.net/216.239.32.101;ns1.wixdns.net/216.239.34.101"
```

### DIFF
--- a/lib/Zonemaster/Engine/Test/DNSSEC.pm
+++ b/lib/Zonemaster/Engine/Test/DNSSEC.pm
@@ -3146,11 +3146,8 @@ sub dnssec10 {
     my ( @with_dnskey, @without_dnskey );
 
     my @nss = grep { $_->isa('Zonemaster::Engine::Nameserver') } (
-                Zonemaster::Engine::TestMethodsV2->get_del_ns_names_and_ips( $zone ) ?
-                    Zonemaster::Engine::TestMethodsV2->get_zone_ns_names_and_ips( $zone ) ?
-                        @{ Zonemaster::Engine::TestMethodsV2->get_zone_ns_names_and_ips( $zone ), Zonemaster::Engine::TestMethodsV2->get_del_ns_names_and_ips( $zone ) }
-                    : @{ Zonemaster::Engine::TestMethodsV2->get_del_ns_names_and_ips( $zone ) }
-                : ()
+                @{ Zonemaster::Engine::TestMethodsV2->get_del_ns_names_and_ips( $zone ) // [] },
+                @{ Zonemaster::Engine::TestMethodsV2->get_zone_ns_names_and_ips( $zone ) // [] }
               );
     my @ignored_nss;
 
--- a/lib/Zonemaster/Engine/Test/Connectivity.pm
+++ b/lib/Zonemaster/Engine/Test/Connectivity.pm
@@ -805,11 +805,8 @@ sub connectivity04 {
     my %ip_already_processed;
 
     my @nss = grep { $_->isa('Zonemaster::Engine::Nameserver') } (
-                Zonemaster::Engine::TestMethodsV2->get_del_ns_names_and_ips( $zone ) ?
-                    Zonemaster::Engine::TestMethodsV2->get_zone_ns_names_and_ips( $zone ) ?
-                        @{ Zonemaster::Engine::TestMethodsV2->get_zone_ns_names_and_ips( $zone ), Zonemaster::Engine::TestMethodsV2->get_del_ns_names_and_ips( $zone ) }
-                    : @{ Zonemaster::Engine::TestMethodsV2->get_del_ns_names_and_ips( $zone ) }
-                : ()
+                @{ Zonemaster::Engine::TestMethodsV2->get_del_ns_names_and_ips( $zone ) // [] },
+                @{ Zonemaster::Engine::TestMethodsV2->get_zone_ns_names_and_ips( $zone ) // [] }
               );
 
     foreach my $ns ( @nss ) {


lib/Zonemaster/Engine/Test/DNSSEC.pm
```diff
@@ -3145,16 +3145,18 @@ sub dnssec10 {
...
+                        @{ Zonemaster::Engine::TestMethodsV2->get_zone_ns_names_and_ips( $zone ), Zonemaster::Engine::TestMethodsV2->get_del_ns_names_and_ips( $zone ) }
+                    : @{ Zonemaster::Engine::TestMethodsV2->get_del_ns_names_and_ips( $zone ) }
+                : ()
+              );
 
     my %ip_already_processed;
     my $testing_time = time;
 
-    for my $ns ( sort @nss ) {
+    for my $ns ( @nss ) {
```
matsduf, line 3159: Why not sorted? That should give consistent order in repeated tests.

marc-vanderwal:
Actually, the output of `zonemaster-cli` in the “How to test this PR” section is wrong: there are two name servers and both are announced in different prefixes, so the WARNING from Connectivity04 is a false warning. And the NOTICE in DNSSEC10 only lists ns1.wixdns.net at the end of the message; ns0.wixdns.net is missing.

Applying my review suggestions gives a more correct output from zonemaster-cli:

```
Seconds Level    Message
======= ======== =======
   0.00 INFO     Using version v7.0.0 of the Zonemaster engine.
   2.00 INFO     The following name server(s) are announced in unique IPv4 prefix(es): "ns0.wixdns.net/216.239.32.101;ns1.wixdns.net/216.239.34.101"
   0.00 INFO     Using version v7.0.0 of the Zonemaster engine.
   0.16 NOTICE   The zone is not DNSSEC signed or not properly DNSSEC signed. Testing for NSEC and NSEC3 has been skipped. Fetched from name servers "ns0.wixdns.net/216.239.32.101;ns1.wixdns.net/216.239.34.101".
```

marc-vanderwal (approved):

mattias-p (approved):

matsduf:
Most test cases only use name servers by IP address. If a name server only has a name and no address it will be disregarded. The name is only needed to make the messages more readable, a nice feature for the users.

The PR adds the same code to two test cases to filter out name servers without address. I feel it is wrong approach. We would then need a scenario for each test case to make sure this filtering is done correctly. And the test case code get more complex.

It would be a better solution if MethodsV2 methods provide the name servers twice, one set of IP addresses, and one set of name/IP pairs. In that way we lower the risk of errors and make the test case code simpler. Maybe even a method that provide the combination of delegation and zone, which is what we use in most test cases (no separation is needed).

If an update of the MethodsV2 specification is needed, we can take that time. I am willing to work on that.

mattias-p:
I agree that the MethodV2 return types aren't the most ergonomic. AFAICT this is partly a consequence of how MethodsV2 are specified. I thought about how the situation could be improved and updated zonemaster/zonemaster#1345 with a design proposal that would allow nicer programmatic interfaces. One of the ideas is that methods should return lists of homogenous types that can be deduplicated using List::Util::uniq instead of arrayrefs of sometimes heterogeneous types.

mattias-p (approved):

matsduf (approved):

PR #1465, opened 2025-07-28
Fix typo in unit test of BASIC02

marc-vanderwal:
## Purpose

This PR fixes a typo in the `t/test-Basic02.t` file that caused one of the scenarios, GOOD-2, not to be tested at all.

The scenario GOOD-2 erroneously pointed to the test zone named `good-1.basic02.xa`, where it should have pointed to `good-2.basic02.xa`.

## Context

Issue found during refactoring.

## Changes

- Fix typo in `t/test-Basic02.t`;
- Rerecord companion `.data` file.

## How to test this PR

Unit tests should pass.


matsduf (approved):

matsduf:
Such an error is hard to find!

matsduf:
> Adding a check in `t/TestUtil.pm` would be great to catch those mistakes, as in most cases the zone name is derived from the test scenario name and test case name, but unfortunately that is not always the case

The only scenarios when it is not is when the zone is the root zone, a TLD or an ARPA zone. The check could ignore those, and then most cases would be caught.

PR #1467, opened 2025-07-30
Introduce a domain-specific language for scenario-based unit testing of Zonemaster-Engine

marc-vanderwal:
## Purpose

In order to support the ongoing effort of testing Zonemaster’s test cases with a scenario-based approach, this PR introduces a domain-specific language (DSL) aimed at easing the writing and maintenance of the corresponding `.t` files.

The design goals of the DSL are as follows:
* Be easy to read and understand for both experienced developers and occasional contributors;
* Be concise, by eliminating as much boilerplate and internal repetition as possible;
* Provide opportunities for the `.t` files to be self-documenting;
* Allow for extensibility should the need arise;
* Stay in the same spirit as `Test::More`;
* Provide “escape hatches” by letting unit test writers use custom callbacks when built-in facilities are insufficient;
* Provide good and useful diagnostics when a test fails.

I’ve converted a handful of scenario-based `.t` files in the new DSL in order to test some edge cases. However, some features of the DSL are currently untested:
* Testing for message arguments (no scenario-based `.t` file currently does that);
* Fake DS records (no scenario-based `.t` file needs undelegated DS records).

Because this PR introduces no change visible to the end user, I haven’t set any of the versioning tags (V-Major, V-Minor or V-Patch).

### Main differences with TestUtil.pm

The DSL strives to stay compatible with TestUtil.pm in terms of features. However, the DSL proved to be a good opportunity to improve on two areas.

Firstly, the DSL allows to test not only for the presence or absence of a message, but also the presence of a message satisfying certain constraints on its arguments. Diagnostic output in case such a test fails is hopefully detailed enough for pinpointing the cause of the failure.

Secondly, while TestUtil.pm only has a boolean property stating whether or not a scenario is testable, the DSL makes a distinction between two cases that I believe should be handled differently:
 * scenarios marked `not_testable`, that cannot be tested because the infrastructure (e.g. authoritative name servers configured to exhibit erroneous behavior or zones badly configured on purpose) is not yet available;
 * scenarios marked `todo`, that cannot be tested out because of a bug in Zonemaster-Engine that is yet to be fixed.

The difference mainly lies in what happens when the test suite is run, and what happens if `ZONEMASTER_RECORD=1` is set in the environment. Scenarios marked `not_testable` are entirely skipped (as when [todo_skip in Test::More](https://perldoc.pl/Test::More#todo_skip) is used) and therefore are not recorded in the `.data` file. Thus, the data file is not polluted with packets that do not correspond to the scenario’s specification. Scenarios marked `todo` are treated as if they were in a [TODO block in Test::More](https://perldoc.pl/Test::More#TODO:-BLOCK): they are executed, but failures are ignored, while the corresponding packets are saved in the data file, so that someone wishing to fix the implementation can test their fix without having to rerecord the `.data` file.

### Diagnostics

When testing for the presence of a message of a certain tag and with certain arguments, it can be tricky to narrow down what exactly happened. The DSL compiles to unit tests that strive to provide helpful diagnostics in this case, as in this sample output (with a test that was made to fail intentionally):
```
t/Test-basic02.t .. 13/?
    #   Failed test 'Messages of tag 'B02_AUTH_RESPONSE_SOA' exist with specified arguments'
    #   at t/Test-basic02.t line 112.
    # Looked for a message whose tag is 'B02_AUTH_RESPONSE_SOA'
    #   where 'domain' equals 'mixed-1.basic02.xa'
    #     and 'ns_list' equals 'ns1.mixed-1.basic02.xa/127.12.2.31;ns1.mixed-1.basic02.xa/fda1:b2:c3:0:127:12:2:31'
    #     and contains no argument other than those listed above
    # Here are all messages that unsuccessfully matched:
    #   B02_AUTH_RESPONSE_SOA domain="mixed-1.basic02.xa"; ns_list=ns1.mixed-1.basic02.xa/127.12.2.31;ns1.mixed-1.basic02.xa/fda1:b2:c3:0:127:12:2:31
#   Failed scenario 'MIXED-1'
#   at t/Test-basic02.t line 117.
# Looks like you failed 1 test of 26.
t/Test-basic02.t .. Dubious, test returned 1 (wstat 256, 0x100)
Failed 1/26 subtests

Test Summary Report
-------------------
t/Test-basic02.t (Wstat: 256 Tests: 26 Failed: 1)
  Failed test:  16
  Non-zero exit status: 1
Files=1, Tests=26,  1 wallclock secs ( 0.03 usr  0.00 sys +  1.06 cusr  0.08 csys =  1.17 CPU)
Result: FAIL
```

## Context

Hackathon session during last face-to-face meeting, and zonemaster/zonemaster#1401.

## Changes

 * Introduce a DSL
 * Port a subset of scenario-based `*.t` files (i.e. using `TestUtil::perform_testcase_testing()`) to the DSL

## Advice for reviewers

This PR contains new code (for the DSL implementation, compiler and documentation) and rewrites a few `.t` files. For the `.t` files, it is better to ignore the diff and look at the end result only. The original and the DSL rewrite of the `.t` files should be semantically equivalent.

If so far the DSL looks fine, is it worth the effort to convert the rest of the scenario-based `.t` files in this format? In the current state, this PR has converted about half of the scenario-based `.t` files so far in order to try and explore edge cases.

## How to test this PR

Unit tests should pass.

If wishing to test this PR manually, you should run the test suite with a tool such as `prove` and enable the `-v` (verbose) option in order to obtain the full diagnostic output. Otherwise, try modifying a test so that it fails and look at the diagnostics output: it should point to the line in the `.t` file where the error occurred.


matsduf:
> Because this PR introduces no change visible to the end user, I haven’t set any of the versioning tags (V-Major, V-Minor or V-Patch).

It should still be a `V-Patch` as all changes requires increase in version number.

t/TestUtil/DSL.pm
```diff
@@ -0,0 +1,657 @@
...
+    check_state 'testing_test_case';
+
+    croak "At least two arguments required" unless scalar @_ >= 2;
+
+    my $definition = pop @_;
+    my @names = _expand_scenario_names(@_);
+
+    my @context = caller(0);
+
+    push @{STATE->{scenarios}}, {
```
marc-vanderwal, line 228: Instead of accumulating `scenario` AST elements into `@{STATE->{scenarios}}`, and other keywords into other keys of that hash, it is better to accumulate them all into a single array that mixes them all. Otherwise, users might get nasty surprises.

t/TestUtil/DSL.pm
```diff
@@ -0,0 +1,657 @@
...
+earlier, the entire set of IP addresses is replaced with the latter
+occurrence’s arguments.
+
+=cut
+
+sub root_hint ($@) {
+    my ($ns_name, @addresses) = @_;
+    check_state 'testing_test_case';
+    STATE->{root_hints}{$ns_name} = \@addresses;
+}
```
marc-vanderwal, line 311: It might be better to name the keyword `root_hints` and to make it take a hashref mapping names to arrayrefs of IP addresses. The reason being that I might want to change root hints somewhere in the middle of a unit test.

marc-vanderwal:
I’ve reworded the documentation for the `expect` keyword and provided some additional examples.

Issue #1148, opened 2022-11-18
Remove items from profile


matsduf:
In profile.json there are some resolver defaults that should be controlled by the test case specifications and it does not make sense to have configurable:

resolver.defaults.dnssec
resolver.defaults.edns_size
resolver.defaults.igntc
resolver.defaults.fallback
resolver.defaults.recurse
resolver.defaults.usevc

Remove those.

marc-vanderwal:
It seems to me that this issue is now resolved. Can we close it?

matsduf:
`resolver.defaults.fallback` still remains. To be deprecated and then removed.

PR #460, opened 2018-12-19
Edns extras

vlevigneron:
- First implementations of extra test cases for EDNS (DNS Flag Day). Nameserver10, Nameserver11, Nameserver12, Nameserver13 and Nameserver14
- See https://github.com/zonemaster/zonemaster/pull/650 forNameserver10
- See https://github.com/zonemaster/zonemaster/pull/661 for Nameserver11
- See https://github.com/zonemaster/zonemaster/pull/658 for Nameserver12
- See https://github.com/zonemaster/zonemaster/pull/653 for Nameserver13
- See https://github.com/zonemaster/zonemaster/pull/654 for Nameserver14

PR #1472, opened 2025-10-09
Ignore and deprecate some profile properties

mattias-p:
## Purpose

Ignore and deprecate profile properties that complicate query parameter semantics for no good reason, and allows user configuration to cause Zonemaster to misbehave for no good reason.

## Context

Fixes the remainder of #1148.

## Changes

Ignore and deprecate these profile properties:
 * `resolver.defaults.igntc`
 * `resolver.defaults.recurse`
 * `resolver.defaults.usevc`

## How to test this PR

Here is a profile file that sets all four properties to unreasonable values:
```json
{"resolver":{"defaults":{"igntc":true,"recurse":true,"usevc":true}}}
```
I don't have any examples of zones that break with that profile, but it does break ASN lookups without the changes in this PR.

### DIFF
--- a/lib/Zonemaster/Engine/Nameserver.pm
+++ b/lib/Zonemaster/Engine/Nameserver.pm
@@ -171,9 +171,6 @@ sub _build_dns {
 
     $res->retry( Zonemaster::Engine::Profile->effective->get( q{resolver.defaults.retry} ) );
     $res->retrans( Zonemaster::Engine::Profile->effective->get( q{resolver.defaults.retrans} ) );
-    $res->usevc( Zonemaster::Engine::Profile->effective->get( q{resolver.defaults.usevc} ) );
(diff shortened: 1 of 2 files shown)


mattias-p:
Here is some context for why the default "fallback" value is true: https://github.com/zonemaster/zonemaster-engine/pull/460

matsduf:
> Here is some context for why the default "fallback" value is true: #460

I do not see it in #460. Can you please point it out?

Is that due to some strange behavior in LDNS that makes `resolver.defaults.igntc` not do what it should? That should be enough.

matsduf:
@mattias-p, please set correct `V-XXX` and `RC-XXX` labels on it.

tgreenx:
> @mattias-p, please set correct `V-XXX` and `RC-XXX` labels on it.

@mattias-p @matsduf I've set RC-Deprecations and V-Patch (maybe V-Minor instead? I'm not sure).

PR #1475, opened 2025-10-29
Implement new MethodV2: Get-Parent-NS-Names-and-IPs

marc-vanderwal:
## Purpose

This PR provides an implementation of Get-Parent-NS-Names-and-IPs.

## Context

See zonemaster/zonemaster#1418.

## Changes

* Implement Get-Parent-NS-Names-and-IPs;
* Update Get-Parent-NS-IPs;
* Update MethodsV2 unit tests to test Get-Parent-NS-Names-and-IPs instead of Get-Parent-NS-IPs and add the new test scenarios introduced in zonemaster/zonemaster#1418;
* Rerecord some of the `t/*.data` files where needed.

## How to test this PR

Unit tests should pass.


lib/Zonemaster/Engine/TestMethodsV2.pm
```diff
@@ -1,6 +1,6 @@
 package Zonemaster::Engine::TestMethodsV2;
 
-use v5.16.0;
+use v5.24.0;
```
matsduf: We set minimum Perl version v5.26 in v2025.1. That is set in Zonemaster-LDNS. Is there any reason to set the minimum Perl version in this file? If so, is there any reason to have it lower than the general minimum Perl version?

matsduf (changes requested):

marc-vanderwal:
> Can the new unit test file format introduced by #1467 be used in this case too?

Alas, no, because MethodsV2 tests do not work quite the same way as test case tests do. I’d need to implement a slightly different DSL if we wanted a format similar to #1467. It would be nice to have that, but it didn’t feel urgent to me because we are much less likely to touch MethodsV2 unit tests than test case tests.

matsduf (dismissed):
Looks fine. Besides reviewing I rely on the scenarios/unit tests having passed.

lib/Zonemaster/Engine/TestMethodsV2.pm
```diff
@@ -253,17 +242,53 @@ sub get_parent_ns_ips {
...
             join "\034", ( $class, $zone->name );
         });
 
 =over
 
+=item get_parent_ns_ips($zone)
+
+[External]
+
+This Method obtains the name servers that serves the parent zone, i.e. the zone from which the Child Zone is delegated from. If more than one name server share the same IP address, only one among them is kept.
```
tgreenx:
```suggestion
This Method obtains the name servers that serve the parent zone, i.e. the zone from which the Child Zone is delegated from. If more than one name server share the same IP address, only one among them is kept.
```

lib/Zonemaster/Engine/TestMethodsV2.pm
```diff
@@ -253,17 +242,53 @@ sub get_parent_ns_ips {
...
+=back
+
+=cut
+
+sub get_parent_ns_ips {
+    my ( $class, $zone ) = @_;
+
+    # FIXME: We really should just be outputting name server IPs here, as the
+    # specification says. Instead we output name server objects, but filtered
+    # to have unique IP addresses.
```
tgreenx:
```suggestion
    # FIXME: We really should just be outputting name server IPs here, as the
    # specification says. Instead we output name server objects (but filtered
    # on unique IP addresses) because these objects are required to
    # perform queries.
```

lib/Zonemaster/Engine/TestMethodsV2.pm
```diff
@@ -253,17 +242,53 @@ sub get_parent_ns_ips {
...
+
+    return undef if not defined $nameservers;
+
+    my %ns_by_ip = ();
+    foreach my $ns ( @$nameservers ) {
+        my $ip = $ns->address->short;
+        $ns_by_ip{$ip} = $ns unless exists $ns_by_ip{$ip};
+    }
+
+    return [ values %ns_by_ip ];
```
tgreenx: For consistency
```suggestion
    return [ sort values %ns_by_ip ];
```

lib/Zonemaster/Engine/TestMethodsV2.pm
```diff
@@ -253,17 +242,54 @@ sub get_parent_ns_ips {
...
             join "\034", ( $class, $zone->name );
         });
 
 =over
 
+=item get_parent_ns_ips($zone)
+
+[External]
+
+This Method obtains the name servers that serve the parent zone, i.e. the zone from which the Child Zone is delegated from. If more than one name server share the same IP address, only one among them is kept.
```
matsduf: Double "from".

```suggestion
This Method obtains the name servers that serve the parent zone, i.e. the zone from which the Child Zone is delegated. If more than one name server share the same IP address, only one among them is kept.
```

matsduf (changes requested):

matsduf:
Just two small comments on the language. Else this looks good. Since the unit tests pass I see no issues.

matsduf (approved):

tgreenx (approved):

PR #1506, opened 2026-02-24
Add artifact build to the CI

MichaelTimbert:
## Purpose

This PR adds artifact creation to CI. This artifact can be useful for release testing and PR review.
The artifact is created only if all tests pass.

## Context

Follow-up to https://github.com/zonemaster/zonemaster-backend/issues/1229

## How to test this PR

You should be able to download the artifact from the action tab and install it by following the Zonemaster documentation.
https://doc.zonemaster.fr/latest/installation/README.html

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -108,7 +108,7 @@ jobs:
         run: sudo apt-get install -y build-essential git libmodule-install-perl
 
       - name: build
-        run: perl Makefile.PL && make dist
+        run: perl Makefile.PL && make all dist
 
       - name: Get short SHA
         id: short_sha
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -105,7 +105,7 @@ jobs:
       - uses: actions/checkout@v4
 
       - name: apt install
-        run: sudo apt-get install -y build-essential git libmodule-install-perl
+        run: sudo apt-get install -y build-essential git libmodule-install-perl gettext
 
       - name: build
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -97,3 +97,38 @@ jobs:
...
+                echo "SHORT_SHA=$(echo ${{ github.sha }} | cut -c1-7)" >> $GITHUB_ENV
+            fi
+
+      - name: Get Zonemaster-Engine version
+        id: version
+        run: |
+          result=`grep "use version; our $VERSION" lib/Zonemaster/Engine.pm`
+          result+='printf $VERSION;'
+          VERSION=`perl -e "$result"`
+          echo "version=$VERSION" >> $GITHUB_OUTPUT
```
marc-vanderwal, line 128: There is a _much_ cleaner way.

From the source tree’s root, the following one-liner:
```
perl -Ilib -MZonemaster::Engine -E 'say Zonemaster::Engine->VERSION'
```
will give the following output:
```
v8.1.1
```
(This will load the `Zonemaster::Engine` module living in the `lib` subdirectory of the source tree, so this is guaranteed to take the version of the right file.)

MichaelTimbert: I agree that is much better, but in the build-artefact environnement Zonemaster::LDNS module is not installed, so loading Zonemaster::Engine will failed.

matsduf:
@MichaelTimbert, what milstone should it be on this PR?

matsduf (approved):

PR #1510, opened 2026-03-03
Preparation for release v2025.2.1 (Zonemaster-Engine)

matsduf:
## Purpose

This PR updates `Changes`, `lib/Zonemaster/Engine.pm` and `Makefile.PL`for the release.

Note: CI will fail this PR until https://github.com/zonemaster/zonemaster-ldns/pull/247 has been merged.

## How to test this PR

Review.

### DIFF
--- a/Changes
+++ b/Changes
@@ -3,49 +3,49 @@ Release history for Zonemaster component Zonemaster-Engine
 v8.1.1 2026-03-04 (part of Zonemaster v2025.2.1 release)
 
  [Fixes]
-- Corrects Nameserver13 implementation (#1505)
-- Fixes inactive upward referral guard (#1494)
-- Updates root hints after B-root renumbering (#1490)
-- Fixes missing value in Zone11 message tag (#1491)
+ - Corrects Nameserver13 implementation (#1505)
+ - Fixes inactive upward referral guard (#1494)
+ - Updates root hints after B-root renumbering (#1490)
+ - Fixes missing value in Zone11 message tag (#1491)
 
 
 v8.1.0 2025-12-17 (part of Zonemaster v2025.2 release)
 
  [Deprecation]
-- Ignores and deprecates some profile properties. To be removed in
-  release v2026.1 (#1472)
+ - Ignores and deprecates some profile properties. To be removed in
+   release v2026.1 (#1472)
 
  [Features]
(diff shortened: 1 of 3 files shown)


Changes
```diff
@@ -1,43 +1,52 @@
 Release history for Zonemaster component Zonemaster-Engine
 
+v8.1.1 2026-03-04 (part of Zonemaster v2025.2.1 release)
+
+ [Fixes]
+ - Corrects Nameserver13 implementation (#1505)
+ - Fixes inactive upward referral guard (#1494)
+ - Updates root hints after B-root renumbering (#1490)
```
mattias-p, line 8: This is tagged with RC-None. Is the tag wrong?

Changes
```diff
@@ -1,43 +1,52 @@
...
+ - Fixes missing value in Zone11 message tag (#1491)
+ - Updates Swedish translation (#1509)
```
mattias-p, line 10: This is not merged yet.

mattias-p (commented):
I have two comments. Otherwise this looks fine.

Oh, and CI breaks. I guess it's only because of the bumped patch position of the Zonemaster::LDNS dependency, but the tests exit early on failure.

matsduf:
> Oh, and CI breaks. I guess it's only because of the bumped patch position of the Zonemaster::LDNS dependency, but the tests exit early on failure.

But that is because it cannot load Zonemaster-LDNS version 5.0.2, isn't it?

```
! Installing the dependencies failed: Installed version (5.0.1) of Zonemaster::LDNS is not in range '5.000002'
! Bailing out the installation for Zonemaster-Engine-8.001001.
39 distributions installed
Error: Process completed with exit code 1.
```

Changes
matsduf: I missed that, but if you run `zonmaster-cli` verbosely (DEBUG) then you can see it, so it is user visiable, isn't it?

Changes
matsduf: On the meeting we said it could be if reviewed/approved today. @tolvmannen has promised to do that. Or else I have to removed it.

Changes
matsduf: I changed the tag. I think it should be included. A user reported the issue.

mattias-p (approved):
I'm approving this so that it can be merged without further synchronization once the translation has been merged.

tgreenx (approved):

PR #1515, opened 2026-03-25
Remove some unmaintainable and redundant tests in t/recursor.t

marc-vanderwal:
## Purpose

This PR removes parts of a unit test that depend on live data from the Internet and are therefore unmaintainable. The same code is exercised by `t/recursor-A.t`, so there is no regression in test coverage in doing so. More details and a further discussion of the rationale can be found in the commit message.

## Context

Development of unrelated feature.

## Changes

* Remove some redundant unit tests.

## How to test this PR

Unit tests should still pass. Running the tests within Devel::Cover should show no decrease in coverage.

Issue #938, opened 2021-06-29
New cache format


mattias-p:
After Engine has performed a run, its cache can be saved to disk. This effectively creates an archive of those pieces of the internet that are needed to perform this particular run.
The cache file can be restored from disk and the same run can be performed again with the same result, provided that it is the 
version of Engine that restores the cache is the same version that stored it.

However there are some limitations to the current implementation.
* The cache file does not include the parameters of the run(s) that were used to collect the cache data.
* The cache file does not include the version number of Engine that was used to create it. This is probably only useful occasionally when debugging.
* The cache file does not include ASN lookup requests. Consequently a test that uses ASN lookup makes new ANS queries no matter if you're performing a fresh run or a run from restored cache data.
* The file format isn't flexible enough to support MethodsNT. E.g. it does not distinguish between NS responses given from parent zone name servers and responses given from the zones own name servers.
* The file format is tightly coupled with Engine's the internal cache representation. This is bad for two reasons. First, it's wasteful because there are lots of repetitions of the strings "Zonemaster::Engine::Packet" and "Zonemaster::LDNS::Packet". Second, if we want to change the internal cache representation, we'll either have to change the cache file format or we'll have to emulate the old cache format just to be able to save and restore the cache files.

A new cache file format should be designed and implemented, and all existing data files should be converted to the new format. We should consider supporting both formats during a transition period.

marc-vanderwal:
We might need to keep the old cache format around, or at least migrate the data files’ contents to a different structure inside the new format.

mattias-p:
We (@marc-vanderwal, @MichaelTimbert and I) noted that we need to store multiple different sections in the new cache format and discussed which one to choose. In the end we settled on CBOR which seems to be a good fit for our use case. It's a binary format similar to JSON, but with good performance.

The Perl implementation that people are using seems to be [CBOR::XS](https://metacpan.org/pod/CBOR::XS) and the Rust implementation that people are using seems to be [ciborium](https://crates.io/crates/ciborium).

We should make a PoC to ensure we can get reproducible serializations of Map data with CBOR.

marc-vanderwal:
For writing CBOR files, it might be useful to consider prepending the CBOR-encoded data with the “magic value” described in [RFC 8949, § 3.4.6](https://www.rfc-editor.org/rfc/rfc8949.html#section-3.4.6), in order to distinguish them from the old format.

matsduf:
[...]

PR #1516, opened 2026-03-26
Update unit tests depending on real-world data

marc-vanderwal:
## Purpose

This PR updates some unit tests so that the associated data files can be rerecorded again.

Some unit tests rely on real-world data across the Internet. There is a facility for storing and replaying network packets so that running the test suite does not generate any requests across the network; this enables running the test suite without depending on Internet access and avoids breakage whenever a network requests gives a different result than what was previously the case.

However, there may be circumstances where rerecording the entirety of the test suite’s data files is necessary. In some cases, this is not possible anymore because of changes that happen beyond our control. This PR tries to increase the likelihood that `ZONEMASTER_RECORD=1 make test` passes without errors.

This PR also addresses a logic error in `t/nameserver.t` that fails to take into account that Zonemaster::Engine only generates some messages when the networks was accessed, instead of returning a cached result. For more information, see the correspoding commit’s message.

This PR supersedes #1515.

## Context

Development of #938.

## Changes

 * Updates to the following unit tests:
   * `t/asn.t`
   * `t/nameserver.t`
   * `t/recursor.t`
   * `t/Test-connectivity03.t`
   * `t/Test-delegation.t`
   * `t/Test-zone01-A.t`
   * `t/Test-zone01-B.t`
   * `t/undelegated.t`

## How to test this PR

Run `make test` and expect unit tests to pass.

Then run `ZONEMASTER_RECORD=1 make test` and expect unit tests to pass.

Finally, run `make test` again and expect unit tests to pass one more time.


matsduf (dismissed):

marc-vanderwal:
I found a flaw in `t/nameserver.t` which, I think, fits well into the scope of this PR as well. @matsduf, can you review again, please?

tgreenx (approved):

tgreenx (approved):

tgreenx (approved):

matsduf (approved):

PR #1517, opened 2026-04-02
Overhaul saved packets’ serialization format

### LINKED ISSUE #938: New cache format

marc-vanderwal:
## Purpose

This PR addresses some long-standing issues with the packet cache in Zonemaster-Engine, its scope, and its serialization format.

The current serialization format in use had some design flaws. It was fairly slow to deserialize due to its use of a combination of plain text lines and JSON. Files were big because of numerous instances of `Zonemaster::Engine::Packet` and `Zonemaster::LDNS::Packet` strings appearing in the JSON data, and the base64 encoding of binary data. The cache’s keys are the result of a one-way computation (in this case, MD5) over query attributes, which makes inspection of files, and migration of formats unnecessarily difficult, not to mention the risk of collisions.

In addition to that, not all DNS queries were cached; AXFR queries, used for example in [Nameserver03](https://doc.zonemaster.net/latest/specifications/tests/Nameserver-TP/nameserver03.html), did not go through it.

CBOR ([RFC 8949](https://www.rfc-editor.org/rfc/rfc8949.html)) proves to be a satisfactory serialization format. It is concise, fast to deserialize, and accepts binary data without base64 encoding.

Cache keys are now the actual DNS packets sent over the wire (with the ID field repurposed for internal flags), making the design simpler and easing migration to other formats if needed.

As for AXFR requests and responses, these are cached as synthetic (_i.e._ fake) DNS packets. This implementation leverages Extended DNS Errors (EDEs) in the synthetic packets to record AXFR failures accurately.

I managed to migrate most `t/*.data` files from the old to the new format, but this required a temporary kludge to log the cache keys being accessed, both in the old (hashed) and the new (wireformat) schemes. After running the unit tests once, it was possible to recompute the cache keys in the new scheme, load the data from the old format, and save it in the new format. Three of those data files required manual rerecording, but these were among the easiest data files to regenerate. The kludges in question are all removed in a follow-up commit.

The only remaining issue left unaddressed in #938 is the caching of ASN queries when done through the RIPE method (_i.e._ Whois on port 43/tcp). Currently, these are still not recorded in the cache. When querying ASN numbers through the Cymru method however, the response is cached because this method uses DNS. Addressing this specific issue requires a more major overhaul of Zonemaster-Engine’s caching system, but this PR lays the groundwork for fixing it.

On my development setup, comparing the combined sizes of all `t/*.data` files and the time spent running `make test` (across 10 runs, using hyperfine), the figures are as follows:

|  | Test suite run time | Data file size |
|---|---|---|
| Before | 92.353 s ±  0.496 s | 73 MB |
| After | 59.138 s ± 0.851 s | 48 MB |

Which is a 36% improvement in test suite run time and 34% improvement in data file size.

## Context

Fixes #938, with a small caveat.

## Changes

 * Refactor of `Zonemaster::Engine::Nameserver` class;
 * Change cache serialization format from JSON to CBOR;
 * Use actual query packet in wireformat as cache key;
 * Also cache AXFR queries and responses.

Each change is discussed in more thorough detail in the corresponding commit message.

## How to test this PR

Unit tests should still pass.


marc-vanderwal:
Unit tests fail because zonemaster/zonemaster-ldns#254, which is a hard dependency for this PR, is not merged yet.

marc-vanderwal:
Also, I’m not quite sure if this change is to be regarded as breaking. Do we consider the cache as being an interface or an implementation detail of Zonemaster-Engine?

matsduf:
> Also, I’m not quite sure if this change is to be regarded as breaking. Do we consider the cache as being an interface or an implementation detail of Zonemaster-Engine?

The format is exportable through `zonemaster-cli`, so from that perspective it is a breaking change. If that special case is disregarded, then it is a implementation detail. I think it is better to be safe than sorry and consider it to be a breaking change,

marc-vanderwal:
> The format is exportable through `zonemaster-cli`, so from that perspective it is a breaking change. If that special case is disregarded, then it is a implementation detail. I think it is better to be safe than sorry and consider it to be a breaking change.

Yes, you are right. I’ve set the appropriate tags.

lib/Zonemaster/Engine/Nameserver.pm
```diff
@@ -440,31 +522,8 @@ sub _query {
...
-                $pkt->edns_data($href->{edns_details}{data});
-            }
-
-            $res = eval { $self->dns->query_with_pkt( $pkt ) };
-        }
-        else {
-            $res = eval { $self->dns->query( "$name", $type, $href->{class} ) };
-        }
+        my $pkt = $self->_make_query_packet( $name, $type, $href );
+        $res = eval { $self->dns->query_with_pkt( $pkt ) };
```
tgreenx, line 534: > One consequence of this split-up is that _query() now always sends
> packets with Zonemaster::LDNS->query_with_pkt(). Previously, _query()
> called Zonemaster::LDNS->query() for non-EDNS queries and
> query_with_pkt() for EDNS queries.

I was wondering if that change is entirely benign. After looking at Zonemaster-LDNS and LDNS code, it appears to be fine. Also the default settings of LDNS objects ([`resolver`](https://www.nlnetlabs.nl/documentation/ldns/resolver_8c_source.html#l00608) and [`packet`](https://www.nlnetlabs.nl/documentation/ldns/packet_8c_source.html#l00863)) match the ones specified in the ["DNS Query And Response Defaults" specification](https://github.com/zonemaster/zonemaster/blob/master/docs/public/specifications/tests/DNSQueryAndResponseDefaults.md).

lib/Zonemaster/Engine/Nameserver.pm
tgreenx, line 1: A few new internal methods have been added in this module, but none of them are documented (e.g. `_make_query_packet`, `_query_cache_key`, etc...).

tgreenx (commented):
LGTM, nice work! Aside from one observation, I have just one comment regarding documentation.

Once https://github.com/zonemaster/zonemaster-ldns/pull/254 is merged and conflicts are resolved, I'll approve.

lib/Zonemaster/Engine/Nameserver.pm
```diff
@@ -374,6 +356,106 @@ sub add_fake_ds {
...
+        }
+    }
+
+    $pkt->do($dnssec);
+    $pkt->edns_size($edns_size);
+
+    return $pkt;
+}
+
+sub _query_cache_key {
```
mattias-p: Maybe rename this to `_make_cache_key`?

lib/Zonemaster/Engine/Nameserver.pm
```diff
@@ -374,6 +356,106 @@ sub add_fake_ds {
     return;
 } ## end sub add_fake_ds
 
+sub _make_query_packet {
+    my ( $self, $name, $type, $href ) = @_;
+
+    $type //= 'A';
+    $href->{class} //= 'IN';
```
mattias-p: Maybe use a local variable instead of mutating the argument?

util/data2dig
```diff
@@ -1,53 +1,78 @@
...
-    'Zonemaster::Engine::Packet' => sub {
-        my ( $ref ) = @_;
-        return Zonemaster::Engine::Packet->new( { packet => $ref } );
+# Decode input into packets
+my $cbor = CBOR::XS->new();
+my $saved_contents = $cbor->decode(do { local $/; <> });
+
+my $format_version = $saved_contents->{format_version};
+say ";;;;";
+say ";;;; BEGIN OF METADATA";
```
mattias-p: "Begin of" is funny grammar.
```suggestion
say ";;;; START OF METADATA";
```

util/data2dig
```diff
@@ -64,46 +89,7 @@ for my $packet ( @packets ) {
 =head1 DESCRIPTION
 
 B<data2dig> exports saved Zonemaster::Engine cache files to human readable
-format as chronologically ordered response packets in dig format.
+format as packets in dig format.
 
+=end
```
mattias-p: Maybe?
```suggestion
=cut
```

lib/Zonemaster/Engine/Nameserver.pm
marc-vanderwal: Thanks for double-checking!

lib/Zonemaster/Engine/Nameserver.pm
```diff
@@ -647,10 +760,37 @@ sub axfr {
...
+    if ( $in_cache ) {
+        if ( $p->rcode() ne 'NOERROR' ) {
+            # Croak with the same error message the real AXFR croaked with.
+            my ( undef, $ede_text ) = $p->packet->ede();
+            croak $ede_text // "AXFR transfer error: REFUSED";
+        }
+
+        my $last_ret;
+        foreach my $rr ( $p->answer() ) {
+            return 0 unless ($last_ret = $callback->($rr)) != 0;
```
mattias-p: There's a lot going on on this line. Maybe refactor it?
```suggestion
            $last_ret = $callback->($rr);
            return 0 if $last_ret == 0;
```

lib/Zonemaster/Engine/Nameserver.pm
```diff
@@ -647,10 +760,37 @@ sub axfr {
...
+        }
+        return $last_ret;
+    }
+    else {
+        my ( $ret, $error, $p ) = $self->_axfr( $domain, $callback, $class );
+        $self->cache->set_key( $idx, $p ) if defined $p;
```
mattias-p: Shouldn't we record non-responses too?

mattias-p (dismissed):
Looks good! Found a few nits. Nothing critical.

lib/Zonemaster/Engine/Nameserver.pm
marc-vanderwal: Oh, I see how the function name could be ambiguous. I meant “key for query cache” as a noun, not “query the cache key” as a verb. I’ll rename to `_key_for_query_cache`.

util/data2dig
marc-vanderwal: Indeed. Also, `perldoc` complained with an error. Changing to `=cut` gets rid of the error message.

lib/Zonemaster/Engine/Nameserver.pm
marc-vanderwal: Yep, noticed that too during rebasing. I’ll change that.

lib/Zonemaster/Engine/Nameserver.pm
marc-vanderwal: Yes, we should.

lib/Zonemaster/Engine/Nameserver.pm
marc-vanderwal: Good point. Using `last` to end the loop early would be even cleaner.

lib/Zonemaster/Engine/Nameserver.pm
marc-vanderwal: None of the underscored methods have any kind of documentation, but I can add a brief comment to those that I introduced.

marc-vanderwal:
In Zonemaster-LDNS, zonemaster/zonemaster-ldns#254 is now merged. Retriggering CI.

tgreenx (approved):

tgreenx:
@matsduf @mattias-p please re-review.

matsduf (approved):

MichaelTimbert:
work on rocky 8

PR #1518, opened 2026-04-07
Update CNAME resolution

### LINKED ISSUE #1505: Address02: Errors at PTR records with CNAME
Hello,
here is Lukas Vacek from CZ.NIC. We are running own public zonemaster instance and we see that someone has a issue with PTR records.

Tests for domain `korpus.cz` ends with error at sections System (Address02 - "System:Address02:CNAME_RECORDS_CHAIN_BROKEN cname_counter=0; cname_rrs=1; name=66.53.113.195.in-addr.arpa") and Connectivity (Address02 - "No response from nameserver(s) on PTR query (66.53.113.195.in-addr.arpa.)."). 

I am not able to figure out why these messages. I went through the DNS reverse tree and nothing suspicious looks to me. The special thing for that domain is that the reverse records has CNAME in path. But the domain records based on [RFC 2181 section 10.2](https://datatracker.ietf.org/doc/html/rfc2181#section-10.2) can be a CNAME. 

Do you have idea why it is happening? Do you find something wrong in the domain tree configuration?

Thank you.

Kind regards,
Lukas Vacek

tgreenx:
## Purpose

This PR slightly updates the CNAME resolution of Engine's recursor. See the `Changes` section below.
Updated test scenarios are defined in https://github.com/zonemaster/zonemaster/pull/1477.

## Context

https://github.com/zonemaster/zonemaster/pull/1477

Also fixes https://github.com/zonemaster/zonemaster/issues/1505

## Changes

- `Zonemaster::Engine::Recursor::_resolve_cname()` now returns undef when the CNAME resolution fails, along with new message tags: `CNAME_UNRESOLVABLE` (ERROR level) and `CNAME_TO_NODATA` (DEBUG level)
- Fix case-sensitiveness of CNAME records
- Update unit tests:
   - Move the content of scenario `TOO-LONG-CNAME-CHAIN` to a new scenario `TOO-MANY-CNAME`,
   - Update scenario `TOO-LONG-CNAME-CHAIN` to check for tag `CNAME_CHAIN_TOO_LONG` instead
   - Add scenario `UNRESOLVABLE-CNAME` 
   - Add scenario `CNAME-CHAIN-TO-NODATA`

## How to test this PR

Unit tests are updated and should pass.


lib/Zonemaster/Engine/Recursor.pm
```diff
@@ -250,7 +250,7 @@ sub _resolve_cname {
         return ( undef, $state );
     }
 
-    # Safe-guard against anormaly long consecutive CNAME chains; no need to recurse
+    # Safe-guard against anormaly long consecutive CNAME lookups; no need to recurse
```
marc-vanderwal:
```suggestion
    # Safeguard against abnormally long consecutive CNAME lookups; no need to recurse
```

t/recursor-A.t
```diff
@@ -259,4 +270,10 @@ if ( $ENV{ZONEMASTER_RECORD} ) {
     Zonemaster::Engine::Nameserver->save( $datafile );
 }
 
+TODO: {
+    local $TODO = "Need to create zones with those errors: ";
+    my @missing_tags = qw( CNAME_UNRESOLVABLE );
+    warn $TODO, "\n\t", join("\n\t", @missing_tags), "\n";
```
marc-vanderwal: You could write the test already, even if you haven’t set up the infrastructure yet. It will fail, but because it’s in a TODO block, it is expected to fail, and therefore this failure will not fail the test suite itself. And if it passes, you’ll get a message about an “unexpectedly passing TODO test” reminding you to remove the TODO block.

marc-vanderwal (commented):
So far, so good.

lib/Zonemaster/Engine/Recursor.pm
```diff
@@ -264,12 +264,12 @@ sub _resolve_cname {
...
-    }
-    else {
-        # What do do here?
+        return ( $p, $state );
     }
 
-    return ( $p, $state );
+    # Catch-all; unforeseen problem in CNAME resolution
+    Zonemaster::Engine->logger->add( CNAME_UNRESOLVABLE => { name => $name, type => $type, target => $target } );
+    return ( undef, $state );
```
marc-vanderwal, line 299: We’ll reach that case when the recursor is trying to resolve a name that is an alias to a subdomain of that name, and there is no resource record of the requested type at the alias’ target. For example, the `break-me.test` domain name in the zone file below:

```zone
$ORIGIN test.
;; SOA records, NS records, etc.

break-me                IN  CNAME  sub.break-me
;; sub.break-me is an empty non-terminal
*.sub.break-me          IN  AAAA   ::
```

marc-vanderwal:
The `CNAME_UNRESOLVABLE` tag is output in some cases where the final query for a CNAME chain gives NODATA. That case can happen fairly frequently and should simply be treated like a negative response (i.e. NXDOMAIN), not like a resolution failure (as if we had gotten SERVFAIL).

I’d suggest a different name for that tag. How about `CNAME_TO_NODATA`?

matsduf:
> I’d suggest a different name for that tag. How about `CNAME_TO_NODATA`?

It think that is a good suggestion. Should there be another tag, `CNAME_TO_NXDOMAIN`?

For me `CNAME_UNRESOLVABLE` sounds like we cannot resolve the CNAME record.

marc-vanderwal:
Yes, a `CNAME_TO_NXDOMAIN` might make sense too.

In any case, I don’t think the ERROR level is appropriate for these tags because they do not indicate a problem so severe that the zone is unresolvable.

matsduf:
@tgreenx, will you fix the conflict?

tgreenx:
> @tgreenx, will you fix the conflict?

@matsduf @marc-vanderwal done, please re-review.

tgreenx:
> I’d suggest a different name for that tag. How about `CNAME_TO_NODATA`?

Added.

> > I’d suggest a different name for that tag. How about `CNAME_TO_NODATA`?
> 
> It think that is a good suggestion. Should there be another tag, `CNAME_TO_NXDOMAIN`?

Not needed, NXDOMAIN responses are already handled by the calling method (`_recurse()`).

marc-vanderwal (dismissed):

lib/Zonemaster/Engine/Recursor.pm
```diff
@@ -277,17 +264,39 @@ sub _resolve_cname {
...
+            Zonemaster::Engine->logger->add( CNAME_FOLLOWED_IN_ZONE => { name => $name, type => $type, target => $target } );
+            return ( $p, $state );
+        }
+
+        # There is a record of type QTYPE but with different owner name than CNAME target; we have to recurse again
+        Zonemaster::Engine->logger->add( CNAME_NO_MATCH => { name => $name, type => $type, target => $target, owner_names => join( ';', map { $_->owner } $p->get_records( $type ) ) } );
+    }
+
+    # CNAME target is out of zone, so make a new recursive lookup
     unless ( $name->is_in_bailiwick( $target ) ) {
```
matsduf, line 280: Even if name "is_in_bailiwick" it can be out of zone. Name can belong to a child zone (or several delegations down).

lib/Zonemaster/Engine/Recursor.pm
```diff
@@ -277,17 +264,39 @@ sub _resolve_cname {
...
+
+        if ( $p and $p->aa and $p->rcode eq 'NOERROR' ) {
+            Zonemaster::Engine->logger->add( CNAME_TO_NODATA => { name => $name, type => $type, target => $target } );
+            return ( $p, $state );
+        }
     }
 
-    return ( $p, $state );
+    # Catch-all; unforeseen problem in CNAME resolution
+    Zonemaster::Engine->logger->add( CNAME_UNRESOLVABLE => { name => $name, type => $type, target => $target } );
```
matsduf, line 298: I think the tag CNAME_UNRESOLVABLE should be defined or at least described. Probably together with CNAME_TO_NODATA and other tags.

lib/Zonemaster/Engine/Recursor.pm
```diff
@@ -277,17 +264,39 @@ sub _resolve_cname {
...
+    return ( undef, $state );
```
marc-vanderwal, line 299: In order to be consistent with the `Zonemaster::Engine::Nameserver->query()` method, I think we should return undef if and only if we had a timeout. Now we hit this case when the packet isn’t something we expect, e.g. a FORMERR response. We should still return that packet to the caller.

lib/Zonemaster/Engine/Recursor.pm
marc-vanderwal, line 234: If we are following a chain of aliases (CNAMEs) and at some point, the owner name of the next item in the chain is in a different case than the previous item’s CNAME’s RDATA, then we erroneously report a broken chain. There’s [an example](https://github.com/zonemaster/zonemaster/issues/1505#issuecomment-4788893619) in the wild of that.

This line should be:
```perl
        $cnames{lc( $rr_owner )} = $rr_target;
```

matsduf: Isn't clearer to write:
```
$rr_target = lc( $rr_target );
$rr_owner = lc( $rr_owner );

$seen_targets{ $rr_target } = 1;
$forbidden_targets{ $rr_owner } = 1;
$cnames{ $rr_owner } = $rr_target;
```

tgreenx: > If we are following a chain of aliases (CNAMEs) and at some point, the owner name of the next item in the chain is in a different case than the previous item’s CNAME’s RDATA, then we erroneously report a broken chain. There’s [an example](https://github.com/zonemaster/zonemaster/issues/1505#issuecomment-4788893619) in the wild of that.

Updated. Also I've updated test scenario `TARGET-NO-MATCH-CNAME` (see https://github.com/zonemaster/zonemaster/pull/1477) with varying case to test for that in the future.

> Isn't clearer to write:

Yes, but we should preserve the case of the target in case we re-query.

lib/Zonemaster/Engine/Recursor.pm
tgreenx: Updated with the wording of RFC9499, i.e. "in-domain".

lib/Zonemaster/Engine/Recursor.pm
tgreenx: I'm not sure. There are various cases where `Zonemaster::Engine::recurse()`, from which this method `_resolve_cname()` is called, will return `undef` already. But maybe you're right, I'll think on it a bit more.

lib/Zonemaster/Engine/Recursor.pm
tgreenx: Updated documentation - but without description of tags.

lib/Zonemaster/Engine/Recursor.pm
matsduf: I support the comment of @marc-vanderwal. Getting empty answer section or nothing matching the last CNAME in a chain is not an error as such, at least not what you expect undef from.

lib/Zonemaster/Engine/Recursor.pm
```diff
@@ -277,17 +264,39 @@ sub _resolve_cname {
...
+        if ( $p->has_rrs_of_type_for_name( $type, $target ) ) {
+            Zonemaster::Engine->logger->add( CNAME_FOLLOWED_IN_ZONE => { name => $name, type => $type, target => $target } );
+            return ( $p, $state );
+        }
+
+        # There is a record of type QTYPE but with different owner name than CNAME target; we have to recurse again
+        Zonemaster::Engine->logger->add( CNAME_NO_MATCH => { name => $name, type => $type, target => $target, owner_names => join( ';', map { $_->owner } $p->get_records( $type ) ) } );
+    }
+
+    # CNAME target is not in-domain, so make a new recursive lookup
```
matsduf, line 279: I do not think this is a proper use of the term "in-domain".


```suggestion
    # CNAME target is not in the same zone as the CNAME owner name, so make a new recursive lookup
```

lib/Zonemaster/Engine/Recursor.pm
matsduf: Why no description of the tags?

matsduf (changes requested):

lib/Zonemaster/Engine/Recursor.pm
tgreenx: Consider that the CNAME resolution method (`_resolve_cname()`) is not called directly (it is only called by the `_recurse()` method) and leaves much of the query/response handling to that method, such as timeouts and responses with different RCODES.
Specifically, the `_resolve_cname()` is only called when the response looks like an actual answer. And in this method there are already multiple cases where `undef` is explicitly returned, such as errors of various kind.

So overall I think that returning `undef` here is fine.

lib/Zonemaster/Engine/Recursor.pm
tgreenx: Done

lib/Zonemaster/Engine/Recursor.pm
```diff
@@ -645,37 +654,64 @@ Returns a list of L<Zonemaster::Engine::Nameserver> objects.
...
+This message tag indicates that the CNAME target was found in the current response and has the requested record type,
+so no additional recursion is needed.
+
 =item CNAME_FOLLOWED_OUT_OF_ZONE
 
+This message tag indicates that the CNAME target is outside the queried name's zone and is resolved through a new recursive lookup.
+
+=item CNAME_TO_NODATA
+
+This message tag indicates that the in-bailiwick CNAME target was resolved authoritatively but without the requested record type.
```
matsduf, line 674: I do not think we should use "in-bailiwick".
```suggestion
This message tag indicates that the CNAME target was resolved authoritatively. The name exists but not with requested record type.
```

lib/Zonemaster/Engine/Recursor.pm
```diff
@@ -645,37 +654,64 @@ Returns a list of L<Zonemaster::Engine::Nameserver> objects.
...
-Note that CNAME records are also validated and, in case of an error, an empty (C<undef>) L<packet|Zonemaster::Engine::Packet>
-is returned and one of the following message tags will be logged:
+Note that the resolution has multiple validation steps and, in case of an error, the first return value will be C<undef> and one of the
+following message tags is logged:
 
 =over
 
 =item CNAME_CHAIN_TOO_LONG
 
+This message tag indicates that the consecutive CNAME chain lookups are longer than the maximum allowed length.
```
matsduf, line 685:
```suggestion
This message tag indicates that the resolved chain of CNAME records is longer than the maximum allowed length.
```

lib/Zonemaster/Engine/Recursor.pm
```diff
@@ -645,37 +654,64 @@ Returns a list of L<Zonemaster::Engine::Nameserver> objects.
...
+
 =item CNAME_LOOP_INNER
 
+This message tag indicates that there is a loop in the CNAME RRset of the current response.
```
matsduf, line 689: I am not sure what you mean. All members of an RRset has always the same owner name. Is the following what you mean?

```suggestion
This message tag indicates that there is a loop in the CNAME record, i.e. the target name is identical to the owner name.
```

lib/Zonemaster/Engine/Recursor.pm
```diff
@@ -645,37 +654,64 @@ Returns a list of L<Zonemaster::Engine::Nameserver> objects.
...
+
 =item CNAME_LOOP_OUTER
 
+This message tag indicates that the CNAME target has already been followed in a previous lookup.
```
matsduf, line 693: Unclear. Is the following what you mean?

```suggestion
This message tag indicates that the target name of the last CNAME record in a chain is identical to a previous CNAME record in the same chain.
```

lib/Zonemaster/Engine/Recursor.pm
```diff
@@ -645,37 +654,64 @@ Returns a list of L<Zonemaster::Engine::Nameserver> objects.
...
+
 =item CNAME_NO_MATCH
 
+This message tag indicates that there is a record of the requested type but with different owner name than the CNAME target.
```
matsduf, line 697: Does this refer to the answer section? That it contains a CNAME record and another DNS record matching the QTYPE but not the CNAME target name?

lib/Zonemaster/Engine/Recursor.pm
```diff
@@ -645,37 +654,64 @@ Returns a list of L<Zonemaster::Engine::Nameserver> objects.
...
+
 =item CNAME_RECORDS_CHAIN_BROKEN
 
+This message tag indicates that the CNAME chain from the RRset of the current response is broken.
```
matsduf, line 701: What is the difference from CNAME_TO_NODATA?

lib/Zonemaster/Engine/Recursor.pm
```diff
@@ -645,37 +654,64 @@ Returns a list of L<Zonemaster::Engine::Nameserver> objects.
...
+
 =item CNAME_RECORDS_MULTIPLE_FOR_NAME
 
+This message tag indicates that there are more than one CNAME record with the same owner name in the RRset of the current response.
+
 =item CNAME_RECORDS_TOO_MANY
 
+This message tag indicates that there are too many CNAME records in the RRset of the current response.
```
matsduf, line 709: Too many for what? Same owner name? A chain?

lib/Zonemaster/Engine/Recursor.pm
```diff
@@ -645,37 +654,64 @@ Returns a list of L<Zonemaster::Engine::Nameserver> objects.
...
+
+=item CNAME_UNRESOLVABLE
+
+This message tag indicates that the CNAME resolution could not be successfully completed.
```
matsduf, line 713: I guess this is not CNAME_TO_NODATA? Is it CNAME_TO_NXDOMAIN? Or is it not possible to resolve target name (SERVFAIL, servers/zone not reachable)?

matsduf (changes requested):

lib/Zonemaster/Engine/Recursor.pm
tgreenx: No, rather that there is a loop in the CNAME chain from within a single DNS response, e.g.:
```
example.tld           CNAME   target1.example.tld
target1.example.tld   CNAME   example.tld
```

lib/Zonemaster/Engine/Recursor.pm
tgreenx: I guess yes, but specifically when doing additional CNAME lookups. So it is a loop too, but not from within a single response (i.e lookup) this time but another one, e.g:

```
Query 1 (example.tld) -> Response 1:

example.tld     CNAME   example.other

Query 2 (example.other) -> Response 2:

example.other   CNAME   example.tld
```

lib/Zonemaster/Engine/Recursor.pm
tgreenx: Yes exactly.

lib/Zonemaster/Engine/Recursor.pm
tgreenx: CNAME_TO_NODATA has a valid CNAME chain. While this message tag is for cases such as in a response there is e.g.:
```
example.tld           CNAME   target0.example.tld
target1.example.tld   CNAME   target2.example.tld
```

lib/Zonemaster/Engine/Recursor.pm
tgreenx: Just too many CNAME records within in a single response.

lib/Zonemaster/Engine/Recursor.pm
tgreenx: > I guess this is not CNAME_TO_NODATA? Is it CNAME_TO_NXDOMAIN? Or is it not possible to resolve target name (SERVFAIL, servers/zone not reachable)?

No. NXDOMAIN and SERVFAIL are handled by the calling method `_recurse()`. This tag is mainly here has a "catch-all" message to indicate something went wrong.

tgreenx:
@matsduf As discussed I've made another pass on the documentation, please re-review

PR #1521, opened 2026-04-14
Independent caching of custom name server recursive lookups

### LINKED ISSUE #1502: Cache key isolation issue in recursive lookups
`Recursor->recurse()` supports passing a custom nameserver set, but the recursive cache key is only `(name, type, class)` (`lib/Zonemaster/Engine/Recursor.pm`, around lines 98-110). That means a result learned from a custom/untrusted nameserver context can be reused later by normal root-based recursion for the same qname/qtype/qclass. It would be safer if cache entries were scoped by resolver context (for example, root mode vs specific NS set), or if custom-NS recursion bypassed the shared cache.

tgreenx:
## Purpose

This PR proposes to change the caching method of the `Zonemaster::Engine::Recursor::recurse()` method by leveraging [Memoize](https://perldoc.pl/Memoize) instead. This allows for easy and independent caching of queries made with custom sets of name servers.
Moreover a few drive-by changes in the `Zonemaster::Engine::Recursor` module are included as well.

## Context

Fixes #1502 

## Changes

Main changes:
- Scope caching of custom name server in recursive lookups, using Memoize
- Remove documentation of `%recurse_cache`
- Update unit tests

Drive-by changes:
- Avoid potential variable mutation issue
- Rename variable to prevent shadowing
- Minor documentation changes
- Properly qualify 'get_ns_from()' as an internal method

## How to test this PR

Unit tests are updated and should pass.
Note: the last test in `t/recursor.t` was changed to only run with live data, as it would fail with recorded data (but this is independent of this PR as it also happens on current master when re-recording the test data).


marc-vanderwal:
Simply [memoizing](https://perldoc.pl/Memoize) this method might be both a much simpler and a more correct solution to the issue this PR is fixing.

What we really want is for `recurse()` to return exactly the same thing if it is called with exactly the same arguments as a previous call. Except this is not the case when calling `recurse()` with exactly the same `$name`, `$type`, `$dns_class` arguments but a different value for the `$ns` argument.

With your solution, the following calls are isolated from each other:

```perl
# The $ns argument should be an arrayref on Zonemaster::Engine::Nameserver objects
# but I just give a list of strings for illustration purposes
Zonemaster::Engine::Recursor->recurse( "www.example.com", "AAAA", "IN" );
Zonemaster::Engine::Recursor->recurse( "www.example.com", "AAAA", "IN",
    [ qw( a.my-root-servers.test b.my-root-servers.test ) ] );
```

But these two are not:

```perl
Zonemaster::Engine::Recursor->recurse( "www.example.com", "AAAA", "IN",
    [ qw( a.my-root-servers.test b.my-root-servers.test ) ] );
Zonemaster::Engine::Recursor->recurse( "www.example.com", "AAAA", "IN",
    [ qw( c.my-root-servers.test d.my-root-servers.test ) ] );
```

So this PR only partially fixes the problem.

I really think that we have a perfect use case for [Memoize](https://perldoc.pl/Memoize) here. It should be done in a way that calling `recurse()` with no `$ns` argument and calling it with `[ root_servers() ]` as an argument are considered equivalent. Just beware of the need for [argument normalization for Memoize](https://perldoc.pl/Memoize#NORMALIZER) or you’ll end up with issues like #1470.

matsduf:
> ## Purpose
> 
> This PR proposes to cache separately the responses of recursive lookups with or without custom name servers. Moreover a few drive-by changes in the `Zonemaster::Engine::Recursor` module are included as well.

@tgreenx, can you make the "purpose" more explicit and easier to understand?

matsduf:
> But these two are not:

Why is it a problem. for Zonemaster, if different nameservers give different responses?

marc-vanderwal:
> ```perl
> Zonemaster::Engine::Recursor->recurse( "www.example.com", "AAAA", "IN",
>     [ qw( a.my-root-servers.test b.my-root-servers.test ) ] );
> Zonemaster::Engine::Recursor->recurse( "www.example.com", "AAAA", "IN",
>     [ qw( c.my-root-servers.test d.my-root-servers.test ) ] );
> ```

> Why is it a problem. for Zonemaster, if different nameservers give different responses?

The second function call, with `{c,d}.my-root-servers.test` as root hints, will still return whatever was cached from the first function call, i.e. what `{a,b}.my-root-servers.test` returned. That seems like a dangerous caveat to me.

matsduf:
@marc-vanderwal, do you consider this PR to be complete?

marc-vanderwal:
I don’t know; @tgreenx is the best person to ask.

matsduf:
> I don’t know; @tgreenx is the best person to ask.

I'm sorry. I mixed you up.

matsduf:
@tgreenx, do you consider this PR to be complete?

tgreenx:
> @tgreenx, do you consider this PR to be complete?

Not anymore, I'll have to rework this PR to take care of the caveat raised by @marc-vanderwal.

tgreenx:
I've updated this PR with the suggestion from @marc-vanderwal, please re-review.

tgreenx:
CI is failing, seems like I'll have to re-record some unit tests data, but the PR is ready.

lib/Zonemaster/Engine/Recursor.pm
```diff
@@ -95,22 +95,40 @@ sub recurse {
...
+# Cache the results of recurse() to speed up repeated queries for the same name, type,
+# class and set of name servers. Parameters are normalized to ensure that the cache is
+# hit for semantically identical queries.
+memoize('recurse', NORMALIZER => sub {
+    my ($class, $name, $type, $dns_class, $ns) = @_;
+
+    $name = name($name);
+    $type = defined $type ? $type : 'A';
+    $dns_class = defined $dns_class ? $dns_class : 'IN';
+    $ns = defined $ns ? $ns : [ root_servers() ];
```
marc-vanderwal: I’d suggest using the defined-or-assignment operator for clarity:
```suggestion
    $type //= 'A';
    $dns_class //= 'IN';
    $ns //= [ root_servers() ];
```

marc-vanderwal:
Looks good! I just have a small style suggestion. I’ll approve as soon as CI is fixed.

tgreenx:
> CI is failing, seems like I'll have to re-record some unit tests data, but the PR is ready.

Unit tests data have been re-recorded in https://github.com/zonemaster/zonemaster-engine/pull/1521/commits/c0c70e1dda1c02c484dc0261ca67e244dcba57fc - CI now passes, please re-review @marc-vanderwal @matsduf.

marc-vanderwal (dismissed):

tgreenx:
@marc-vanderwal sorry, I've pushed a small update so that we have a proper unit test when clearing the cache with `Recursor->clear_cache()`. Could you re-approve?

marc-vanderwal (dismissed):

tgreenx:
@marc-vanderwal @matsduf I rebased on latest develop to fix recent conflicts in unit tests, please re-review.

marc-vanderwal (approved):

matsduf (approved):

MichaelTimbert:
work on rocky 8

PR #1528, opened 2026-05-27
Harden Nameserver15 against creative version strings

marc-vanderwal:
## Purpose

This PR addresses situations where Zonemaster-GUI fails to display a test’s results if at least one of the byte strings obtained in response of `version.bind` or `version.server/CH/TXT` queries contained specific sequences of non-ASCII bytes.

The same issue could also lead to injection of terminal control codes (e.g. [ANSI escape codes]) in zonemaster-cli.

Before:
image

After:
image

And it also works on the GUI:
image

## Context

See the following issues for context:
* zonemaster/zonemaster-backend#1216
* zonemaster/zonemaster#1437
* zonemaster/zonemaster-gui#544
* zonemaster/zonemaster-cli#458

## Changes

* Add string escaping function in `Zonemaster::Engine::Util`;
* In Nameserver15, convert `version.bind`/`version.server` strings using that function before logging the message.

**Note**: this PR will not fix test results that already are in the database. To fix old tests, we would need a migration script in Zonemaster-Backend.

## How to test this PR

* Unit tests were added and should still pass.
* With Zonemaster-CLI, run `zonemaster-cli --test nameserver15 servfail.fr`. One of the name servers’ version strings contains [ANSI escape codes] and various non-ASCII and non-printable characters. They should all be replaced with backslash-escaped decimal codes, and no text should appear in blue.
* With Zonemaster-Backend installed and running, run `zmtest servfail.fr`. The results should appear normally, instead of getting an internal server error message. A test against `s42.re` should also complete successfully and have its results display normally.

[ANSI escape codes]: https://en.wikipedia.org/wiki/ANSI_escape_code

### DIFF
--- a/lib/Zonemaster/Engine/Test/Nameserver.pm
+++ b/lib/Zonemaster/Engine/Test/Nameserver.pm
@@ -1747,10 +1747,10 @@ sub nameserver15 {
                         push @wrong_record_class, $ns;
                     }
 
-                    my $string = $rr->txtdata;
+                    my $string = Zonemaster::Engine::Util::escape($rr->txtdata);
                     $string =~ s/^\s+|\s+$//g; # Remove leading and trailing spaces
 
-                    if ( $string and $string ne "") {
+                    if ( $string ne "" ) {
                         push @{ $txt_data{$string}{$query_name} }, $ns;
                         delete $sending_version_query{$ns};
                     }
--- a/lib/Zonemaster/Engine/Test/Nameserver.pm
+++ b/lib/Zonemaster/Engine/Test/Nameserver.pm
@@ -5,17 +5,17 @@ use warnings;
 
 use version; our $VERSION = version->declare( "v1.1.0" );
 
+use JSON::PP;
 use List::MoreUtils qw[uniq none];
 use Locale::TextDomain qw[Zonemaster-Engine];
-use Readonly;
-use JSON::PP;
 use Net::IP::XS;
+use Readonly;
 
-use Zonemaster::Engine::Profile;
 use Zonemaster::Engine::Constants qw[:ip];
+use Zonemaster::Engine::Profile;
 use Zonemaster::Engine::Test::Address;
-use Zonemaster::Engine::Util;
 use Zonemaster::Engine::TestMethods;
+use Zonemaster::Engine::Util qw( escape_unprintable );
 
 =head1 NAME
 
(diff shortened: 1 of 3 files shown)


lib/Zonemaster/Engine/Util.pm
```diff
@@ -285,6 +319,18 @@ Check if a test is blacklisted and should run or not.
 
 Check if IP version operations are permitted. Tests are done against Zonemaster::Engine::Profile->effective content.
 
+=item escape
```
tgreenx: What about a more specific name for this method? e.g:
```suggestion
=item escape_byte_string
```

tgreenx (dismissed):
LGTM. Just one minor suggestion that I can live without.

lib/Zonemaster/Engine/Util.pm
matsduf: Or "escape_unprintable"

matsduf (dismissed):
Testing looks fine.

lib/Zonemaster/Engine/Util.pm
marc-vanderwal: Sure, I can rename this function to `escape_unprintable`.

tgreenx (dismissed):

lib/Zonemaster/Engine/Test/Nameserver.pm
```diff
@@ -1747,8 +1747,11 @@ sub nameserver15 {
                         push @wrong_record_class, $ns;
                     }
 
-                    my $string = escape_unprintable($rr->txtdata);
-                    $string =~ s/^\s+|\s+$//g; # Remove leading and trailing spaces
+                    # Trim all ASCII whitespace first, then escape (so that
+                    # other ASCII whitespace characters like tabs or form
+                    # feeds also get trimmed instead of being turned into
```
matsduf: I am not convinced especially form feeds should not be escaped. Both are unprintable control characters, even line feed is in this context. I think it is a cleaner solution to escape them all.

marc-vanderwal: Fair point. Let’s just trim spaces (ASCII 0x20) and nothing else.

tgreenx (approved):

matsduf (approved):

matsduf:
Successfully tested for the 2026.1 release.

PR #1530, opened 2026-06-04
Replace Perl INIT block with explicit initialization

### LINKED ISSUE #1439: "Too late to run INIT block at /usr/local/share/perl/5.34.0/Zonemaster/Engine.pm line 31." log message
When running zm-rpcapi on Ubuntu 22.04 LTS I see the message `Too late to run INIT block at /usr/local/share/perl/5.34.0/Zonemaster/Engine.pm line 31.` when running `systemctl status zm-rpcapi`

> Mar 03 09:15:06 zonemaster-backend-dev-eu-north-1c systemd[1]: Starting LSB: A JSON-RPC frontend for Zonemaster Backend...
Mar 03 09:15:12 zonemaster-backend-dev-eu-north-1c zm-rpcapi[1081]: Too late to run INIT block at /usr/local/share/perl/5.34.0/Zonemaster/Engine.pm line 31.
Mar 03 09:15:14 zonemaster-backend-dev-eu-north-1c zm-rpcapi[1081]: 2025-03-03T09:15:14Z [1081] [NOTICE] [Zonemaster::Backend::Config] Loading config: /etc/zonemaster/backend_config.ini
Mar 03 09:15:14 zonemaster-backend-dev-eu-north-1c zm-rpcapi[1081]: 2025-03-03T09:15:14Z [1081] [INFO] [Plack::Sandbox::_2fusr_2flocal_2fbin_2fzonemaster_backend_rpcapi_2epsgi] Enabling add_api_user method
Mar 03 09:15:14 zonemaster-backend-dev-eu-north-1c zm-rpcapi[1081]: 2025-03-03T09:15:14Z [1081] [INFO] [Plack::Sandbox::_2fusr_2flocal_2fbin_2fzonemaster_backend_rpcapi_2epsgi] Enabling add_batch_job method
Mar 03 09:15:14 zonemaster-backend-dev-eu-north-1c systemd[1]: Started LSB: A JSON-RPC frontend for Zonemaster Backend.

tgreenx:
## Purpose

This PR removes the use of a Perl INIT block in Engine and replaces it with an explicit call to `init_engine()` at module load time.

The current use of INIT causes warnings when `Zonemaster::Engine` is loaded in environments where Perl’s INIT phase has already completed (e.g. PSGI/Plack-based deployments). Since the initialization logic is already idempotent, the INIT block is no longer necessary.

## Context

Fixes https://github.com/zonemaster/zonemaster-engine/issues/1439

## Changes

- Replace INIT block with an explicit call to `init_engine()` at module load time

## How to test this PR

- Unit tests should still pass.
- Starting Zonemaster-Backend should no longer show a "Too late to run INIT block" warning message.

### DIFF
--- a/lib/Zonemaster/Engine.pm
+++ b/lib/Zonemaster/Engine.pm
@@ -250,7 +250,7 @@ This manual describes the main L<Zonemaster::Engine> module. If what you're afte
 
 =item init_engine()
 
-Run the initialization tasks if they have not been run already. This method is called automatically in INIT block.
+Run the initialization tasks if they have not been run already. This method is called automatically at runtime.
 
 =item test_zone($name)
 
--- a/lib/Zonemaster/Engine.pm
+++ b/lib/Zonemaster/Engine.pm
@@ -26,19 +26,16 @@ use Zonemaster::Engine::Test;
 use Zonemaster::Engine::Recursor;
 use Zonemaster::Engine::ASNLookup;
 
-INIT {
-    init_engine();
-}
-
 our $logger;
 our $recursor = Zonemaster::Engine::Recursor->new;
 
-my $init_done = 0;
-
+my $init_done;
 sub init_engine {
     return if $init_done++;
     Zonemaster::Engine::Recursor::init_recursor();
 }
+
+init_engine();
 
 sub logger {
     return $logger //= Zonemaster::Engine::Logger->new;
--- a/lib/Zonemaster/Engine/Nameserver.pm
+++ b/lib/Zonemaster/Engine/Nameserver.pm
@@ -8,7 +8,6 @@ use version; our $VERSION = version->declare("v1.1.16");
 use Class::Accessor qw[ antlers ];
 
 use Zonemaster::Engine::DNSName;
-use Zonemaster::Engine;
 use Zonemaster::Engine::Packet;
 use Zonemaster::Engine::Nameserver::Cache;
 use Zonemaster::Engine::Recursor;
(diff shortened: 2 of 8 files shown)


mattias-p (dismissed):
Looks good!

Though we should probably also update the documentation so we don't claim that init_update is called in the INIT phase.

tgreenx:
> Looks good!

Unfortunately it seems to break a few unit tests due to circular dependencies between Zonemaster-Engine modules. I'll try to fix.

> Though we should probably also update the documentation so we don't claim that init_update is called in the INIT phase.

Will update!

mattias-p:
> Unfortunately it seems to break a few unit tests due to circular dependencies between Zonemaster-Engine modules. I'll try to fix.

Oh, no, not that again. I don't know how many times I've tried to chip away at that. Though I think I've been focusing more on disentangling Zonemaster::Engine::Util.

Here's an approach that I think would work. Feel free to use it or ignore it, or something in between.

* Zonemaster::Engine should contain no logic at all.
* All the logic that currently lives in Zonemaster::Engine should be moved out to other modules. Perhaps some new ones need to be invented to accommodate functionality that doesn't belong in any existing module.
* All(?) procedures under Zonemaster::Engine could be kept around for the convenience of other packages (e.g., Zonemaster::CLI and Zonemaster::Backend).
* Nothing below Zonemaster::Engine should import Zonemaster::Engine.

matsduf (approved):

mattias-p (approved):

marc-vanderwal (approved):

MichaelTimbert:
work on rocky 8

PR #1531, opened 2026-06-10
Revert "Remove obsolete TO_JSON methods"

marc-vanderwal:
## Purpose

This PR reverts a commit introduced in #1517 that deleted what I thought was dead code. It turns out it wasn’t dead code at all.

The issue manifests itself by the fact that some messages, like `B02_AUTH_RESPONSE_SOA`, get `null` instead of a domain name for the `domain` argument when converted to either the raw representation, or JSON.

I wrongly assumed that no other place in the code would attempt to convert domain names to JSON by calling the `TO_JSON` method. Also, JSON::PP is set up in such a way that when a blessed reference is to be converted and the class has no `TO_JSON` method, its JSON representation is `null` instead of generating a fatal error. Hence the appearance of `null`s in raw log entry representation or JSON representations.

## Context

Follow-up to #1517.

## Changes

* Revert a commit removing `TO_JSON` methods in two classes
* Add unit tests

## How to test this PR

Unit tests have been added and should pass.

For good measure, run the following command and expect the following output:
```
$ zonemaster-cli --test basic02 --level info zonemaster.net --json | \
    jq '.results[] | select(.tag == "B02_AUTH_RESPONSE_SOA") | .args.domain'
"zonemaster.net"
```

### DIFF
--- a/t/logger.t
+++ b/t/logger.t
@@ -1,6 +1,7 @@
-use Test::More;
-use Test::Fatal;
 use File::Slurp;
+use JSON::PP;
+use Test::Fatal;
+use Test::More;
 
 BEGIN {
     use_ok( 'Zonemaster::Engine' );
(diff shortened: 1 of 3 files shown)


matsduf:
I am sceptical to revert commits. We have to make sure that it does not mess things up as we have seen in the past.

marc-vanderwal:
I understand your skepticism, but the commit I revert is a fairly trivial one, so I think it is safe to do it that way.

matsduf (dismissed):

t/logger.t
```diff
@@ -36,6 +37,32 @@ ok( scalar( @{ Zonemaster::Engine->logger->entries } ) >= 2, 'expected number of
...
 
+$entry = info( 'TEST2', { an => 'argument', domain => Zonemaster::Engine::Util::name( 'logger.test.example' ) } );
+
+# FIXME: Ideally, we really should instead have domain=logger.test.example
+# without the double quotes around the domain name.
+#
+# Providing the domain name as a regular string gives us no double quotes
+# around the domain name. Providing it directly as a Zonemaster::Engine::DNSName
+# does give us double quotes, however. That’s because blessed objects get
+# turned into their JSON representations when.
```
tgreenx:
```suggestion
# turned into their JSON representations then.
```

t/logger.t
```diff
@@ -36,6 +37,32 @@ ok( scalar( @{ Zonemaster::Engine->logger->entries } ) >= 2, 'expected number of
...
+#
+# It’s an annoying inconsistency, but any attempt at a fix might cause
+# repercussions when translating a log entry to a human language, a string on
+# JSON. For now, we expect the broken behavior, but that should be fixed
```
tgreenx: > a string on JSON

This reads wrong I think

marc-vanderwal: That should have been “or”. Fixing.

t/logger.t
marc-vanderwal: I meant something very different. I’ll fix that.

matsduf (approved):

tgreenx (approved):

MichaelTimbert:
tested on rocky 8

PR #1546, opened 2026-08-06
Fix edge cases in Get-Delegation and Get-Del-NS-Names-and-IPs when dealing with fake delegation

### LINKED ISSUE #1545: MethodsV2 Get-Delegation and Get-Del-NS-Names-and-IPs do not return entire fake delegation
When giving fake delegation for the root zone, if there is a name server in the fake delegation that does not have any IP addresses, neither Get-Delegation nor Get-Del-NS-Names-and-IPs return that bare name server, despite the specification saying that they should.

The following reproducer script illustrates the problem:

```perl
#!/usr/bin/env perl

use v5.26.0;
use warnings;
                                                                                                                                         
use Data::Dumper;
use Test::More 'no_plan';

use Zonemaster::Engine;

my $zone = Zonemaster::Engine::Zone->new( { name => Zonemaster::Engine::DNSName->new( '.' ) } );

# Add fake delegation
Zonemaster::Engine::Recursor->remove_fake_addresses( '.' );
Zonemaster::Engine::Recursor->add_fake_addresses( '.', {
    ns1 => [ ],
    ns2 => [ qw(127.14.5.66 fda1:b2:c3:0:127:14:5:66) ]
} );
Zonemaster::Engine::Recursor->clear_cache();

{
    my @got = sort { "$a" cmp "$b" } @{ Zonemaster::Engine::TestMethodsV2->_get_delegation( $zone ) };
    my @expected = qw( ns1 ns2/127.14.5.66 ns2/fda1:b2:c3:0:127:14:5:66 );
    is_deeply( \@got, \@expected ) or diag ("Here’s what we got:\n" . join(", ", @got));
}

{
    my @got = sort { "$a" cmp "$b" } @{ Zonemaster::Engine::TestMethodsV2->get_del_ns_names_and_ips( $zone ) };
    my @expected = qw( ns1 ns2/127.14.5.66 ns2/fda1:b2:c3:0:127:14:5:66 );
    is_deeply( \@got, \@expected ) or diag ("Here’s what we got:\n" . join(", ", @got));
}

done_testing;
```

Currently, this is the output (on `develop`):
```
not ok 1
#   Failed test at reproducer.pl line 34.
#     Structures begin differing at:
#          $got->[0] = 'ns2/127.14.5.66'
#     $expected->[0] = 'ns1'
# Here’s what we got:
# ns2/127.14.5.66, ns2/fda1:b2:c3:0:127:14:5:66
not ok 2
#   Failed test at reproducer.pl line 40.
#     Structures begin differing at:
#          $got->[0] = 'ns2/127.14.5.66'
#     $expected->[0] = 'ns1'
# Here’s what we got:
# ns2/127.14.5.66, ns2/fda1:b2:c3:0:127:14:5:66
1..2
# Looks like you failed 2 tests of 2.
```

marc-vanderwal:
## Purpose

Both the Get-Delegation and Get-Del-NS-Names-and-IPs methods in MethodsV2 had implementations that deviated slightly from the specifications, in such a way that they could return incorrect data in some edge cases involving fake delegation. This PR brings these methods back in line with the specification; they now return correct results in these situations.

In order to help fix this bug, I’ve introduced a new class for a new type of data structure called a NameserverSet. This class implements sets that specialize in holding a mix of Nameserver and DNSName objects. The inspiration comes from the Consistency05 specification update in zonemaster/zonemaster#1523, where the test procedure prescribes sets containing a mix of name/IP pairs and plain names, where adding a name/IP pair to a set already containing the same name also removes the plain name. NameserverSet objects have the same semantics.

Although I initially used NameserverSets to update the implementation of Consistency05, I then stumbled upon the problem described in #1545. Then, I discovered that NameserverSets were a very useful abstraction to help fix the issue. We use these sort of sets of either Nameservers, DNSNames or both in many places in the codebase and I hope that NameserverSets could help make the code more concise, both in test methods and test cases.

*Note:* this PR also introduces another user-visible change. One of the commits improves the `cmp` operator overloads in the Zonemaster::Engine::DNSName and Zonemaster::Engine::Nameserver classes. Name server lists in CLI and GUI output are now properly sorted on name server name first, IP version second, and numeric IP address number (as a 32-bit integer in IPv4 and 128-bit integer in IPv6) third. For example, `ns1.example/2001:db8::1111` sorts after `ns1.example/2001:db8::2`. Some message tags might still contain name server lists sorted differently, depending on how the test cases are implemented.

## Context

Fixes #1545.

## Changes

* Improve the `cmp` operator overloads in Zonemaster::Engine::DNSName and Zonemaster::Engine::Nameserver
* Introduce the `Zonemaster::Engine::NameserverSet` class
* Fix an incorrect implementation of Get-Delegation and Get-Del-NS-Names-and-IPs, bringing it back in line with the specification
* Refactor these two methods (and especially Get-Del-NS-Names-and-IPs) in order to improve general code style

## How to test this PR

* Unit tests should still pass.
* Run the reproducer script in #1545. Its output should match the following text:
```
ok 1
ok 2
1..2
```

### DIFF
--- a/lib/Zonemaster/Engine/NameserverSet.pm
+++ b/lib/Zonemaster/Engine/NameserverSet.pm
@@ -1,0 +1,338 @@
+package Zonemaster::Engine::NameserverSet;
+
+use v5.26.0;
+use warnings;
+
+use Scalar::Util qw(blessed);
+
+use Zonemaster::Engine::DNSName;
+use Zonemaster::Engine::Nameserver;
+
+####
+#### Constructors
+####
+
+sub new {
+    my ( $proto, @initial_items ) = @_;
+    my $class = ref $proto || $proto;
+
+    my $self = {
+        _names => {},
+    };
+
+    bless $self, $class;
+    $self->push( @initial_items );
+
+    return $self;
+}
+
+####
+#### Mutators
+####
+
+sub push {
+    my ( $self, @items ) = @_;
+
+    for my $item ( @items ) {
+        next if not defined $item;
+        next if $item eq '';
+
+        if ( $item->isa('Zonemaster::Engine::DNSName') ) {
+            if ( not exists $self->{_names}{lc $item} ) {
+                $self->{_names}{lc $item} = $item;
+            }
+        }
+        elsif ( $item->isa('Zonemaster::Engine::Nameserver') ) {
+            my ( $name, $addr ) = ( $item->name, $item->address->short );
+
+            my $r = \$self->{_names}{lc $name};
+            if ( not defined $$r or ref $$r ne 'HASH' ) {
+                $$r = {};
+            }
+            $$r->{$addr} = $item;
+        }
+        else {
+            my $name = Zonemaster::Engine::DNSName->new( "$item" );
+            if ( not exists $self->{_names}{lc $name} ) {
+                $self->{_names}{lc $item} = $name;
+            }
+        }
+    }
+
+    return $self;
+}
+
+####
+#### Access to items
+####
+
+sub get {
+    my ( $self, $name ) = @_;
+
+    if ( not exists $self->{_names}{ lc $name } ) {
+        return ();
+    }
+
+    my $r = $self->{_names}{ lc $name };
+    if ( ref $r eq 'HASH' ) {
+        return values %$r;
+    }
+
+    return $r;
+}
+
+sub get_ips {
+    my ( $self, $name ) = @_;
+
+    if ( exists $self->{_names}{ lc $name } ) {
+        my $r = $self->{_names}{ lc $name };
+        if ( ref $r eq 'HASH' ) {
+            return values %$r;
+        }
+    }
+
+    return ();
+}
+
+sub items {
+    my ( $self ) = @_;
+
+    my @items;
+
+    for my $k ( keys %{ $self->{_names} } ) {
+        my $r = $self->{_names}{$k};
+        if ( ref $r eq 'HASH' ) {
+            CORE::push( @items, map { $r->{$_} } keys %$r );
+        }
+        else {
+            CORE::push( @items, $r );
+        }
+    }
+
+    return @items;
+}
+
+sub names {
+    my ( $self ) = @_;
+
+    my @result;
+
+    for my $k ( keys %{ $self->{_names} } ) {
+        my $r = $self->{_names}{$k};
+        if ( ref $r eq 'HASH' ) {
+            CORE::push( @result, ( values %$r )[0]->name() );
+        }
+        else {
+            CORE::push( @result, $r );
+        }
+    }
+
+    return @result;
+}
+
+sub sorted_items {
+    my ( $self ) = @_;
+
+    my @items;
+
+    for my $k ( sort keys %{ $self->{_names} } ) {
+        my $r = $self->{_names}{$k};
+        if ( ref $r eq 'HASH' ) {
+            CORE::push( @items, sort map { $r->{$_} } keys %$r );
+        }
+        else {
+            CORE::push( @items, $r );
+        }
+    }
+
+    return @items;
+}
+
+####
+#### Basic operations
+####
+
+sub is_empty {
+    my ( $self ) = @_;
+
+    return ( scalar %{ $self->{_names} } == 0 );
+}
+
+####
+#### Set theory operations
+####
+
+sub difference {
+    my ( $self, $other ) = @_;
+
+    my $only_in_left = __PACKAGE__->new();
+    my $only_in_right = __PACKAGE__->new();
+
+    my @left_items = $self->sorted_items();
+    my @right_items = $other->sorted_items();
+
+    while ( @left_items and @right_items ) {
+        my $cmp = $left_items[0] cmp $right_items[0];
+        if ( $cmp < 0 ) {
+            $only_in_left->push( shift @left_items );
+        }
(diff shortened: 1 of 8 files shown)


matsduf:
> Name server lists in CLI and GUI output are now properly sorted on name server name first, IP version second, and numeric IP address number (as a 32-bit integer in IPv4 and 128-bit integer in IPv6) third. For example, `ns1.example/2001:db8::1111` sorts after `ns1.example/2001:db8::2`.

This is a good change!

matsduf:
Obviously, we could improve the scenarios set for MethodsV2. Do you have suggestions for some new scenarios to be added?

marc-vanderwal:
> > Name server lists in CLI and GUI output are now properly sorted on name server name first, IP version second, and numeric IP address number (as a 32-bit integer in IPv4 and 128-bit integer in IPv6) third. For example, `ns1.example/2001:db8::1111` sorts after `ns1.example/2001:db8::2`.
> 
> This is a good change!

The change isn’t effective everywhere: there are test cases that still list name servers while sorting them wrong. Likely because the code explicitly converts the Nameserver objects to strings before sorting. Fixing this is out of scope for this PR but might be part of some broader refactoring of test case code.

> Obviously, we could improve the scenarios set for MethodsV2. Do you have suggestions for some new scenarios to be added?

Due to time constraints, I haven’t taken a more detailed look at the scenarios yet (I know I should have), but we are definitely missing something. Looking at the reproducer script I mentioned in the PR’s description, it shouldn’t be too hard.

matsduf:
> The change isn’t effective everywhere: there are test cases that still list name servers while sorting them wrong. Likely because the code explicitly converts the Nameserver objects to strings before sorting. Fixing this is out of scope for this PR but might be part of some broader refactoring of test case code.

I understand that, but I think that it is great to have sorting well defined that can be used in all test cases when they are updated.

> Due to time constraints, I haven’t taken a more detailed look at the scenarios yet (I know I should have), but we are definitely missing something. Looking at the reproducer script I mentioned in the PR’s description, it shouldn’t be too hard.

Improving the scenarios will be an eternal task!

lib/Zonemaster/Engine/TestMethodsV2.pm
```diff
@@ -393,24 +394,51 @@ Returns an arrayref of L<Zonemaster::Engine::Nameserver> objects, or C<undef> if
...
 
+    # Takes the names from NS resource records from a packet ($p)’s section ($ns_section), plus the
+    # address records from the additional section belonging to those NS names that are in-domain
+    # with respect to the zone name, and returns the combined names and possibly IPs as a
+    # Zonemaster::Engine::NameserverSet.
+    my sub packet_ns_names_and_in_domain_ips {
+        my ( $p, $ns_section ) = @_;
+
+        # Set of NS names from authority section
+        my %authority = map { lc name( $_->nsdname() ) => 1 }
```
marc-vanderwal: This variable shouldn’t be called `%authority`; this name was fine before refactoring, but not anymore.
Be sure to also edit the comment above.

t/methodsv2.t
```diff
@@ -765,7 +765,8 @@ my %subtests = (
               ns1.parent.child-ns-cname-4.methodsv2.xa/fda1:b2:c3:0:127:40:1:41
               ns2.parent.child-ns-cname-4.methodsv2.xa/127.40.1.42
               ns2.parent.child-ns-cname-4.methodsv2.xa/fda1:b2:c3:0:127:40:1:42 ) ],
-        [ qw( ns1-cname.child.parent.child-ns-cname-4.methodsv2.xa/127.40.1.51 ) ],
+        [ qw( ns1-cname.child.parent.child-ns-cname-4.methodsv2.xa/127.40.1.51
+              ns2-cname.child.parent.child-ns-cname-4.methodsv2.xa ) ],
```
marc-vanderwal, line 769: This will require an update of the corresponding test scenario specification.

Edit: done; see zonemaster/zonemaster#1528.

matsduf (dismissed):

lib/Zonemaster/Engine/NameserverSet.pm
```diff
@@ -0,0 +1,338 @@
...
+
+Items are sorted lexicographically based on their names, and if multiple
+L<Zonemaster::Engine::Nameserver> objects have the same name, these are sorted
+on IP addresses such that IPv4 addresses are sorted before IPv6 addresses, and
+within an address family, each address is sorted based on their integer
+values.
+
+=item is_empty()
+
+Returns true if and only ef the set contains no items.
```
tgreenx:
```suggestion
Returns true if and only if the set contains no items.
```

lib/Zonemaster/Engine/NameserverSet.pm
```diff
@@ -0,0 +1,338 @@
...
+=over
+
+=item new()
+
+Constructs a new empty name server set.
+
+=item push( $name_or_ns )
+
+Adds an item to the set, which can either be a L<Zonemaster::Engine::DNSName>
+or a L<Zonemaster::Engine::Nameserver> object.
```
tgreenx: It looks like the implementation also accepts arbitrary types (e.g. strings), which are then automatically attempted to be converted to `Zonemaster::Engine::DNSName` objects. The documentation should mention that.

tgreenx (approved):

matsduf (approved):

matsduf:
@marc-vanderwal, please merge.

PR #1552, opened 2026-09-15
KINDNS preset profile

tgreenx:
## Purpose

This PR adds the first "preset" profile, in this case related to KINDNS.

## Context

Relates to https://github.com/zonemaster/zonemaster/pull/1538

## Changes

- Add `share/profiles/kindns.json`

## How to test this PR

N/A


matsduf (dismissed):

matsduf (commented):
I suggest that is kept as a ".json" file. Suggestion `kindns-profile.json`.

tgreenx:
> I suggest that is kept as a ".json" file. Suggestion `kindns-profile.json`.

Well I went with that initially but then I remembered we also support YAML configuration file format (and even provide a working `profile.yaml` file), so I went with a more generic extension name.

matsduf:
> > I suggest that is kept as a ".json" file. Suggestion `kindns-profile.json`.
> 
> Well I went with that initially but then I remembered we also support YAML configuration file format (and even provide a working `profile.yaml` file), so I went with a more generic extension name.

I do not see the gain of not including ".json". `kindns-profile.json` makes it clear what format it has.

